PR #14, opened 2022-08-06
bugfixes and small aesthetic fixes

ikethefifth:
hope this is all ok!


Thlumyn:
Looks good, thanks 
We might have to change the draw menu items function so that it sets the button for the current screen to unavailable but this is still much better than the way its current coded in imo

ikethefifth:
What does button availability do? In testing the behavior seemed exactly the same - I clicked on the button for the screen already active, and nothing happened, as I expected

Thlumyn:
It just pulls up different color formatting for the button, so its more obvious to the user that the button isnt supposed to do anything

PR #17, opened 2022-08-07
apprentice bugfixes and updates

ikethefifth:
- save apps/former apps using semicolons
- med cat apps have mentors now
- former app list doesn't get duplicates


sedgestripe:
you are the best

PR #25, opened 2022-08-12
Save files from csv to JSON

djotaku:


ikethefifth:
You might want to leave the ability to load from CSV if no JSON file is found, so people can convert their save files

sedgestripe:
this is awesome

djotaku:
@ikethefifth that's a great suggestion that I hadn't thought of. However, it looks like there are some conflicts to resolve with some changes that happened while I was working on my PR. I need to investigate whether it'd easier for me to reconcile on my end and then push the updates to the PR or if one of the write access folks on this repo resolves and then I re-add CSV.

@sedgestripe thanks!

ikethefifth:
I would recommend merging on your end if possible so you can test it before it goes into the actual repository. But I just got write access so I can try if you want

PR #35, opened 2022-08-20
Restructured cat class + adding relationship class

Lixxis:
Before I continue working on the relationship class, I wanted to bring my changes to the dev branch. 
As far as I have tested, everything should work. Old clans can be taken along and are then converted into the new (more readable for me) format. I've just merged the latest changes to my branch, so everything should be up-to-date.

I know these are some major changes in parts, but the Cat class acting on two different ways is a bit 
 weird, at least for me.


Thlumyn:
The game doesnt run for me when I try to open this

PR #74, opened 2022-09-04
fixes space issue, new patrol event, scars (hopefully) appear in game 

Clayteeth:


Clayteeth:
added events and cats py to my files not sure if thall show in here

PR #103, opened 2022-09-08
Apprentices are not patrol leaders unless patrol is only apprentices

shrimpsicle:
ozzie ☁#2473 in #dev-bug-reports pointed out that apprentices could be chosen as patrol leaders even if they were not the most experienced cat, so now when creating a patrol, each cat will be checked if their status is apprentice or not. If not, they are added to list possible_patrol_leaders and a patrol leader is picked from that. If the patrol only contains apprentices, it will pick a random apprentice to be the leader

### DIFF
--- a/scripts/patrol.py
+++ b/scripts/patrol.py
@@ -13,6 +13,7 @@ class Patrol(object):
         self.patrol_event = None
         self.patrol_leader = None
         self.patrol_cats = []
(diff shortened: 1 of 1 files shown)


shrimpsicle:
follow up comment: im jade#1212 in the discord :)

Issue #102, opened 2022-09-08
cats with changed names retaining previous suffixes


shrimpsicle:
something ive noticed is if you change the name of a kit/apprentice before they finish their training (ex Icepaw -> Cloudpaw), when the cat finishes their training instead of receiving a proper suffix they will stay as whatever the changed name was (ex Cloudpaw doesn't become Cloudtail, stays as Cloudpaw).

sedgestripe:
solved in #114 by @rmalusa72!

PR #114, opened 2022-09-10
Name change bugfix

rmalusa72:
Previously, as mentioned in issue #102, changing a kit or apprentice name to one with the same kit/paw suffix overwrote the hidden suffix, meaning they would remain kit or paw even when growing up. This change keeps the hidden suffix unchanged if the new suffix matches the current special status.


shrimpsicle:
thank you!

PR #115, opened 2022-09-11
Readability - no functionality changes

rmalusa72:
- Simplified the conditional statements in buttons.py that add hotkey labels
- Combined the two ways to trigger a button (clicking on it and pressing hotkey) in draw_button, as they are the same
- Replacing ohter_cat with other_cat, opposit_relationship with opposite_relationship
- Shortening the list of suffixes by using the * syntax for duplicating elements in a list
- Adding some comments to these sections


rmalusa72:
Realized there's a problem with the names list; my mistake!

PR #116, opened 2022-09-11
Removing some repetitive stuff/typos 

rmalusa72:
- Simplified the conditional statements in buttons.py that add hotkey labels
- Combined the two ways to trigger a button (clicking on it and pressing hotkey) in draw_button, as they are the same
- Replacing ohter_cat with other_cat, opposit_relationship with opposite_relationship
- Shortening the list of suffixes by using the * syntax for duplicating elements in a list


sedgestripe:
resubmit if possible?

rmalusa72:
Will try to when I have time 👍

PR #143, opened 2022-09-23
fix: bengal and marbled swap

Ryos00:
also did a little fiddling with torties, but it didn't fix color changing bug


sedgestripe (approved):
try not to commit pycache files

PR #220, opened 2022-10-27
Add another github actions workflow for building (win/linux/mac)

Dpeta:
Adds build.yml with a few more jobs for building that runs on a push, tested the windows & linux builds and it seems to work :0c

Builds are uploaded as artifacts: [rc2](https://github.com/Dpeta/clangen/actions/runs/3349784242)*

There's also an extra step for adding it to a release when a tag is created, like [here](https://github.com/Dpeta/clangen/releases/tag/v0.1rc2)*, not sure if that's desirable but it seems to be possible at least. If it isn't, the 'Release' step in the jobs should be removed.

Quick thoughts about the build method:
- ~~Decided to run with the ``--onefile`` flag mostly to hide the pygame DLLs, this also means resources + sprites are added to the executable with ``--add-data`` instead of just being dropped in the folder though. If leaving those customizable is preferred it might be better to add a step for manually copying them over instead of using ``--add-data``.~~ N/A, using spec defined settings now
- I *think* the Windows builds ought to run on Windows 7 since the builds use Python 3.8, though win7 users will have to install [universal CRT](https://pyinstaller.org/en/stable/usage.html#windows) if they didn't get it via windows update.
- Linux builds require at least glibc 2.31 or later to be installed because pyinstaller is pyinstaller, probably won't run on old distros. (no hannah montana linux 😿)
~~- Disabled UPX on Windows since some antivirus software doesn't like it, if that's not important re-enabling it might reduce the file size.~~ N/A, using spec defined settings now


Dpeta:
Seems like it's putting the saves in the onefile-generated tmp directory with all the DLLs- Hadn't noticed that >.>
image

Might be better to just run without onefile for now then.

sedgestripe (approved):

Dpeta:
My rationale for disabling UPX again is the usual disadvantages, slower startup time and more anti-virus false-positives, and fairly minor gains (~1.5mb): https://github.com/Dpeta/clangen/actions/runs/3390802856
image

Removing numpy as a dependency also lowers build time by ~3min:
image

Thlumyn (approved):

PR #242, opened 2022-11-06
Apply some optimizations to relationship events and cats

Dpeta:
Hopefully this is equivalent to the previous code, but decently faster. All tests pass:
```
----------------------------------------------------------------------
Ran 60 tests in 0.016s

OK
```

Main changes:
 - Get cat ID via itertools instead of randint.
 - In is_potential_mate, first check if cats have parents/sibs that necessitate the full range of checks.
 - Replace instances of random.randint in relationship_checks.py with random.random.

Minor changes:
 - Replaced handle_relationships loop with itertools.repeat loop, slightly faster since the iterator is not used. 
 - Commented out cat relationships shuffle in handle_relationships.
 - Rewrote get_parents to return parents without intermediately appending them to a list. 
 - Rewrote is_grandparent to iterate through parents instead of checking for length.
 - Replaced some empty list checks via len() with just the list itself, lists already default to False if empty.
 - Replaces some instances of ``x is not None`` with just ``x`` where equivalent.
 - Replaces some instances of ``x is None`` with ``not x`` where equivalent.
 - Added some quick conditions before slow checks.
 - Replaced some conditions with roughly equivalent but faster equivalents.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -843,12 +843,27 @@ class Cat():
         if is_former_mentor and not former_mentor_setting:
             return False
 
-        # check for relation
-        far_related = self.is_grandparent(other_cat) or other_cat.is_grandparent(self)
-        direct_related = self.is_sibling(other_cat) or self.is_parent(other_cat) or other_cat.is_parent(self)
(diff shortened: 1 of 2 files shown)


Lixxis:
I have to check, if random index alone, does the job of not choosing every time the same relationships.

Otherwise, the performance optimization looks good 👍

PR #243, opened 2022-11-06
Revert main to a working version

Dpeta:
Undoes the last three commits in the main branch via cli git revert, should revert it to a working version while preserving history.

If the plan is still to merge dev into main (which I think is a bad idea btw), it's probably better to reset it by one commit instead to synchronize it with dev.


Dpeta:
boomcat

PR #254, opened 2022-11-08
Any Color Heterochromia

sparrows-garden:
Heterochromia functions by having sheets eyes2, eyesextra2, and eyesparalyzed2 which contain the right eye only. If a cat is heterochromic, a second eye color (eye_colour2) is generated that is guaranteed to be different from the first. This allows for any combination of eye colors. They are listed as color1 and color2 on the cat's profile. Inheritance and naming still only use the primary eye color. Chance of heterochromia remains the same.

Sprite indexing for eyes has also been streamlined so now, as long as the width of the sprite sheet stays the same, anyone adding eye colors only has to list all the eye colors in a single array.


Ryos00:
Really loving this idea! However, as it currently is, it causes old saves to break if any of the cats have the old heterochromia. I suggest adding in a way to load the old heterochromia files and update them to the new ones on the next save

PR #271, opened 2022-11-10
Other: Restructure death events

Lixxis:
Some restruction for death events to make a simpler integration of illnesses and injuries.


Dpeta:
I'll skim through it too, please wait with merging 🥹🙏

scripts/events.py
```diff
@@ -817,138 +827,181 @@ def other_interactions(self, cat):
...
+        name = str(cat.name)
+        other_cat = choice(list(Cat.all_cats.values()))
+        countdown = int(len(Cat.all_cats) / 3)
+        while cat == other_cat or other_cat.dead or other_cat.status == 'leader' or other_cat.exiled:
+            other_cat = choice(list(Cat.all_cats.values()))
+            countdown -= 1
+            if countdown <= 0:
+                return
+        other_name = other_cat.name
         clan_has_kits = any(
```
Dpeta, line 857: Could maybe be replaced with the kits lambda from the scars/disaster functions

```python
        alive_kits = list(filter(lambda kitty: (kitty.age == "kitten"
                                                and not kitty.dead
                                                and not kitty.exiled),
                                 Cat.all_cats.values()))
```
Not sure if that's the best way to do this either but it's ~35% faster.

Might be better either way to move this down to under the 1/400 block where it's used though, there's no need to check this for every call rn since iterating through every cat is pretty intense. It takes ~20% of the entire function's runtime.

scripts/events.py
```diff
@@ -817,138 +827,181 @@ def other_interactions(self, cat):
...
+                    ])
+                else:
+                    cause_of_death.extend([
+                        f'{name} fell into a sinkhole and died',
+                        f'{name} fell into a hidden burrow and could not get out',
+                        f'{name} was buried alive when a burrow collapsed on them'
+                    ])
+            elif game.clan.biome == "Forest":
+                if cat.status == 'leader':
+                    cause_of_death.extend([])
```
Dpeta, line 998: Should probably be at least one cause here

scripts/events.py
```diff
@@ -1245,8 +1196,9 @@ def handle_deaths(self, cat):
                 game.cur_events_list.append(
                         choice(cause_of_death) + ' at ' +
                         str(other_cat.moons + 1) + ' moons old')
-
-        elif cat.moons > randint(150, 200):  # extra chance of cat dying to age
+        # extra chance of cat dying to age
+        elif cat.moons > randint(150, 200):
```
Dpeta, line 1200:
```suggestion
        elif cat.moons > int(random.random() * 51) + 150:  # cat.moons > 150 <--> 200
```
~3x faster

scripts/events.py
```diff
@@ -817,138 +827,181 @@ def other_interactions(self, cat):
...
+            game.clan.leader_lives -= 1
+            cat.die()
+            game.cur_events_list.append(
+                choice(cause_of_death) + ' at ' + str(cat.moons + 1) + ' moons old.')
 
-                    
-                if game.clan.all_clans:
+        #Several/All Lives loss
+        elif not int(random.random() * 200) and cat.status == 'leader':  # 1/200
+            triggered_death = True
```
Dpeta, line 906: Death isn't guaranteed at this point, there's still another random on line 907.

Dpeta (commented):
If the intention with triggered_death is to not try to kill a cat twice, it might be easier to just, always run both death functions in order and do a quick check if a cat is dead at the start of the illness function.

scripts/events.py
Lixxis: Moved the function to 1/400 block and replaced it

scripts/events.py
Lixxis: Added a possibility

Lixxis:
@Dpeta 
I thought about it, after your mention, but I think this is a better way for tests.
If I add tests, then I can check if everything went correctly with this cat, when the function said it triggered. Maybe there is a better solution for this, but I would leaf it that way, if it doesn't affect runtime.

Dpeta (commented):
🎉

sedgestripe (approved):

PR #279, opened 2022-11-19
fix: Stop overlapping cats

larkgz:
Cats would overlap on the clan page because they'd get drawn onto the coordinates without checking for overlap. This adds a check for overlap and doesn't place the cat if it sees that there is.


larkgz:
Closing due to incoming AI refactor.

PR #292, opened 2022-12-02
bugfix: mentors not showing current app

ZtheCorgi:
pretty sure i fixed buggy mentors! i just looked at the differences in the code between this and experimental from nov 14th since mentors worked perfectly there


Lixxis (approved):
I downloaded the branch and tested the change of mentor.

PR #294, opened 2022-12-05
bugfix: promote deputy crash

Ryos00:


Ryos00:
no idea why it has all of those other commits on there? seems my branch got de-synced at some point

PR #303, opened 2022-12-14
Rewrote UI to use pygame_gui

keyraven:
The UI has been entirely rewritten, and now uses the pygame_gui graphic library. 
Benefits include, but are not limited to:
 - less CPU usage (no more looping through every cat on every frame)
 - More readable code. The result of all button presses are shown in the screen object. 
 - Smoother scrolling thanks to pygame_gui's text-boxes. 
 - More freedom and versatility in future UI development.  

There are still small adjustments to be made regarding the placement of some text and images. The cat locations on the clan screen are also slightly off. Allegiance formatting needs to be fixed.


Ryos00:
Updated to dev ^^

larkgz:
Should be fixed prior to merge imo: 
- Probably shouldn’t allow non-alphanumeric characters in clan names. Commas cause problems.
- Getting error: `DFScreen has no attribute dead_cats` when navigating to Dark Forest page.
- For some reason the game is moonskipping in intervals greater than one (4–7). I think some things in `one_moon()` are happening where they’re not supposed to. Might have been a bad merge?

Can probably be fixed later:
- Text gets clipped at the bottom in search bar in Cat List and Starclan page.
- It would be nice if text input bars were themed the same colour as the backing image and so the carat was visible.
- When clicking on the text of disabled names (e.g. `Star` of a leader) it causes the text to disappear.
- Console is printing cat IDs on pages. I think this was supposed to be commented out?

keyraven:
Thanks for your review! I have pushed a fix for the dark forest bug, and the game now filters non-alphanumeric characters from entered cat names, clan names, and custom genders.

Ryos00 (approved):

anjubatus (approved):
Good job, I'm liking how the code looks now and the additions to how the screens function, and the text/font stuff looks more proper like this. I do feel like this update made the small text even smaller on profile pages etc. than before which I hope is changed for the sake of people's eyes but that's a minor complaint. The bugs I noticed while playing were fixed.

PR #305, opened 2022-12-15
Code Cleanup and Toggleable Den Labels

keyraven:
I have gone through the code and removed (or commented out) all of the unused screens and functions. I have also removed buttons.py and text.py, since they are no longer needed. 

I fixed a bug where relation values of 1 would result in the respective relation bar being 100% full. Toggle-able den labels have been added back in.


keyraven:
I also added a fix for the issue where the leader of newly created clans would have zero lives.

PR #342, opened 2022-12-23
Bugfix: Fix huge save files (November release)

larkgz:
Lists are mutable, so Relationships were using the same list as default in construction, causing new logs to start with the same list. This caused save files to become unacceptably large due to redundant information. Also, as a temporary fix for large save files, only the most recent log entry will be opened and stored. Otherwise, people would be stuck with 300MB+ save files forever.

If you'd like to save the commit of the original release, please tag it before you merge.


Ryos00 (approved):

DasBlackfur:
Nice!

DasBlackfur (approved):

PR #460, opened 2023-01-10
Add actions workflows for testing encoding and running unittest tests.

Dpeta:
This just runs the unittest scripts pretty plainly, also adds a "test_encoding.py" script with a matching workflow to test if all files in resources are ascii decodable.


Dpeta:
Think it works? https://github.com/Dpeta/clangen-fork/actions/runs/3879749228/jobs/6617226422

PR #477, opened 2023-01-13
Update unit tests

Dpeta:
Kinda complicated by the fact that I'm not sure which errors are just the tests being outdated; For the relationship events I assumed that the new chances were correct.

Currently also:
 - Updates module path in relationship events tests
 - Replaces mock with the standard library version; I don't think we need any specific behavior backported? Unless someone is running these tests with python 3.6 or smthin...


Dpeta:
redoing this

PR #478, opened 2023-01-13
Bugfix/Other: Misc

keyraven:
- Apprentice page buttons on the patrol screen work again
- Failsafe against displaying romantic affection towards relatives. Even if they somehow gained romantic affection, it will not be displayed.
- If first-cousin mates are turned off, romantic affection between first cousins will always be displayed as zero, even if they gained affection while first-cousin mates were enabled.
- Faded siblings and children are added to the cat.sibling and cat.children list upon startup.
- Medicine cat apprentices will no longer retire due to permanent condition.

### DIFF
--- a/scripts/events_module/condition_events.py
+++ b/scripts/events_module/condition_events.py
@@ -676,7 +676,7 @@ class Condition_Events():
         }
 
         if not triggered and not cat.dead and not cat.retired and cat.status not in \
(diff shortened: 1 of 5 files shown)


Lixxis:
Oh noo was too slow :O

The failsave with romantic attraction makes it harder to see the bugs... I would not make them disappear, because the values are there, and the code will interact with them.
This is a bug and we have to fix it.

PR #483, opened 2023-01-15
Convert all utf-8 characters to ascii equivalents and expand encoding test

Dpeta:
- Replaces utf-8 characters with ascii equivalents, mostly “” to "".
- Encoding test now also iterates through python files.
- Encoding test now prints a diff with the non-ascii decodable characters.
- Explicitly defines the encoding as ascii for some files.

I'm not sure if replacing “” in doc strings is actually useful; though it's at least consistent for the test.

### DIFF
--- a/scripts/events_module/generate_events.py
+++ b/scripts/events_module/generate_events.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+# -*- coding: ascii -*-
 try:
     import ujson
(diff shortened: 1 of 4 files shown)


Dpeta:
think when we need to use utf-8 we should just explicitly make all files and opens utf-8 & mybe just rely on the test to find unintentional ascii characters

PR #503, opened 2023-01-18
bugfix: no_herbs tags

ZtheCorgi:
previously, the no_herbs tag was not making the cats gather no herbs on the expected success outcome. now it works as intended


ZtheCorgi:
hold on. needs more testing than i thought

PR #514, opened 2023-01-24
Content: Mediator Role

keyraven:
Adds an optional mediator role. 
Features include:
 - Mediators will occasionally travel to other clans and effect your clans reputation. 
 - Mediators can take apprentices. New apprentices will not choose to be mediators unless you already have a full mediator in the clan. 
 - New setting to allow elders and warriors to choose to become mediators. 
 - Once a moon, you can choose two cats to "mediate" between, which will raise relationship values. Or, if you desire, you can sabotage the relationship, lowering relationship values. 

Others:
Removed "close tab" buttons to allow for more space in the drop-down menus

Note: the "switch to mediator" button is still temporary. Role management is planned to be moved to a new page.


Ryos00 (approved):

Lixxis (changes requested):
I still have to test, but I noticed a possible error while looking into the files:
in thoughts/alive/___.py in the section "dead" the 'mediator apprentice' is missing -> this may lead to a KeyError

keyraven:
Good Catch. I have added mediator apprentice to the dead sections of the thoughts dictionaries.

Lixxis (approved):

Ryos00 (approved):

PR #530, opened 2023-01-26
Other: Double Resolution

keyraven:
The game resolution has been doubled, and now runs in 1600 x 1400


keyraven:
There are still a few things that must be done

PR #596, opened 2023-02-06
Relationship rework

hobohime:
Reworked DE_IN_CREASE to allow for some events to now raise both cats affection that wasn't previously accounted for but were worded in a way that implied it. 

Removed some similarly worded relationship events.

Took out "tidy" "dreamy" and "cowardly" and related relationship events.

Changed a few instances of 'blushing' 'blushed' and 'blushes' to cat flavored text. 

Changed "imagines what their kits with (cat) would look like." to "imagines what their family with (cat) would look like.", "wonders what (cat) thinks about kits.", to "wonders what (cat) thinks about mates.", "keeps shyly glancing over at (cat) as the Clan talks about kits." to "keeps shyly glancing over at (cat) as the Clan talks about mates", to take the focus off of reproduction and onto romance. 

Changed "is asking (cat) how babies are made." to "is asking (cat) where kits come from." and adjusted affection gains for it.

Moved "promises (cat) that they'll be mates when they become warriors." to the apprentice_to_others.json and included  "promises to runaway with (cat) if they no longer want to be a medicine cat." as a comparable event for medcat apprentices from warrior apprentices.


resources/dicts/relationship_events/special_character.json
```diff
@@ -29,7 +29,6 @@
     ],
     "calm": [
       "relaxing with (cat).",
-      "is soothing (cat)'s irrational thoughts.",
```
Lixxis, line 32: Is there a reason why this is removed?

resources/dicts/relationship_events/cat_to_other/apprentice_to_other.json
```diff
@@ -98,7 +99,8 @@
             "wishes they had (cat)'s mentor instead.",
             "offers to help (cat) clean out the elder's bedding.",
             "promises not to tell (cat)'s mentor that they've been sneaking out of camp.",
-            "playfully teases (cat) for missing a catch."
+            "playfully teases (cat) for missing a catch.",
+            "promises (cat) that they'll be mates when they become warriors."
```
Lixxis: I know there is not currently the possibility for exact romantic interaction between apprentices, so I would suggest removing this string. Because not every close apprentices want to have a mate-relationship with each other.

Lixxis (changes requested):
Generally there are multiple changes in the IN and DE creasing files, which are not fitting. 

eg: 
removing **"from romantic increase"** for the cat which has the relationship interaction, like: 
* "thinks that (cat) is really charming."
* "might be reading a little too far into (cat)'s kindness."
* "just noticed how beautiful (cat)'s eyes are."

You should redo these changes and look at the original again.

resources/dicts/relationship_events/DE_IN_CREASE/INCREASE_HIGH.json
```diff
@@ -1,58 +1,43 @@
...
-            "wonders what (cat) thinks about kits.",
-            "thinks (cat) has a lovely sense of humor.",
             "is the only cat who can make (cat) laugh.",
             "asks (cat) to help them freshen up on some battle moves, and compliments them constantly during it.",
             "asks (cat) to go fishing with them, but they keep on getting distracted by each other.",
-            "gives (cat) a beautiful feather they found, and blushes when thanked.",
-            "gives (cat) a wonderfully colored pebble they found, and blushes when thanked.",
-            "gives (cat) a circular Twoleg object they found, and blushes when thanked.",
-            "compliments (cat) on how well groomed their pelt is today.",
+            "gives (cat) a beautiful feather they found, and hides behind their wiskers when thanked.",
```
Ryos00: *whiskers

resources/dicts/relationship_events/special_character.json
hobohime: Yes, I forgot include the reason why on my PR. This was removed because the calm personality relationship events are heavily loaded to be emotional labor with this and the 'is helping (cat) calm down." 

I made the decision to remove it and leave the less emotionally charged event in, but I can re-add it if needed.

scribblecrumb: I think it would be better to add, rather than remove.  Perhaps coming up with some less emotional calm events would be preferable.

hobohime: I'll add that back in right now then.

Ryos00 (approved):

resources/dicts/relationship_events/DE_IN_CREASE/DECREASE_HIGH.json
```diff
@@ -20,39 +21,28 @@
         ],
         "like": [
             "is telling jokes about (cat).",
-            "whines about (cat).",
-            "is tired from (cat) putting them on so many patrols.",
-            "is bossing (cat) around.",
             "started a fight with (cat).",
             "has a fight with (cat) about what's right.",
+            "is not backing down in an argument with (cat).",
+            "scorns (apprentice) for not catching enough prey.",
```
Lixxis: I'm not sure if someone changed the code, if not (apprentice) has to be (cat)

Lixxis (changes requested):
Only one thing I saw, otherwise I think it is alright ! If this is changed or correct @Ryos00 you can merge it.

resources/dicts/relationship_events/DE_IN_CREASE/DECREASE_HIGH.json
hobohime: I'll fix all instances of this right now

Ryos00 (approved):

Ryos00 (approved):

PR #645, opened 2023-02-13
Bugfixes/Tweaks

keyraven:
- Slightly increase mated kit-chance. 
 - Prevent unmated cat from rolling the primary kit roll if unmated kits are turned off. 
 - De-coupled sprite from rank. This should fix the issue of long-haired lost cats having the kit sprite if their age and rank do not match.  
 - Moved some "already pregnant" checks for clarity.

### DIFF
--- a/scripts/events_module/relation_events.py
+++ b/scripts/events_module/relation_events.py
@@ -170,7 +170,6 @@ class Relation_Events():
         else:
             chance = 100
 
-        print(str(cat.name), chance)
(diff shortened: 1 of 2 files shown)


Lixxis (changes requested):
I think there went something wrong, as a warrior it was right, but not as an elder: 
image

keyraven:
@Lixxis  - Good catch! That issue has been fixed

PR #664, opened 2023-02-16
QOL: "would you like to save" on window close

lillithrosepup:
very bad implemintation but eh


lillithrosepup:
oops i didnt mean to pr, accidentally left in game_config.json stuff

PR #669, opened 2023-02-17
Pyinstaller Version File

lillithrosepup:
Adds a commit.txt file to pyinstaller builds containing the commit hash from the push that triggered it, eg: `42b3eed45eca0e738ed7bfe94cad3ab658017e9e`. This can be used to help debugging, as we can see the state of the code  as it is on the player's end


Lixxis (commented):
Not sure if this is really needed. 
We already see the commit, that triggered the action in the action overview:
image

Ryos00 (approved):

PR #690, opened 2023-02-21
Added "null" a lot

hobohime:
added null to every patrol the game expects an outcome in the any forest .json in an attempt to have the patrols return freshkill and not crash the game. Added "medium_prey" to some patrols that used to give "small_prey".


scribblecrumb (changes requested):
I'm not entirely sure this fixes the problem you're facing.  Which bug is this that you're referencing in the description?  I feel like this might be better addressed at the root of the problem, this seems like a bandaid fix to me.

PR #700, opened 2023-02-22
Bugfix: Backstory_contraint was out of scope

lillithrosepup:


scripts/events_module/generate_events.py
```diff
@@ -146,7 +146,7 @@ def filter_possible_events(self, possible_events, cat, other_cat, war, enemy_cla
             if "other_cat" in event.tags and not other_cat:
                 continue
 
-            if "backstory_constraint" and cat.backstory not in backstory_constraint:
+            if "backstory_constraint" and cat.backstory not in event.backstory_constraint:
```
Lixxis: Can you change it to following?
`if event.backstory_constraint and cat.backstory not in event.backstory_constraint`

The string if condition does not make any sense otherwise

Lixxis (changes requested):

Lixxis (approved):

PR #709, opened 2023-02-22
Bugfix: Ceremony Tags

keyraven:
- Fixed no-mentor type ceremonies showing up for new medicine cats who have mentors. 
- Added new "no_previous_mentor" tag for ceremonies where the new cat has never had a mentor. 
- Updated count_patrols.py to grab all patrol files automatically, without having to list them.


keyraven:
I want to add more more tag.

PR #757, opened 2023-02-26
Bugfixes/Unmated Kit Chance

keyraven:
- Fixed the fit cancel chance to properly increase with population for unmated cats. It also also no longer linear. 
- Slightly decreased primary kit roll chance for unmated cats. 
- Mediator apprentices will now properly gain their skill when aging up. 
- The delete button on the switch clan screen will now scale in full-screen. 
- Fixed senior cats sometimes dying twice in one moon. 
- Typos and some awkward wording.

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -1431,13 +1431,11 @@ class Events():
             self.death_events.handle_deaths(cat, other_cat, self.at_war, self.enemy_clan, alive_kits)
             triggered_death = True
             return triggered_death
-            return triggered_death
 
         # chance to die of old age
         if cat.moons > int(random.random() * game.config["death_related"]["old_age_death_chance"]) + game.config["death_related"]["old_age_death_start"] and not triggered_death:  # cat.moons > 150 <--> 200
             self.death_events.handle_deaths(cat, other_cat, self.at_war, self.enemy_clan, alive_kits)
             triggered_death = True
-            return triggered_death
             return triggered_death
 
         # classic death chance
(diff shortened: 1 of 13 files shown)


scripts/events.py
```diff
@@ -1436,8 +1438,14 @@ def handle_injuries_or_general_death(self, cat):
...
 
+        # classic death chance
+        if game.clan.game_mode == "classic" and not int(random.random() * 500):  # 1/500
+            self.death_events.handle_deaths(cat, other_cat, self.at_war, self.enemy_clan, alive_kits)
+            triggered_death = True
+            return triggered_death
+
         # disaster death chance
-        if game.settings.get('disasters') and not triggered_death:
+        if game.settings.get('disasters'):
```
scribblecrumb, line 1444: What's the reason for this change?  If I recall correctly, `triggered_death` needs to be checked in order to ensure a cat doesn't die and then also trigger a disaster

scribblecrumb (changes requested):

scripts/events.py
keyraven: It was removed because I changed the function to return early if it triggered a death, so the check for "triggered death" was no longer necessary.  I added the early returns to prevent a cat from dying twice as a senior. If you look above, the disaster roll with not be reached if an earlier death chance is triggered.

scribblecrumb: Gotcha! makes sense

scribblecrumb (approved):

PR #771, opened 2023-02-28
more typo fixes

Clumxy:


resources/dicts/thoughts/other.json
```diff
@@ -213,7 +213,7 @@
             "Hisses at a missed catch",
             "Is hoping they can find something to fill their belly tonight",
             "Licks at an old, aching scar in an attempt to soothe it",
-            "Was almost found by a c_n patrol",
+            "Was almost found by an c_n patrol",
```
Ryos00: this should actually always be 'a' because of our little system that will change it to 'an' if the clan name starts with a vowel

resources/dicts/thoughts/other.json
```diff
@@ -122,7 +122,7 @@
             "Saw a helpless cat get taken in a metal box by Twolegs",
             "Sniffed around a known Clan territory curiously ",
             "Chow down on some food a Twolegs left out for them",
-            "Envy the strong kinship the Clan cats has",
+            "Envy the strong kinship the Clan cats have",
```
Ryos00: "Envy" should probably be changed to "Envies" as well

resources/dicts/thoughts/other.json
lillithrosepup: oh wow i had no clue that was there. good to know!

resources/dicts/thoughts/other.json
Clumxy: i think i fixed it

PR #783, opened 2023-02-28
Case check on same clan name, and no special chars

lillithrosepup:


scripts/screens/clan_creation_screens.py
```diff
@@ -152,10 +152,11 @@ def handle_name_clan_event(self, event):
                 self.elements["error"].set_text("Your Clan's name cannot be empty")
                 self.elements["error"].show()
                 return
-            if new_name in game.switches["clan_list"]:
-                self.elements["error"].set_text("A clan with that name already exists.")
-                self.elements["error"].show()
-                return
+            for clan in game.switches["clan_list"]:
+                if new_name.lower() in clan.lower():
```
Dpeta:
```suggestion
                if new_name.casefold() in clan.casefold():
```
Probably doesn't matter in practice but preferred for caseless matching.

Dpeta (commented):
🐈✅

PR #793, opened 2023-03-01
med_cat retagging

hobohime:
Deleted all instances of med_cat from the beach .json where it wasn't implied a full medcat had to be there


resources/dicts/patrols/beach/med/leaf-bare.json
```diff
@@ -53,7 +53,7 @@
     "patrol_id": "bch_med_gatheringblackberry_leaf-bare1",
     "biome": "beach",
     "season": "leaf-bare",
-    "tags": ["med_only", "cold_injury", "scar", "FROSTMITT", "herb", "blackberry", "med_cat"],
+    "tags": ["cold_injury", "scar", "FROSTMITT", "herb", "blackberry", "med_cat"],
```
Ryos00: deleted med_only instead of med_cat ^^"

PR #844, opened 2023-03-04
two new med app/war app patrols!

capillarys:
don't think there were any before, and we need at least Some in the game T.T


resources/dicts/patrols/general/medcat.json
```diff
@@ -1004,5 +1004,54 @@
     "max_cats": 2,
     "antagonize_text": null,
     "antagonize_fail_text": null
+},
+{
+     "patrol_id": "gen_med_warappmedapp_herbcontest1",
+        "biome": "Any",
+        "season": "Any",
+        "tags": ["apprentice", "two_apprentices", "warrior_app", "many_herbs1", "random_herbs", "respect", "comfort", "dislike", "jealousy"],
+        "intro_text": "In order to spice up their patrol, p_l challenges app1 to a herb-gathering competition.",
```
scribblecrumb: in both these patrols, app1 should be changed to app2.  on apprentice only patrols, p_l will always double as app1, so you shouldn't use the app1 abbreviation at all.

capillarys: alright that makes sense! sorry i missed that

PR #862, opened 2023-03-06
Fixed unittests, maybe

hobohime:
whatever I did had "testSaves" fill up with two test files, and suddenly it now passes unittests. I don't think I deleted anything permanently.


hobohime:
Okay, a few hours went by and I now know what's going on, but not how to fix it. The git submodule that threw the critical error was made to populate the testSave area that I somehow managed to get the files for on this PR. When the git submodule broke, it broke the ability for the unittest to get units to test, therefore causing all unit tests to fail. 

The solution is to make a new submodule to do the job of the old one, and replace where its pulling the test saves from, to ensure that the unit tests are able to run.

I'm keeping this PR open until someone with more experience then I can verify that this is the case. This PR does NOT address the underlying issue, but will be useful to diagnose the problem.

PR #864, opened 2023-03-06
Content and Feature: Shipwreck and Season Selection

scribblecrumb:
- the Shipwreck map is now available for the Beach biome! It has it's own special cat positions layout.
- During clan creation you can now choose which season you start in as well as preview all the camp seasonal variations
- starting_season is now save in the clan save file, old saves will default to Newleaf
- New utility function added called get_current_season() to handle math for finding the current season. It's only used in the save loading and timeskip season progression, but I figured it might make for a valuable utility in the future
- Adjusted the wording on the mate selection screen blurb bc it didn't make a ton of sense, adding some vague explanations for compatibility lines and heart symbols


scribblecrumb:
[Allow Player to Choose Starting Season](https://trello.com/c/UCrdDGuO/48-allow-player-to-choose-starting-season)

PR #884, opened 2023-03-07
Fixing plurality

lillithrosepup:
Clumping this in a big pr to add as i go bc i know yall are asleep


lillithrosepup:
nvm i lied

PR #902, opened 2023-03-08
Content: Pregnancy is now a condition

scribblecrumb:
- Pregnancy added as a condition, funnily enough it's considered an injury within the code lmao
- As such, pregnancy is now noted in the cat's profile.  While I was there I also fixed the visual bug where congenital conditions displayed their moons_with total as well as the 'born with' message.
- Labor mortality now scales with the cat age, older cats will be more likely to die while kitting
- Moved all pregnancy/birth related strings into a json
- Pregnancy "severity" is now randomly determined at the first moon.  It can start as minor and move to major in the second moon, or it can be major from the start.  Minor pregnancies are more likely.  The only thing this effects is the cat's ability to work.  This is indicated with a moon event.
- Cats will now guess a more general litter size.  They can guess that they have a 'small' or 'large' litter, they may also just say they're unsure of the size.  This guess has a chance to be incorrect, but the majority of the time it will be correct.


Lixxis (changes requested):
This is the traceback when in classic mode someone is going to have kits.
image

Lixxis:
And because I just saw it, the condition is already major in the first moon of pregnancy, not sure if this is intended or not 👍

scribblecrumb:
> This is the traceback when in classic mode someone is going to have kits. image

Fixed!

As for the major pregnancy in the first moon, yes that's intentional.  The pregnancy severity is randomly determined in the 1st moon, so it can start as major or it can start as minor as progress to major.  The chance is weighted towards starting as minor, though.

scribblecrumb:
[Pregnancy As A Condition](https://trello.com/c/grGkmIWQ/50-pregnancy-as-a-condition)

PR #905, opened 2023-03-08
User defined names additions

lillithrosepup:
users can remove prefixes/suffixes by starting it with -

users can add special suffixes in the format:
<status>:<prefix>
or remove it with
-<status>

more info in docstring


lillithrosepup:
closing in favor of new fork w less commits

PR #909, opened 2023-03-08
Content: Affair Changes, new pregnancy strings

keyraven:
- Changed love affairs to consisted the highest romantic relation, excluding the cat's mate, as an affair partner. 
- Unmated cat can now have love affairs with their highest romantic relation. If that highest romantic relation has a mate, the chance for this to happen will be reduced. 
- Unmated cats now have a much higher chance for random affairs. 
- Love affair chance now depends on the average of the love between the mate/cat and affair/cat. 
- Kit cancel chance now uses the average love, comfort and trust between mated pairs. 
- Added birth strings for two unmated parents.


Ryos00 (approved):

keyraven:
Fixed a bug where mates cats wouldn't have kittens with their mate

PR #919, opened 2023-03-09
Fix issue with ev_text_adjust

lillithrosepup:


scripts/utility.py
```diff
@@ -445,7 +445,11 @@ def event_text_adjust(Cat,
     if "acc_singular" in adjust_text:
         adjust_text = adjust_text.replace("acc_singular", str(plural_acc_names(cat.accessory, False, True)))
 
-    adjust_text = adjust_text.replace("c_n", str(clan.name) + "Clan")
```
keyraven, line 448: I suggest removing game.clan as the default value for the clan parameter. Python default parameter are set by-value at function declaration, so it will always be none.

keyraven: I suggest using clan = None as the default parameter. Everything else can be kept the same

PR #944, opened 2023-03-11
Refactor settings menu

lillithrosepup:


Ryos00 (approved):

scribblecrumb (changes requested):
image

hmmm i got a crash when i tried to toggle one of the relation settings

lillithrosepup:
Fixed the link, and added button

lillithrosepup:
image

scribblecrumb (changes requested):
image
Seems there's a blank tool tip for the hotkey toggle, but tbh we should probably just remove that toggle completely bc it does nothing at the moment.

image
it also seems that the other buttons that are in the scrolling container, but not visible, can still be interacted with. I'm not sure if this is a new bug or an old one, but can you look in to it?

image
also got this crash when I tried to close the game without saving my settings (should this have triggered your error window or is that only on certain errors?)

lillithrosepup:
I removed the hotkey setting, and am looking into the scroll issue. My new box is only triggered by what triggered the old box via
```py
except FileNotFoundError as e:
    game.switches['error_message'] = 'Can\'t find clan_cats.json!'
    game.switches['traceback'] = e
    raise
```
(example error)

PR #945, opened 2023-03-11
Backend: create_new_cat function

scribblecrumb:
I've condensed the two create_new_cat() functions into one function in utility.py! I've also rewritten quite a bit of the function to make it significantly less spaghetti and more flexible with plenty of parameters to make specific cats with.

In the process, the add_new_cats() function in patrol.py has been massively streamlined and should be much easier to read and understand.  I do think there's some tweaking to be done with how new_cat tagging is handled, as it seems to rely on a mix of tags and patrol_id filtering atm, which isn't ideal.  We should retag relevant patrols and stop using patrol IDs to filter, as that's just hard to keep track of in this instance.  However, it should be much easier to add new specific new_cat tags with how I've rewritten the function.

I've added some documentation and commenting around these functions to hopefully make it easier to navigate in the future.


scribblecrumb:
[Move create_new_cats function to utility.py](https://trello.com/c/RuaHKP9Z/37-move-createnewcats-function-to-utilitypy)

PR #946, opened 2023-03-11
Add multiple names to each colour prefix category + alphabetize and sort them

SunlitFable:


scribblecrumb (changes requested):
A lot of these names seem to be ones that were previously removed for various reasons, it would be good to present the new name list in one of the contributor channels so that it can be discussed!

hobohime:
A lot of the names added are names planned for biomes, while they are good names, part of the plan ATM is to have some names only exist in biomes, such as "sheep" and "anemone".

PR #956, opened 2023-03-12
Collar Fixes and Adding Collar to Med Cat Patrol

Ronfaure:
Fixing some of my previous patrols that added two of the word 'collar', as well as adding it to the medicine cat misc events.


Thlumyn:
@Ronfaure Looks like theres been some changes to the medicine.json file, can you resolve those conflicts on your side?

PR #960, opened 2023-03-12
Fix incorrect version number comparison

archanyhm:
This caused version numbers to be incorrectly determined when running on source in case git is unavailable


scripts/version.py
```diff
@@ -16,7 +20,8 @@ def get_version_info():
             is_source_build = True
             try:
                 version_number = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip()
-            except:
+            except Exception as e:
+                logger.exception(f"Git CLI invocation failed: {str(e)}")
```
larkgz: Calling `logger.exception` actually automatically prints the exception so I don’t think you need to manually print `e`

PR #963, opened 2023-03-13
Implement a stream duplexer for stdout/stderr

archanyhm:
This is done to make the console output made through print calls accessible on binary releases, particularly for nightly builds.

In other words:
Instead of just printing the output of `print`-calls to the console, they're additionally being logged to a file in the game's log folder.


Lixxis (changes requested):
Just a short question:
do the logs take an unlimited amount of space, or do they make a queue when saving and only X amount of logs will be stored?

archanyhm:
> Just a short question: do the logs take an unlimited amount of space, or do they make a queue when saving and only X amount of logs will be stored?

As of now, there's no "stale log purging" mechanism in place; and neither is there a mechanism in place that imposes a size limit on the log files. Theoretically they could take up "unlimited space". That problem already exists for the regular clangen log files though — and there has been a rudimentary discussion about implementing log rotation.

Something like "only keep the last 10 logs of a kind".
Does that work?

Lixxis:
> ###
> [Mar 13, 2023](#issuecomment-1466740566)

Ok, wasn't there when the other logging was added and didn't know it is already in discussion 👍 
Do you want to make this change in this PR, or do you going to open another one for those changes?

archanyhm:
> > [Mar 13, 2023](#issuecomment-1466740566)
> 
> Ok, wasn't there when the other logging was added and didn't know it is already in discussion 👍 Do you want to make this change in this PR, or do you going to open another one for those changes?

I'd implement that in a separate PR later today.

Lixxis (approved):

PR #965, opened 2023-03-13
Tweaks: Apprentice Graduation Balancing

keyraven:
- Passive EX now works. 
- Apprentices now receive patrol EX from a fixed range. They receive this EX on both success and failure. 
- Apprentice now receive EX on mediate fail. 
- Adjusted mediate EX
- Adjusted passive EX range
- Sick apprentices now have a 33% chance to receive passive EX. 
- Raised minimum graduation age to 10 moons. 
- Changed the wording of the first cousin setting back to it's previous text. The changed text was not properly accurate. 

Hopefully this should help a bit with balancing.

### DIFF
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -209,13 +209,13 @@
 		"comment": "Set this to -1 for it to be infinite"
 	},
 	"graduation": {
-		"base_app_timeskip_ex": [[3, 4, 5, 6, 7, 8, 9], [5, 10, 15, 40, 15, 10, 5]],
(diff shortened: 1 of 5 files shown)


Ryos00 (approved):

Lixxis (approved):
Do we still need the prints?

Lixxis:
@keyraven Sorry not sure if you see my normal comment

keyraven:
@Lixxis  Yes, I think the prints are still necessary.  I want to continue to monitor balancing, it would be helpful for beta testers to be able to see where apprentices got their EX.  I believe this will be the last balancing patch needed, but the numbers are still based off of a simpler model, and I haven't done much in-game balance testing.

PR #988, opened 2023-03-14
Added 1 para patrol for ryos

redhairedl:
An apprentice only patrol with chancce of paralysis


resources/dicts/patrols/general/training.json
```diff
@@ -1094,4 +1094,45 @@
     "antagonize_text": null,
     "antagonize_fail_text": null
 }
-]
+{
+    "patrol_id": "gen_training_twoapps_forbiddentunnels1",
+    "biome": "Any",
+    "season": "Any",
+    "tags": ["two_apprentices", "rel_two_apps", "big_change", "trust",  "scar", "injury", "broken back", "app_only", “training”],
```
scribblecrumb: if you have the "scar" tag then you need to also include the name of the scar that this patrol should be giving on classic mode.

resources/dicts/patrols/general/training.json
```diff
@@ -1094,4 +1094,45 @@
...
+    "intro_text": "app1 and app2 while exploring the territory come across the forbidden tunnels.",
+    "decline_text": "They don't want to get into trouble and they have a lot more territory to explore anyway.",
+    "chance_of_success": 50,
+    "exp": 10,
+    "success_text": [
+            "app1 and app2 have a wonderful time exploring the forbidden tunnels.
+They return to camp sharing a mysterious smile, swearing not to speak of this to anyone.",
+            "app1 expected a little hesitation but app2 defies that expectation and darts right into the tunnels playfully batting at them as they do so.
+They play for it seems like forever and they make up a lie to tell their mentors on their way back to camp."
```
scribblecrumb: "They play for what" would read better than "They play for it"

resources/dicts/patrols/general/training.json
```diff
@@ -1094,4 +1094,45 @@
...
+            null,
+            "s_c boasts to app2 that they go into the tunnels allllllll the time, their head and tail held high as they do so.
+app2 seems to be unconvinced so s_c crawls into the narrow opening of the tunnel and exclaims that there is nothing to it.
+Just at that moment the tunnel makes an ominous rumble and collapses.
+They both scream in absolute terror and app2 runs back to camp praying to StarClan that they can find help and soon, as they return with the help they can only hear the heartbroken wail of s_c that they can’t feel their legs anymore. ."
+     ],
+    "win_skills": null,
+    "win_trait": null,
+    "fail_skills": null,
+    "fail_trait": ["adventurous", "bold", "confident", "daring"],
```
scribblecrumb: maybe add "troublesome", "playful", and "childish" to the fail traits?

resources/dicts/patrols/general/training.json
```diff
@@ -1094,4 +1094,45 @@
...
+    "min_cats": 2,
+    "max_cats": 2,
+    "antagonize_text": null,
+    "antagonize_fail_text": null,
+    "history_text": [
+            "This cat their back was broken after attempting to enter the forbidden tunnels."
```
scribblecrumb: You've got some typos here, it should be "This cat's back was broken--"

scribblecrumb (changes requested):
Remember to keep an eye out for smart quotes, you can see how a bunch of text is red, that means that there's a formatting error.  It looks like the "training" tag has smart quotes around it.  I'm not sure what's up with the outcome texts, but I do wonder if it's due to them having some text breaks? They can just be written all on one line, jsons don't like having one string broken into two lines.
If you're writing these in a google doc then please make sure to turn off smart quotes.  I'd also suggest just writing it strait into the code via an IDE, then you can avoid these problems completely.

PR #1026, opened 2023-03-18
typos + added new ceremonies

capillarys:
had this written yonks ago but just never PRed, and now that branch won't merge whoops so here goes


capillarys:
nvm it got closed because i changed the name of my branch nooooo

PR #1047, opened 2023-03-21
Bougfix

lillithrosepup:


scripts/screens/cat_screens.py
```diff
@@ -1739,7 +1739,7 @@ def update_disabled_buttons_and_text(self):
             # Button to prevent kits:
             if self.toggle_kits:
                 self.toggle_kits.kill()
-            if self.the_cat.age in ['young adult', 'adult', 'senior adult', 'elder'] and not self.the_cat.dead:
+            if self.the_cat.age in ['young adult', 'adult', 'senior adult', 'senior', 'elder'] and not self.the_cat.dead:
```
scribblecrumb: You can remove 'elder' from this list, as 'senior' has replaced it

scribblecrumb (changes requested):

scripts/screens/cat_screens.py
lillithrosepup: ahh i didnt remove just incase there as a reason for it to be there, thanks

PR #1064, opened 2023-03-21
Updated the name lists!

hobohime:
I took out some duplicate names and kept others as some of them are canon names and we keep cannon names in the normal pool until someone decides differently.

Didn't touch the names for wetlands and desert bc they're not implemented yet, there is no reason to limit names that cannot be gotten in game yet.

### DIFF
--- a/resources/dicts/names/names.json
+++ b/resources/dicts/names/names.json
@@ -73,16 +73,17 @@
   "normal_prefixes": [
     "Adder", "Alder", "Ant", "Antler", "Apple", "Apricot", "Arc", "Arch", "Aspen", "Aster",
     "Badger", "Barley", "Basil", "Bass", "Bay", "Bayou", "Beam", "Bear", "Beaver", "Bee", "Beech", "Beetle", "Berry",
-    "Big", "Birch", "Bird", "Bite", "Bitter", "Bittern", "Blizzard", "Bloom", "Bud",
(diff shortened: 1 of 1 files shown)


capillarys:
psst i thought duplicates were meant to make a name more common? or do you think those names weren't meant to be common (just out of curiosity)

lillithrosepup:
> psst i thought duplicates were meant to make a name more common? or do you think those names weren't meant to be common (just out of curiosity)

It does make it more common, but i asked ryos a bit ago and they are p sure its an accident that the lines were duped.

resources/dicts/names/names.json
```diff
@@ -73,16 +73,17 @@
   "normal_prefixes": [
     "Adder", "Alder", "Ant", "Antler", "Apple", "Apricot", "Arc", "Arch", "Aspen", "Aster",
     "Badger", "Barley", "Basil", "Bass", "Bay", "Bayou", "Beam", "Bear", "Beaver", "Bee", "Beech", "Beetle", "Berry",
-    "Big", "Birch", "Bird", "Bite", "Bitter", "Bittern", "Blizzard", "Bloom", "Bud",
```
Ryos00, line 76: I'd probably leave "Birch" in here since it's canon and used multiple times

Ryos00 (changes requested):

Ryos00 (approved):

capillarys:
ahhh big excite ty hobo for names-ing and ty ryos for ze merge!! i shall now try to get some drunkelskogs in

PR #1067, opened 2023-03-22
Allow locking the season to whatever the clan started with

lillithrosepup:


resources/game_config.json
```diff
@@ -249,6 +249,10 @@
 	},
 
 
+	"lock_season": false, 
+	"_": "Forces the season to be locked at the clan's starting_season",
```
Thlumyn: can you change this to say comment instead of _ to be in line with how the comments on the other config options are?

Thlumyn (changes requested):

Lixxis (approved):

Ryos00 (approved):

PR #1070, opened 2023-03-22
Suffix overwrite

lillithrosepup:
Changes the "edit internal suffix" box to where when its enabled, the special suffix is completely hidden
image
image
image
image


lillithrosepup:
ignore the todo: test, i caved in and loaded up a codespace

PR #1089, opened 2023-03-23
Always show cat's mate in mediation screen

lillithrosepup:


keyraven:
This was not originally put in since it would often cause text to overflow when displayed both the name of a cats mate, and their relatedness to the target cat.

scribblecrumb (changes requested):
Could implement the name shortening (Bubblingnoise to Bubblingn...) code, though it would look a little funky to see Bubblingn...'s mate.  Not the worst thing though.  A different idea could be a tooltip that appears when hovering over the mate text that lists the mate's name?

lillithrosepup:
what should the limit be? 9?

scribblecrumb:
eh, whatever looks good I guess

keyraven:
I think the limit needs to be small enough so that the message never goes goes beyond a single line. Otherwise, you will get overflowing with the relation information. However, that might require a really small limit, like 3 or 4,

scribblecrumb:
It might be better to show the mates name upon hovering over the mate text!

Lixxis:
@scribblecrumb 
There were some commits, does the issues you address still exist?

lillithrosepup:
Closing as inactive

PR #1092, opened 2023-03-23
Bugfix/Content: Track Dead Mates, Dead Cat Mates

keyraven:
- When a cat's mate dies and they move on, that mate will now be tracked as a previous mate. 
- Moved the "moving on" roll outside the relationship loop, so it only rolls once. 
- You can now give starclan cats mates (only other dead cats). 
- Mediators can only travel to other clans if they can work. 
- Dead cat relationships files are now deleted. 
- Cats will now move on from mates if their mate is lost or exiled.
- When a cat dies, their relationships are now cleared (set to an empty dictionary). Note: There are some situations where a relationship for a dead cat is created. Mostly when creating a opposite relationship.  That is probably fine for now. 
- Fixed the issue of lost cats from previous saves "returning" to a new clan.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -525,7 +525,6 @@ class Cat():
         self.injuries.clear()
         self.illnesses.clear()
 
-
         # Deal with leader death
         text = ""
         if self.status == 'leader':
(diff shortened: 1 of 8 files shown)


keyraven:
@Lixxis I have tried to mess with relationships as little as possible, but there is some stuff related to relationships here. Can you look everything over to check that I didn't do anything stupid?

Lixxis (approved):

PR #1101, opened 2023-03-24
Maybe fixed an issue where s_c is r_c

hobohime:
It had spaces rather then underscores. It's a pretty rare patrol to pull with my rng so I want to get the beta testers on it asap.


scribblecrumb:
hmm, good catch but that wouldn't have any affect on the s_c and r_c cat determination.  That tag only controls relationships

PR #1117, opened 2023-03-26
Content/Bugfixes: Apprentice graduation setting, faded cats, loading checks

keyraven:
- Faded cats are no longer considered their own siblings. 
- Added a setting to force apprentice graduation at 12 moons. 
- Adjusted classic mode apprentice EXP balance. Gave a buff to passive EXP, and removed apprentice patrol EX. 
- Trait and skill updating now occurs at graduation, regardless of age. 
- Added save integrity check, that runs after all cats have been loaded in.  Right now, all it does it check and correct one-side mate relationships (cat.mate = MATE_ID, but mate.mate = None), but more can be added. Added a setting in game_config to turn off these checks, if someone wants to save the load time.


keyraven:
Note: unit test is failing since the skill update has been moved out of update_status and into ceremony()

keyraven:
Unit tests have been fixed!

PR #1137, opened 2023-03-28
aug10 save file converter

lillithrosepup:
ported from web ver !

### DIFF
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -759,7 +759,7 @@ class Clan():
             self.load_clan_txt()
         else:
             game.switches[
-                'error_message'] = "There was an error loading the clan.json"
+                'error_message'] = "There was an error finding the clan.json"
             
         self.load_clan_settings()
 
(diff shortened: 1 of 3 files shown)


Ryos00 (approved):

Lixxis (commented):
I think we should wait for this, because of the feature freeze. 
It might have more impact than we thought, I would suggest adding this after the new update.

lillithrosepup:
imo, this would be better to include in the update bc a bunch of people will see the new update after not playing for a while and it would help lessen the load on techsupport from all the people moving over. and even if it doesnt work somehow, i made sure it doesnt delete the old csv, so they can still use the web convertor

Lixxis:
I see no reason, why they are going to be more ppl playing this update than any other update before. 
I'm not worried about that, the files are going to be ruined I'm more afraid, that this load would result into a buggy clan with errors. The next update is soon and in my opinion such a conversion would need more testing.

Lixxis:
@ImLvna 
If you fix the merge conflict, we can try to merge it

larkgz:
Some minor corrections/additions for the conversion:
- rather than `status`, `age` should be be the one set to `senior`
- if a cat has the trait `clever`, it should be converted to `wise` due to the clever trait being removed
- `eye_colour2` is typically `null` if the cat does not have heterochromia

lillithrosepup:
closing as it seems it is already ingame?

PR #1140, opened 2023-03-29
Bugfixes/Small Content: Role Screen, Alignment Issues, ext

keyraven:
- The "small content" is the role management screen icons. I kinda forgot those were completed awhile ago (my bad)
- Newborns now get the kitten blurb on the role management screen. 
- Fixed deputies who return on the same moon as becoming lost being listed as deputy but having the "warrior" status. 
- Fixed mentor box on the choose mentor screen often overflowing (it still may happen, but a lot less often)
- Replaced removed traits in ceremony_master.json.  Re-tagged ceremonies that were labels to only occur for removed traits. 
- Fixed a crash related to cats joining at timeskip. 
- Removed game.rank_changed_timeskip.  Cat are now resorted at the end of every timeskip. This was basically already happening, and will fix some issues with cat becoming out of order. 
- On the mediation screen, cat traits will be shortened if they are too longer, to prevent text overlapping.


scripts/events.py
```diff
@@ -217,10 +217,9 @@ def one_moon(self):
         self.check_and_promote_leader()
         self.check_and_promote_deputy()
 
-        # Resort if needed
-        if game.ranks_changed_timeskip and game.sort_type == "rank":
-            game.ranks_changed_timeskip = False
-            Cat.sort_cats()
+
+        #Resort
+        Cat.sort_cats()
```
Lixxis: Wouldn't this result in a lag if we have many cats? It might only result in unnecessary workload, is it really necessary to resort the cat every moon?

scripts/screens/cat_screens.py
```diff
@@ -2513,14 +2513,22 @@ def update_selected_cat(self):
...
+            "mediator apprentice": "resources/images/mediator_app_icon.png",
+            "warrior": "resources/images/warrior_icon.png",
+            "apprentice": "resources/images/warrior_app_icon.png",
+            "kitten": "resources/images/kit_icon.png",
+            "newborn": "resources/images/kit_icon.png",
+            "elder": "resources/images/elder_icon.png",
+        }
+        
+        if self.the_cat.status in paths:
+            icon_path = paths[self.the_cat.status]
```
Lixxis: For further improvement, I would suggest making this change:

```
        main_dir = "resources/images/"
        paths = {
            "leader": "leader_icon.png",
            "deputy": "deputy_icon.png",
            "medicine cat": "medic_icon.png",
            "medicine cat apprentice": "medic_app_icon.png",
            "mediator": "mediator_icon.png",
            "mediator apprentice": "mediator_app_icon.png",
            "warrior": "warrior_icon.png",
            "apprentice": "warrior_app_icon.png",
            "kitten": "kit_icon.png",
            "newborn": "kit_icon.png",
            "elder": "elder_icon.png",
        }

        if self.the_cat.status in paths:
            icon_path = os.path.join(main_dir,paths[self.the_cat.status])
```

Lixxis (changes requested):

scripts/screens/relation_screens.py
```diff
@@ -2771,7 +2771,11 @@ def draw_info_block(self, cat, starting_pos: tuple):
...
             col1 += " moons"
-        col1 += "\n" + cat.trait
+        if len(cat.trait) > 11:
+            _t = cat.trait[:10] + ".."
+        else:
+            _t = cat.trait
+        col1 += "\n" + _t
         self.selected_cat_elements["col1" + tag] = pygame_gui.elements.UITextBox(col1,
                                                                                  scale(pygame.Rect((x + 42, y + 252),
                                                                                                    (160, -1))),
```
scribblecrumb: Rather than shortening the trait, could we just make col1 larger? changing the x dimension to 180 seems to do the trick. 
image

Though it might be good to do both, since I believe "attention-seeker" takes more space than "compassionate"? or it might be about the same, it's good to have that redundancy regardless.  I'd just prefer for as much of the word to be shown as possible.

scripts/screens/cat_screens.py
keyraven: This is a good improvement - I will add it. Thanks!

scripts/events.py
keyraven: Before, the cats were resorted any time there was a rank change and "rank" type sorting was selected. The tag that kept track of  rank changes was not always catching everything. In addition, if a cat died, they would not be placed in starclan properly, since Starclan used cat.dead_for rather than cat.moons for age sorting. 
Flagging rank changes wasn't very effective in reducing lag. As clans grow larger, it's very common to have rank changes every moon, which resulted in re-sorting every moon. If we were to also resort every time a cat died, we would almost certainly be resorting every moon.  The only clans which won't trigger resort every moon would be small. 
 
 The only type of sort which is not going to commonly change every moon is id sorting. I have changed these lines to:
 
 if game.sort_type != "id":
      Cat.sort_cats()

scripts/screens/relation_screens.py
keyraven: This is a good point! I have updated the box size and increased the shortening cutoff.

scripts/events.py
Lixxis: Good point.

Lixxis (approved):

PR #1160, opened 2023-03-31
Fix typos in thoughts

SunlitFable:


resources/dicts/thoughts/alive/alive_outside/loner.json
```diff
@@ -34,7 +34,7 @@
             "Spies a kittypet outside of the window, jealous of the protection and food they have",
             "Spies a kittypet and taunts them that they can't enjoy freedom",
             "Taking a nap, hoping something exciting would escape their boredom",
-            "Was running carelessly across the thunderpath and was nearly hit by a monster"
+            "Was running carelessly across the Thunderpath and was nearly hit by a monster"
```
Lixxis, line 37: In patrols, thunderpath is written in lower case. There are only some other exceptions in the code (which might be some errors), but most of the time it is written in lower case.

Lixxis (changes requested):

resources/dicts/thoughts/alive/alive_outside/loner.json
Ryos00: They're supposed to be capitalized ^^" if there's anywhere they're lowercase, that's an error

Ryos00 (approved):

PR #1183, opened 2023-04-01
Fix optional ujson

lillithrosepup:


tests/test_json.py
```diff
@@ -9,12 +9,8 @@
...
 """
 import unittest
-try:
-    import ujson
-    from ujson import JSONDecodeError
-except ImportError:
-    import json as ujson
-    from json import JSONDecodeError
+import ujson
+from json import JSONDecodeError
```
Lixxis, line 13: Is there no difference between JSONDcodeError from ujson and json? Even so, I suggest to use 
```suggestion
from ujson import JSONDecodeError
```
to be in line with everything

scripts/game_structure/load_cat.py
```diff
@@ -3,20 +3,15 @@
...
 from scripts.cat.cats import Cat
 from scripts.cat.pelts import choose_pelt, vit, point_markings
 from scripts.utility import update_sprite, is_iterable
 from random import choice
-try:
-    from ujson import JSONDecodeError
-except ImportError:
-    from json import JSONDecodeError
+
+from json import JSONDecodeError
```
Lixxis, line 14: Is there no difference between JSONDcodeError from ujson and json? Even so, I suggest to use 
```suggestion
from ujson import JSONDecodeError
```
to be in line with everything

Lixxis (changes requested):

scripts/game_structure/load_cat.py
lillithrosepup: ujson does not have a JSONDesoceError, so the code was pulling it from json anyway

[Caused pylint to fail](https://github.com/ImLvna/clangen/actions/runs/4585415188/jobs/8097549533)

Lixxis (approved):

PR #1193, opened 2023-04-03
BUG FIX

drKhaki:
BUG FIX
- Info scroll container no longer remains on screen when switching from info window


drKhaki:
This is blaqqula on Discord<3

drKhaki:
Someone else also fixed the bug and their pull request got accepted<3

PR #1204, opened 2023-04-03
Bugfixes: Mediator, Their Thoughts, and a Ceremony. 

keyraven:
- Closing and restarting the game no longer allows a mediator to mediate more than once in a moon. 
- Fixed an incorrectly tagged ceremony. 
- Removed two dead mediator thoughts that used the o_c_n placeholder, since other_clan names currently can not appear in thoughts.


Lixxis (approved):
I quickly play tested and had no issues on the mediator screen.

PR #1206, opened 2023-04-04
BUGFIX - kittypet/loner no thoughts

drKhaki:
[this is blaqqula on Discord]
it appears as for now, kittypets and loners can only think about outsider cats so if the program gave them a cat who is not an outsider, no thoughts would be available...
(this can be changed in the future as I believe a more efficient thought system can be made)

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -1072,9 +1072,15 @@ class Cat():
                     other_cat = None
                     break
         # for cats currently outside
+        # ----------------------------
+        # FIXED KITTYPETS/LONERS NO THOUGHTS
+        # it appears as for now, kittypets and loners can only think about outsider cats so if
+        # the program gave them a cat who is not an outsider, no thoughts would be available...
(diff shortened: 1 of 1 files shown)


scripts/cat/cats.py
```diff
@@ -1072,9 +1072,15 @@ def thoughts(self):
                     other_cat = None
                     break
         # for cats currently outside
+        # ----------------------------
+        # FIXED KITTYPETS/LONERS NO THOUGHTS
```
Lixxis: As discussed/mentioned in the contributor-chat, the whole section is not really needed as a comment.

Lixxis (changes requested):

scripts/cat/cats.py
drKhaki: fixed!

drKhaki: whoops! fixed!

Lixxis (approved):

PR #1236, opened 2023-04-06
Thought Addons

Clumxy:
A bunch of new thoughts. This does not mean that the Thought Addons are Over.


resources/dicts/thoughts/alive/kitten.json
```diff
@@ -129,32 +175,18 @@
         "thoughts": [
             "Thinks they saw a StarClan cat in their dreams",
             "Is mesmerized by the night sky",
-            "Thinks they're the leader of their own Clan",
+            "Thinks theyre the leader of their own clan",
```
SunlitFable, line 178: was this intentionally changed?

resources/dicts/thoughts/alive/general.json
Ryos00, line 1: A lot of these need to be in places with another status constraint because if they're in this group, they'll appear whether or not the cats actually exist or not

resources/dicts/thoughts/alive/kitten.json
Clumxy: No. It wasn't.

resources/dicts/thoughts/alive/general.json
Clumxy: i will fix that

Clumxy:
This PR has been closed.

Reason: Bug fixing.

PR #1245, opened 2023-04-09
Add auto updating capabilities

archanyhm:


tests/test_save.py
```diff
@@ -5,6 +5,12 @@
 os.environ["SDL_VIDEODRIVER"] = "dummy"
 os.environ["SDL_AUDIODRIVER"] = "dummy"
 
+os.environ["SDL_VIDEODRIVER"] = "dummy"
```
Lixxis, line 8: Is there a reason why adding the lines 3 times?

scripts/update.py
```diff
@@ -0,0 +1,235 @@
...
+from requests import Response
+
+from scripts.progress_bar_updater import UIUpdateProgressBar
+from scripts.utility import quit
+from scripts.version import get_version_info
+
+use_proxy = False  # Set this to True if you want to use a proxy for the update check. Useful for debugging.
+
+
+class UpdateChannel(StrEnum):
```
Lixxis, line 25: After I saw you did this I also wanted to use StrEnum on my Inhertiance stuff, but it didn't work, so I tested to start the game. 
image

Do I need to use poetry? poetry.lock used python version 3.11 but in your requirements, we don't have that constraint.

Lixxis (changes requested):

tests/test_save.py
archanyhm: That is courtesy of Luna. I gotta ask why exactly that was included in the commits myself.

scripts/update.py
archanyhm: Nope, the issue here is that I've been negligent and used StrEnum without checking the supported version. It's only available on 3.11+

I've fixed this by including the StrEnum package as a drop-in replacement.

scripts/screens/organizational_screens.py
```diff
@@ -189,6 +203,19 @@ def screen_switches(self):
         self.open_data_directory_button.hide()
         self.closebtn.hide()
 
+        self.update_button = UIImageButton(scale(pygame.Rect((1154, 50), (382.5, 75))), "",
+                                             object_id="#update_button", manager=MANAGER)
+        self.update_button.visible = 0
+
+        try:
+            global has_checked_for_update
+            if game.settings['check_for_updates'] and not has_checked_for_update and has_update(UpdateChannel(get_version_info().release_channel)):
```
Lixxis: There is still an issue with the enum I assume. (updated all requirements in requiremnts.txt) I would suggest to make the StrEnum as a normal Enum and just write the strings yourself.
image

Lixxis (changes requested):

Lixxis (approved):

PR #1246, opened 2023-04-09
ADDITION - UI moons and seasons

drKhaki:
- In your settings, you can now enable a clan page UI addition which displays clan age and current clan season
- the box can be closed or open, allowing two display options (closed option comes with a tooltip)
- stats button moves to the right of the screen on clan page, as the additional box only appears there
- two clan layout positions were sliiitghly moved so the cat didn't hid behind the box (very minor change)


scribblecrumb:
I would suggest decreasing the tool-tip width!  If you need help figuring out how to do that, let me know.  It's something you have to set in the tool tip theme .json

As for the arrow button, I feel like it's rather small and difficult to see against some backgrounds (screenshot for an example of this).  I would suggest trying to adapt one of the square arrow buttons we have, perhaps you can sink half of it behind the moon/season box and just have the head of the arrow poking out?  If you need a mock-up to better understand what I mean then i can whip one up for you.

image

The full-size box image looks a bit crunched, double check that you inputted the correct dimensions for it!  It might not be bad to make it a bit less wide as well, though I think we should account for the space '1000 Moons' would take up.  If that's already been accounted for then disregard!

drKhaki:
Alright, will do!
Someone mentioned the arrow being too big so I shrinked it ^^
But yeah, I kinda dislike the image too now lol, will change!

scribblecrumb:
Yeah, I saw that Anju suggested making it smaller xD Which isn't necessarily bad, but we do want to ensure that the button is clearly visible and easy to click.  We gotta keep in mind people with motor disabilities who may struggle to get their mouse onto a teeny button

drKhaki:
Well said! Do you have the time to get me an arrow, please? ^^

scribblecrumb:
You should be able to re-use the right arrow button already in the game! I believe their object ID is # arrow-right-button

drKhaki:
How's this:
slika
slika

drKhaki:
How's this:
slika
slika

drKhaki:
also, may I ask where do I assign the tooltip object ID? I didn't seem to notice it in the uiimagebutton class ^^

drKhaki:
ALL WAS FIXED! Tooltips are now nice and smol<3
(Art contrbutor role here I come!)

scribblecrumb:
image
looks like it's still a lil crunched ^^;

PR #1255, opened 2023-04-10
FIXED - moons and seasons UI + searchboxes + updated credits

drKhaki:
- moons and seasons UI is now available on most screens requested
- wide mode disabled in the Events screen
- the arrow nuw turns around in wide mode
- adjusted crystal cave medicine cat den label position
- made cat list screens move for 50px down if the moons and seaons UI is enabled
- updated the credits text

FIXES:
- fixed a searchbox in Unknown residence being different in size that all the others


scribblecrumb:
Rather than graying out the arrow button on screens it's not allowed, how about just removing the arrow altogether? I think it would look cleaner.

image
You've also still got some crunching happening on the framing image.

I'm also not really sure why moving down the cat list and the med den label were necessary? I'm not a big fan of how the change looks, I think the cat list especially looks rather odd with the large gap.  If it was a problem with button inputs, that can be solved with proper button layering and I'm happy to help with it!

drKhaki:
all was fixed! however, I can't seem to shake this crunching...it went away when I resized them to height 140 (photoshop and code) but that was too small (it is now 145)
I tried resaving all the files, saving them with different drawing programs,... I made sure that the size ratio in UIImageButton is correct too

drKhaki:
Changed it to 150 and it isn't crumped
I also am quite bothered by the sun

drKhaki:
There!

PR #1262, opened 2023-04-10
QoL: Keybinds

lillithrosepup:


resources/gamesettings.json
```diff
@@ -36,6 +36,11 @@
             "The cursor will be replaced with a cat paw.",
             false
         ],
+        "keybinds": [
+            "Keybinds",
+            "Enables certian keybinds to be used throughout the menus for quick navigation.",
```
Thlumyn, line 41: spelling

scripts/screens/clan_creation_screens.py
```diff
@@ -306,6 +370,44 @@ def handle_choose_background_event(self, event):
         elif event.ui_element == self.elements['done_button']:
             self.save_clan()
             self.open_clan_saved_screen()
+    
+    def handle_choose_background_key(self, event):
```
Thlumyn, line 374: we really dont have a list of possible biomes? with it hardcoded in like this, I can see this getting forgotten about and causing problems when more biome are added

keyraven:
Is a toggle necessary for this feature? 
The keybind will still trigger when a player is entering text. This isn't a issue at the moment, but I am concerned it might cause issues if more keybinds are added.

lillithrosepup:
> The keybind will still trigger when a player is entering text. This isn't a issue at the moment, but I am concerned it might cause issues if more keybinds are added.

fixed!

lillithrosepup:
reopening to bump

PR #1267, opened 2023-04-11
bugfix: moonskip crash

ZtheCorgi:
fixed a crash that would happen upon skipping a moon after several cats died and the only cat remaining was the leader


ZtheCorgi:
sorry i don t know whats up with the reversions i did that by accident :sob: :sob:

PR #1269, opened 2023-04-11
Zabe's fix

lillithrosepup:
fixed a crash that would happen upon skipping a moon after several cats died and the only cat remaining was the leader


lillithrosepup:
misclick this was supposed to go to release not dev

PR #1286, opened 2023-04-12
ci: Auto upload to itch

larkgz:
Currently, @Thlumyn has to manually upload each update to itch.io, causing them to become out of sync with the updates that are pushed to the auto-updater. This PR adds an action to upload to itch.io automatically (on release tags) as well.

Important notes: 
* Everyone with write access to this repository will be able to cut a release on itch.io. Be careful!
* In order to use this action, the secret `BUTLER_API_KEY` must be set in this repository, where its value is the API key used for authentication with butler. The API key can be obtained from itch.io by installing butler and following [this guide](https://itch.io/docs/butler/login.html#running-butler-from-ci-builds-travis-ci-gitlab-ci-etc).
* After this action succeeds, you will probably have extra channels, since this action uploads to new channels. On itch.io, you will have to delete the old channels and rename the display names of the new channels (to have nicer names). You will only have to do this the first time.
* The MacOS file pushed to itch.io is the dmg instead of the zip because a `.dmg.zip` will open the `.dmg` instead of the game while using the app.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
         env:
           BUTLER_API_KEY: ${{ secrets.BUTLER_API_KEY }}
         if: startsWith(github.ref, 'refs/tags/') && env.BUTLER_API_KEY != ''
-        run: butler push Clangen_Linux64_glibc2.31+.tar.xz sablesteel/clan-gen-fan-edit:linux_64_glibc_231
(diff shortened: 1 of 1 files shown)


Thlumyn:
I'm still looking into this, 

@ other senior devs - please don't approve or merge this until I give the ok, as there is no butler api key set up in secrets right now

PR #1287, opened 2023-04-12
pronouns for general patrols

capillarys:
since key's pronouns thing is up, i've finished retagging general patrols (aka the patrols in the general folder that apply regardless of biome)!! summoning key to take another look at my tagging hhh bc i probably made a lot of mistakes. but it's been fun learning this new system!

also tried to catch general typos/proofreading issues, and consolidated some patrols' tags for ease of reading. 

also i will note that some of the "they" pronouns are a bit confusing, as to whether it's referring to the whole patrol or just the one cat in particular... but anywho i hope this all works?


keyraven:
I haven't gone through every line, but I skimmed through everything and it looks good!
Thanks again for doing this!

keyraven (approved):

PR #1301, opened 2023-04-14
Content: Mate / Inheritance Update

Lixxis:
**Changes:** 
- moved relation/ancesor tracking into an own class
- inheritance will be build and updated automatically, no save/load needed
- easier check if cats are related
- more in depth information on family tree
- moved some romantic/mating/pregnancy settings to game_config
- birthrate has more components to work against inbreeding (one pair will not have hundreds of kits)
- relationship logs can now be shown ingame
- buffed relationship interactions inpact
- multiple mates are now possible <3


scripts/events_module/relationship/romantic_events.py
```diff
@@ -169,45 +171,60 @@ def handle_breakup(self, relationship_from, relationship_to, cat_from, cat_to):
                 # game.relation_events_list.insert(0, text)
                 game.cur_events_list.append(Single_Event(text, ["relation", "misc"], [cat_from.ID, cat_to.ID]))
 
-    def big_love_check(self, cat, upper_threshold=40, lower_threshold=15):
+    def handle_confession(self, cat_from):
```
keyraven, line 174: With some quick testing, I noticed that cats who are already mates will very commonly "confess" their love to each other. I guess they were renewing their vows! /j Not a game-breaking bug, but very common.

scripts/screens/relation_screens.py
```diff
@@ -1098,6 +1087,7 @@ class ChooseMateScreen(Screens):
     mate = None
     current_page = 1
     selected_cat = None
+    selected_mate_index = 0
```
keyraven, line 1090: Another small bug - when you select a mate for a cat with only one mate, the arrow buttons will not appear. In order to assign them another mate, you must leave the screen and return.

scripts/events_module/relationship/pregnancy_events.py
```diff
@@ -54,110 +77,34 @@ def handle_having_kits(self, cat, clan):
...
+        second_parent, affair = self.get_second_parent(cat, game.settings['affair'])
+        second_parent_relation = None
+        if second_parent and second_parent.ID in cat.relationships:
+            second_parent_relation = cat.relationships[second_parent.ID]
+        elif second_parent:
+            second_parent_relation = Relationship(cat, second_parent)
+            cat.relationships[second_parent.ID] = second_parent_relation
+
+        # check if the second_parent is not none and if they also can have kits
+        if second_parent and not self.check_second_parent(cat, second_parent):
```
keyraven: Right now, no kits can occur if you have "no unknown fathers" set to True.  The else statement is reached if the second parent is none or the second parent is able to have kits, If unknown fathers is true, then the else statement will always early-return the function. 

I suggest changing this line to: 
```
if second_parent:
     if not self.check_second_parent(cat, second_parent):
          return
```

keyraven (commented):
Added a few comments for obvious bugs I ran into when quickly testing. But it seems pretty great!

scripts/events_module/relationship/romantic_events.py
Lixxis: There was general an issue with the utility function of getting the highest romantic function, test is added and working now.

scripts/events_module/relationship/pregnancy_events.py
Lixxis: Don't know why the interface don't show it, I added your change.

Ryos00 (approved):

keyraven (approved):

PR #1302, opened 2023-04-15
Thought Addons 1#

Clumxy:
Some new Thoughts. this does **not** mean that the Thought Overhaul is over.


keyraven (changes requested):
Since this is new content, can you tag this with the correct pronoun tags? [Here is the guide. ](https://docs.google.com/document/d/1qcks1E25NkOfU6zdwUlccU42VRkHv_WNaCsMpUP5jDA/edit)

Clumxy:
> 

Sorry, but I dont think i really have the time to do that. These were made before the pronouns were added. Im sorry and I hope you understand.

resources/dicts/thoughts/alive/kitten.json
```diff
@@ -129,32 +175,18 @@
         "thoughts": [
             "Thinks they saw a StarClan cat in their dreams",
             "Is mesmerized by the night sky",
-            "Thinks they're the leader of their own Clan",
+            "Thinks theyre the leader of their own clan",
```
SunlitFable: i pointed this out last time, but it still hasn't been fixed?

Clumxy: I'll try to fix that.

PR #1305, opened 2023-04-15
Fixed poisoned hunt events

bambiroe:
Fixed the same cat getting poisoned multiple times in one hunt.


larkgz:
(Same as my other comment)

Rather than modifying a list, it may make sense to just replace the choices call with sample, which selects randomly without replacement.

PR #1306, opened 2023-04-15
FIX: Poisoned hunt events [release ver.]

bambiroe:
Fixed the same cat getting poisoned multiple times in one hunt.


larkgz (commented):
Rather than modifying a list, it may make sense to just replace the `choices` call with `sample`, which selects randomly without replacement.

keyraven (approved):

Ryos00 (approved):

PR #1309, opened 2023-04-15
Other: Safe_Save, to prevent file nullification. 

keyraven:
All safe files now go through safe_save, which checks for file nullification. All files new save files are written to a temp file first.  If the temp file is not nullified, the temp file is moved and renamed. Otherwise, saving is attempted again. If the file is nullified 15 times, it throws a exception, which is caught and displayed to the user. 

We are not certain was is causing the file nullification. It seems to happen to a random selection of files. but most commonly clan_cats.json and settings.txt. Hopefully, it can be fixed by trying to save the file again. It that doesn't work, this still ensures people's data is not deleted.


scripts/game_structure/game_essentials.py
```diff
@@ -191,6 +193,50 @@ def update_game(self):
...
+            
+            # If_data is not empty, and _binary_data is
+            # all null_bytes or empty. 
+            if _data and not any(_binary_data):
+                i += 1
+                if i > max_attempts:
+                    print(f"{file_name} has been nullied too many times. Saving Failed.")
+                    raise RuntimeError(f"{file_name} was nullied too many times!")
+                print(f"{file_name} was nullifed. Trying again.")
+                continue
```
archanyhm: To verify the integrity of the file, can't we just read it back entirely and check if `write_data` equals whatever was written in the file? That might be a bit more inefficient, but _guarantees_ that the file is saved correctly.

keyraven: That's a good point. I originally only read part of the file for performance reasons, but some quick tests with a large save show the decrease in performance is minimal. I have changed the function to read in the entire file.

scripts/clan.py
```diff
@@ -766,22 +766,11 @@ def save_clan(self):
         if game.clan.game_mode in ['expanded', 'cruel season']:
             self.save_freshkill_pile(game.clan)
 
-        with open(get_save_dir() + f'/{self.name}clan.json', 'w',
-                  encoding='utf-8') as write_file:
-            json_string = ujson.dumps(clan_data, indent=4)
-            write_file.write(json_string)
+        game.safe_save(f"{get_save_dir()}/{self.name}clan.json", clan_data)
 
-        if os.path.exists(get_save_dir() + f'/{self.name}clan.txt'):
```
Lixxis, line 774: Not quite sure if we should remove this line.
When an old save gets transformed, the old txt file will then stay forever.

scripts/game_structure/game_essentials.py
```diff
@@ -191,6 +193,46 @@ def update_game(self):
...
+            Raises a RunTimeError is 
+            the file is nullied too many times. """
+        
+        # If write_data is not a string, 
+        if type(write_data) is not str:
+            _data = ujson.dumps(write_data, indent=4)
+        else:
+            _data = write_data
+            
+        dir_name, file_name = os.path.split(path)
```
Lixxis, line 211: Maybe adding a safety check that the file_name is not empty?

scripts/game_structure/game_essentials.py
```diff
@@ -191,6 +193,46 @@ def update_game(self):
...
+        
+        temp_file_path = get_temp_dir() + "/" + file_name + ".tmp"
+        i = 0
+        while True:
```
Lixxis, line 217: I can't see the purpose of this while loop. Can you explain it to me?

scripts/clan.py
keyraven: Fair point! I have added these lines back.

scripts/game_structure/game_essentials.py
keyraven: Added!

scripts/game_structure/game_essentials.py
keyraven: The while loop is to save the file again if something goes wrong. Since file nullification seems to occur to random files, there is a chance that trying again will resolve the issue. 
If you'd prefer, I can refactor the while loop into a for loop.

Ryos00 (approved):

Lixxis (approved):

PR #1311, opened 2023-04-15
[releasebranch] Debian package for debian/ubuntu/chromeos + butler

lillithrosepup:
Installs clangen to /opt/clangen, symlinks it to /usr/local/clangen for console access, and creates a .desktop file

this also cherrypicks lago's butler additions into releasebranch


lillithrosepup:
did it wrong ugh

PR #1356, opened 2023-04-18
Update leader.json

dusk0220:
testing war events


resources/dicts/events/misc_events/forest/leader.json
```diff
@@ -1,3 +1,12 @@
-[
+[{
+    "event": "War"
```
scribblecrumb: this is un-needed! to mark this as a war event, you'll want to add "war" to the tags.

scribblecrumb (changes requested):

PR #1384, opened 2023-04-21
Move version.ini and opendatadir.bat to be generated dynamically

lillithrosepup:
all logic is now in build.py

(@archanyhm review pls)


archanyhm:
We don't need an OpenDataDir.sh for macOS/Linux since those create a symlink on startup.

PR #1390, opened 2023-04-22
Moving around cat loading functions

Thlumyn:
Split json_load into json_load and json_load_cat so that the cat loading from json logic can be reused for individual cat imports (and in other potential cases where loading one cat might be needed)\

Also ugghhh I really wish I hadnt made the main clangen thing on my person github it looks so messy


Thlumyn:
Edit: okay my attempt to use browser based github were not successful lmao, give me a moment

Thlumyn:
Heck, Im going to close this until I figure out why its failing the pylint

PR #1415, opened 2023-04-27
blue-golden shell peltcolor

ghost:
adds a new pelt color to the game! based on the blue-golden shell coloring seen mostly in british shorthair cats


scribblecrumb:
Hey Sicko!  Very pretty color, but after some discussion it's been decided that it's just a bit too similar to the light brown we already have (though it would make for a very cool mod color).  We're very very picky about new colors being added, due to the amount of work they present for future additions, and due to a desire to make sure all colors have a unique footprint in the game.  Trust me, the most recent brown additions went through a million iterations and butt-loads of discussion before they were added.  Thank you for the PR, though!  You're welcome to try creating an addition for a different section of the game that we're less strict on (like patches).

PR #1428, opened 2023-05-03
Update scripts/cat/cats.py

dnbow:
Changed the generation in scripts/cat/cats.py to be in function Cat.expand instead of Cat.__init__, so it may be regenerated


dnbow:
did the wrong thing

PR #1430, opened 2023-05-03
Move f2 debug to a console system

lillithrosepup:
**Note:** This has an eval command. It gives sufficient warning of what it can do, and discourages users from pasting stuff. I included the command for tech support, to help diagnose and possibly fix problems. I am 100% willing to remove this at any time.

image

image
image

This menu is planned for use by @dnbow for a more extensive testing experience, with features such as spawning cats and more

Requesting review from @archanyhm

### DIFF
--- a/main.py
+++ b/main.py
@@ -26,6 +26,8 @@ from scripts.housekeeping.stream_duplexer import UnbufferedStreamDuplexer
 from scripts.housekeeping.datadir import get_log_dir, setup_data_dir
 from scripts.housekeeping.version import get_version_info, VERSION_NAME
 
+from scripts.debugmode import debugmode
+
 directory = os.path.dirname(__file__)
 if directory:
     os.chdir(directory)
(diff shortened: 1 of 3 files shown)


archanyhm:
This looks pretty fine to me!

While I agree that an `eval` command might pose a risk to end users, I think there's enough measures in place to prevent malicious use. If a user decides to run such a command that has been sent to them by a third party despite the warning that appears, the "point of failure" from a security perspective is the user. It's fair to assume at this point, the user would be willing to run standalone malicious software provided by the "attacker".

One question her ethough:
Is it possible to copy/paste code in the console? If so, can the `understandrisks` command be run through a "script"? That's my only concern as of now.

(I'll review the code changes themselves once I'm back home. Thanks for the great work already, though!)

lillithrosepup:
Yeah, internally uiconsolewindow uses uitextentryline which has copy and pasting, but couldnt a script type each key manually? And even then, why would a script have to go thru my console, it could just open cmd.exe and do arguably much more harm easier

PR #1445, opened 2023-05-08
Performance-friendly changing of a Sprites opacity

dnbow:
I dont know if the fading fog is still wanted as it was a place holder so I removed it

### DIFF
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -1427,7 +1427,7 @@ def update_sprite(cat):
             
             if cat.df:
                 faded_sprite.blit(sprites.sprites[f"lineartdf{cat_sprite}"], (0, 0))
-            else:
+            elif cat.dead:
                 faded_sprite.blit(sprites.sprites[f"lineartdead{cat_sprite}"], (0, 0))
             new_sprite = faded_sprite
 
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -1411,7 +1411,6 @@ def update_sprite(cat):
 
             new_sprite.blit(sprites.sprites['fademask' + stage + cat_sprite],
                             (0, 0), special_flags=pygame.BLEND_RGBA_MULT)
-
             if cat.df:
                 temp = sprites.sprites['fadedf' + stage + cat_sprite].copy()
                 temp.blit(new_sprite, (0, 0))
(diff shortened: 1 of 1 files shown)


lillithrosepup:
swag

Ryos00 (approved):

keyraven (changes requested):
This was my original method, however, it does not work properly for pixels that are already partially transparent. set_alpha() sets every pixel to the same value, which is not the intended effect.  The Starclan lineart has a lot of partial transparency.  

I also attempted to leave the lineart unchanged and only apply transparency to the color of the cat, but that effect was deemed too subtle, since it often just looks like the cat was turning black.

dnbow:
i think this is how i got around that

scripts/utility.py
```diff
@@ -1420,6 +1419,17 @@ def update_sprite(cat):
                 temp = sprites.sprites['fadestarclan' + stage + cat_sprite].copy()
                 temp.blit(new_sprite, (0, 0))
                 new_sprite = temp
+                
+        if cat.opacity < 100 and not cat.prevent_fading and game.settings["fading"]
```
keyraven, line 1423: Missing a `:`

scripts/utility.py
```diff
@@ -1420,6 +1419,17 @@ def update_sprite(cat):
...
+            new_sprite.set_alpha(int(cat.opacity * 2.55))
+            faded_sprite = Surface((spriteSize, spriteSize), pygame.HWSURFACE | pygame.SRCALPHA).convert_alpha()
```
keyraven: Surface is not defined. use `pygame.Surface`

keyraven (changes requested):
Concerns:

- You took the method of not applying the opacity to the lineart.  As I said in my previous review, this method was not used because it was too subtle.  The fading effect must be clear from the list screen, not just the profile screen.  It works well enough for ginger cats, but it not clear enough at the list-page level for black and gray cats.  This is not a method I want to use. 
- You are applying the lineart twice. This doubles the partial transparency for all cats with opacity applied. 
- You are applying the lineart after the masks for missing limbs are applied. This means that when cat starts fading, they gain transparent limbs.

dnbow:
ahhhh i see, ill close it for now but if i find a solution that solves that ill reopen it

PR #1452, opened 2023-05-11
Update cat_screens.py

Snowstar38:
- Modifies the get_death_text function to be less nested
- Fixes a crash when a leader is murdered twice in the same moon and the history is opened
- Changes text for a leader's final death depending on how many lives they had left

I hope I am doing this right; this is my first time coding/using GitHub

### DIFF
--- a/resources/dicts/patrols/beach/border/leaf-bare.json
+++ b/resources/dicts/patrols/beach/border/leaf-bare.json
@@ -38,36 +38,5 @@
             "reg_death": "r_c was swept out to sea by a flood.",
             "lead_death": "{VERB/p_l/were/was} swept out to sea by a flood"
         }
-    },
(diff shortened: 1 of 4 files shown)


keyraven:
Currently, all cats say "lost their final lives.." before the death text.  I'm assuming this should only occur for leaders.

keyraven:
Did you get my comment? I'll give a couple more days for a response, but then I'll have to close this PR for inactivity.

Snowstar38:
Yeah, I'm sorry! I think I have a fix; just have to figure out how to modify this PR to add it since I've never used GitHub before. Thanks for your patience - I completely missed your response.

keyraven:
Let me know when you are ready for another review.

PR #1453, opened 2023-05-12
itch: Fix itch toml for linux

larkgz:
Can specify platform so it doesn't refer to files that don't exist. I don't know if anyone's using the itch.io app on Linux, but it doesn't hurt to fix this issue.

### DIFF
--- a/.itch.toml
+++ b/.itch.toml
@@ -1,4 +1,11 @@
 [[actions]]
 name = "play"
+platform = "windows"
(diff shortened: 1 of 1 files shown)


keyraven:
What would occur on mac? That platform is not explicitly mentioned.

larkgz:
As far as I know at the moment, nothing. The toml doesn’t seem to be in the right folder for MacOS, so it doesn't seem to affect anything.

PR #1458, opened 2023-05-14
five thoughts + five welcome events + pronoun tagging

lulIaby:


keyraven:
You have a fancy non-ascii ` '  ` in `resources/dicts/thoughts/dead/darkforest/medicine_cat.json`, which is causing the encoding test to fail.

keyraven (approved):
The failing test is due to an old issue in dev. Should be fine upon merging.

PR #1469, opened 2023-05-18
Added a few thoughts

sarhecker:


keyraven (changes requested):
It looks like main.py has been emptied, and a new empty file `=` has been created.  The game will not run without main.py.

sarhecker:
> It looks like main.py has been emptied, and a new empty file `=` has been created. The game will not run without main.py.

all fixed i think. sorry about that!

keyraven (approved):

PR #1475, opened 2023-05-21
Content: New border Patrols Including other clan

redhairedl:


Lixxis (changes requested):
The format for the "success_text" and "fail_text" has changed into a dictionary and not a list.
Please reformat your patrols like the current patrols format. There may be some other changes beyond the ones I mentioned. (Didn't take a deeper look into the new format)

+ please also provide a description for the PR

Ryos00 (approved):

PR #1488, opened 2023-05-26
Other: Pelt Refactor

keyraven:
- Removed the many pelt-pattern specific classes, and replaced them with a single Pelt class. 
- Moved a bunch of appearance-specific properties into the Pelt class. 
- Moved all the free-floating appearance lists into the pelt class as static attributes. 
- Moved all the appearance_utility functions into the pelt class. Deleted appearance_utility.py. 
- pelt.white is no longer a thing. 
- Removed saving of "age", "pelt_white", and sprites for young and senior adult. They were unneeded.
- Condition files are no longer saved to file in classic mode. This should allow closing and re-loading to clear conditions if a bug gives a classic cat a condition. 
- Moved generation of the save dictionary for a single cat into the cat class, so it's centralized in one place. 
- Cats can now generate with natural accessories (this was a bug)


Ryos00 (approved):

scripts/cat/pelts.py
```diff
@@ -1,626 +1,1072 @@
...
-        self.length = length
+    @staticmethod
+    def generate_new_pelt(gender:str="female", parents:tuple=(), age:str="adult"):
+        new_pelt = Pelt()
+        
+        pelt_white = new_pelt.init_pattern_color(parents, gender)
+        new_pelt.init_white_patches(pelt_white, parents)
+        new_pelt.init_sprite()
+        new_pelt.init_scars(age)
+        #self.init_accessories(age)
```
Lixxis: Why adding this line if it is commented out?
Maybe set boolean if an accessory should be init or not?

scripts/cat/pelts.py
```diff
@@ -1,626 +1,1072 @@
...
+        
+        self.reverse = reverse
+        self.skin = skin
 
-    def __init__(self, colour, white, length):
-        self.white = white  # boolean; does cat have white on it or no
-        self.colour = colour
-        self.length = length
+    @staticmethod
+    def generate_new_pelt(gender:str="female", parents:tuple=(), age:str="adult"):
```
Lixxis: Maybe no default for gender? Might lead to hidden bugs.

scripts/utility.py
```diff
@@ -1280,11 +1272,11 @@ def update_sprite(cat):
         # Don't update the sprite if the cat is faded.
         return
 
-    # First make pelt, if it wasn't possible before
+    """# First make pelt, if it wasn't possible before
```
Lixxis: Still needed?

Lixxis (changes requested):

scripts/cat/pelts.py
keyraven: It is commented out because the original function actually didn't work, and I was worried that making it work would break something down the line.  It set variable called "acc_display" , which is never referenced anywhere else in the code - not even in the __init__ for the cat. I'm guessing it allowed that new parameter to be set because this function was being called in __init__.

scripts/cat/pelts.py
keyraven: That's a fair point.  I'll remove the default.

scripts/utility.py
keyraven: I don't believe so, since cats are simply assigned a default "Pelt" rather than None in the define-variable part of __init__. I can remove the commented out code entirely to make things cleaner

scripts/cat/pelts.py
Lixxis: If it is not used, we should clean it up and remove the function + the comment line with the function.
In my opinion, there is no need to store old unused code, these are just unnecessary lines of code.

scripts/utility.py
Lixxis: I would do so. The cleaner the code the better.

scripts/cat/pelts.py
keyraven: I did some testing, and I think it's safe to comment out the line and allow accessory generation.  Since no accessories generating was a bug, it's probably best to fix it. I did lower the chances of cats generating with accessories, since it was set pretty high.

Lixxis (approved):

PR #1493, opened 2023-05-27
Force file sync

larkgz:
Attempt at stopping save file nullification, which seems to be caused by computers forcibly shutting down (e.g. by holding down the power button, power outage) before all file operations are complete.

Modifies `save_save()` since it's being called a bunch anyway.


keyraven (approved):
From a quick check, this does seems to fix the file nullification issue. I don't want to force shut-down by computer more than once, so I can't be 100% certain. But from a single shut-down, it works!

archanyhm:
Sounds good to me. We'll let the users do the testing from here on.

PR #1494, opened 2023-05-27
added option for exile reason

sarhecker:


keyraven:
Where is the exile reason stored? Is it displayed anywhere?

sarhecker:
> Where is the exile reason stored? Is it displayed anywhere?

exile reason is stored in  "exile_events" in the .json file of a cat and its displayed in the history tab

scripts/cat/history.py
```diff
@@ -139,6 +143,8 @@ def make_dict(cat):
             "possible_scar": cat.history.possible_scar,
             "scar_events": cat.history.scar_events,
             "murder": cat.history.murder,
+            "possible_exile": cat.history.possible_exile,
```
keyraven, line 146: What is the use of "possible_exile"? Are there events that can cause time-delayed exile?

scripts/game_structure/windows.py
```diff
@@ -621,6 +621,63 @@ def process_event(self, event):
...
+
+    def process_event(self, event):
+        super().process_event(event)
+
+        if event.type == pygame_gui.UI_BUTTON_START_PRESS:
+            if event.ui_element == self.done_button:
+                exile_reason = sub(r"[^A-Za-z0-9<->/.()*'&#!?,| ]+", "", self.exile_entry_box.get_text())
+                #self.the_cat.exile_reason = exile_reason
+                self.the_cat.exile()
+                self.history.add_death_or_scars(self.the_cat, text=exile_reason,exile=True)
```
keyraven, line 670: It it best to create a new function to add exile reason, since "add_death_and_scars" is already used for a lot of purposes. That particular function also needs to be re-done, since it causing a few bugs right now.

keyraven (changes requested):

PR #1496, opened 2023-05-28
fix name crash 2

ozziekai:


scripts/game_structure/windows.py
```diff
@@ -414,9 +414,9 @@ def process_event(self, event):
                                                     use_prefix,
                                                     None,
                                                     self.the_cat.pelt.colour,
-                                                    self.the_cat.eye_colour,
```
keyraven, line 417: There is also some bits above this section that should be changes to fix the error (adding pelt. in front of it.)

keyraven (changes requested):

keyraven:
Oh, nevermind, I see you made a separate PR for the other part

keyraven (approved):

PR #1505, opened 2023-05-31
Content: adoptive parent screen

Lixxis:
There is now the ability to choose / change the adoptive parents of a cat. 

**Additionally:** 
* adoptive parents are only automatically set at birth
* settings for same sex kits were changed (now they really adopt kits instead of not having them)
* adapted balance function for pregnancy chance


scripts/screens/all_screens.py
```diff
@@ -85,6 +86,7 @@
 
 choose_mate_screen = ChooseMateScreen('choose mate screen')
 choose_mentor_screen = ChooseMentorScreen('choose mentor screen')
+choose_adoptive_parent_screen = ChooseAdoptiveParentScreen('choose adaptive parent screen')
```
keyraven: there is a incorrectly using word (maybe a typo) in the name "adaptive" vs "adoptive"

keyraven (changes requested):
I think we already discussed this, but are you planning on adding the a "relation-warning" if the added adoptive parent caused in-breeding or related mates somewhere else in the clan?

Ryos00 (approved):

scripts/screens/all_screens.py
Lixxis: thanks for mentioning, fixed it

Lixxis: Thanks for mentioning, I fixed the typo

About the "relation-warning", yes just added to my todo list. I will start to work on it next, but I want to have the changes already on the dev because there was a major change for automatically adoptive parents and changing the adoptive parents is currenlty only possible via file editing.

keyraven (approved):

keyraven:
looks great! awesome stuff :)

PR #1533, opened 2023-06-07
Fix - Adoptive parents screen: next/prev cats not including kittens and adolescents

Maxxicloudd:
Key if this breaks something i'm sorry /lh


sedgestripe:
@Maxxicloudd looks like you reverted the change accidentally, there are no files changed with this pr

Maxxicloudd:
> @Maxxicloudd looks like you reverted the change accidentally, there are no files changed with this pr

You are right, i am the silly

PR #1537, opened 2023-06-07
Bugfixes: Faded Cats, Event Load, Adoptive Parent Screen

keyraven:
- Fixed some issues that faded cats caused for the adoptive parent screen
- Next and previous cat buttons now work properly on the adoptive parent screen
- Adjusted inheritance to properly fetch and log faded kits, grandkits, siblings, parents siblings and cousins. 
- faded_offspring now stored faded offspring through both blood and adoptive parents.
- Saved events now load properly.


keyraven:
@Lixxis  I made some changes to inheritance update function - can you take a look to make sure everything looks good?

PR #1558, opened 2023-06-10
[UPDATE] Credits

drKhaki:
Do merge into dev when you have the time<3


Lixxis (commented):
I don't know what the others think, but the collection of the seniors at the top had a little bit of structure, to be honest. Otherwise, the alphabetical order seems like a good idea!

drKhaki:
Ahh, agreed, senior contribs being on top feels more organised! Done<3

PR #1618, opened 2023-06-26
Sort dead cats by absolute age when filtered by age

Snowstar38:
- The game will now sort dead cats by their moons+dead_moons

- Kept the original functionality of get_adjusted_age for the rank (default) view so players can still see dead cats in the order they died


Snowstar38:
- Now when you go to a cat's individual relationships, the relationships with the highest values (positive or negative) overall will show first

This helps when playing with a large Clan by putting the most significant individuals in a cat's life in the forefront.

PR #1649, opened 2023-07-15
careful leader ceremony (starclan)

capillarys:
doing a little dancey dance now that this is done!!!


sedgestripe (changes requested):
@capillarys this is great! just noticed a small typo "With an approving mod" should be nod

capillarys:
HELP not a mod... discord in my brain fr... lemme get that fixed real quick!

sedgestripe (approved):

PR #1660, opened 2023-07-19
Add "remove mentor" button

Snowstar38:
This way, you can un-assign a mentor without assigning a new one.

It's useful for if you micromanage apprentices/mentors and have a lot of them automatically assigned at once when they hit 6 moons (so you don't have to remember if the mentor's "1 current apprentice" is one of the new ones you'll be reassigning, or an older app you want them to keep).


keyraven:
I worry this might give the wrong impression to players. Removing the mentor will only last until moon-skip, when they will be assigned a new random mentor. I agree it could be useful for applications you describe, but without proper content, I think most people will assume this will remove a mentor for good, and then may be frustrated when they are reassigned a mentor the following moon.

Snowstar38:
Understood - I just added a bit more info to the screen explaining it. Do you think this would work, or is the option itself too confusing, do you think?

image

Ryos00 (approved):

PR #1670, opened 2023-07-23
Add dynamic name lengths

Snowstar38:
- Cats' names in the family tree and cat list screens will now be limited by the actual width in pixels, rather than a set # of characters

Previously, if there was a cat named "llllllllll" and a cat named "mmmmmmmmmm", they would both be cut off at the same # of characters, despite one being much shorter visually.


keyraven (changes requested):
I like this change! My only note is that there is a lot of duplicated code for handling the length of the string. It would be more tidy to create a new utility function to handle the length adjustments. This will also allow easy tweaking of this behavior.

scripts/screens/cat_screens.py
```diff
@@ -2357,10 +2356,11 @@ def update_selected_cat(self):
         )
 
         name = str(self.the_cat.name)
-        if len(name) > 17:
-            name = name[:14] + "..."
-        self.selected_cat_elements["cat_name"] = pygame_gui.elements.UILabel(scale(pygame.Rect((790, 140), (320, -1))),
-                                                                             name,
+        length_limit = 300
+        font_size = 26
+        short_name = shorten_text_to_fit(name, length_limit, font_size)
```
keyraven: I suggest removing the variables "length_limit" and "font_size", and simple pass these values directly into shorten_text_to_fix.  These variables are not used outside of the function call, and removing unnecessary variables increase readability.

keyraven (changes requested):
Looks pretty good! I just have one small note, but other than that, it's a very nice addition!

keyraven (approved):

PR #1671, opened 2023-07-23
Complex relationships for decrease events

Snowstar38:
- Relationship events that decrease values can now also affect other values
  - Example: a "Decrease platonic like" event will now decrease "comfortable" and increase "dislike".

- Changed passive influence default value in game_config to compensate for these additional relationship effects
  - Effectively lowered the default passive_buff, since there are now more complex value additions and therefore relationship values are being increased/decreased more frequently (from ~77% of "main" relationship type value added to other values -> ~66%)

### DIFF
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -69,7 +69,7 @@
 		"max_interaction": 5,
 		"max_interaction_special": 8,
 		"compatibility_effect": 5,
-		"passive_influence_div": 1.3,
+		"passive_influence_div": 1.4,
 		"chance_for_neutral": 10,
 		"chance_of_special_group": 8,
 		"chance_romantic_not_mate": 15,
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -69,7 +69,7 @@
 		"max_interaction": 5,
 		"max_interaction_special": 8,
 		"compatibility_effect": 5,
-		"passive_influence_div": 1.4,
+		"passive_influence_div": 1.5,
 		"chance_for_neutral": 10,
 		"chance_of_special_group": 8,
 		"chance_romantic_not_mate": 15,
--- a/scripts/cat_relations/relationship.py
+++ b/scripts/cat_relations/relationship.py
@@ -433,16 +433,22 @@ class Relationship():
     #                            complex value addition                            #
     # ---------------------------------------------------------------------------- #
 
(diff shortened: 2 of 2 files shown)


keyraven:
What is the justification for making this change?  Was there some imbalance in the relationship values?

Snowstar38:
I had some relationship events that fired with "decrease" effects on a timeskip, and when I looked at the cats' relationship tabs, I realized that it wasn't actually affecting the relationship values at all if the two cats didn't have any relationship values to subtract from before that point. There would be event text that seems like it'd be a big deal (i.e., having a fight over something or sharing secrets) but then there was no change in the relationship values. You have to click on "view empty" to even see them, so I thought it was a bug and spent some time looking into it because it seemed so unintuitive that a dramatic relationship event could happen but it affects literally nothing.

The text doesn't specify what kind of event it was, so the player won't understand why some events that sound dramatic and negative (and say "negative effect") WILL have an effect (like "increase dislike") while other events that sound dramatic and negative WON'T affect anything (like "decrease trust").

Ex: a "high intensity" "decrease admiration" or "decrease dislike" event that fires between two cats that have no relationship values will result in the relationship still being "empty", though the event will appear in the log when you pull it up between the two cats.

Since additive interactions can have "splash effects", I figured subtractive ones should as well - otherwise you can have two cats with plenty of "decrease" events in their history that still have a 0 in every relationship value because nothing was ever actually changed.

keyraven:
Okay, that makes sense! I'm going to wait for review from Lixxis, since she wrote this system. I am unfamiliar with the system as a whole, and unsure if the lack of "splash effects" was done for a balance reason.

resources/game_config.json
```diff
@@ -69,7 +69,7 @@
 		"max_interaction": 5,
 		"max_interaction_special": 8,
 		"compatibility_effect": 5,
-		"passive_influence_div": 1.3,
+		"passive_influence_div": 1.5,
```
Lixxis, line 72: Please take a look, that this is not too much of a buff for passive influence

Lixxis (approved):

resources/game_config.json
Snowstar38: Oh, the way it does it, it divides by that number, so a larger passive_influence_div actually results in a smaller amount of passive influence - I was adjusting it down because there will be a lot more passive influence events now. Or have I misunderstood what you mean? 😅😅

PR #1675, opened 2023-07-24
Fix Crash when killing, other tweaks

keyraven:
- Remove adjusting was/were in leader ceremony custom deaths. 
- Improved some code readability.


Snowstar38:
If it helps, the reason why I made "prompt"/"prompt_processed" and "initial"/"initial_processed" distinct is because we want the processed text to display with the correct pronouns (not tags) in the text box, while still writing the actual raw text with pronoun tags to the file - you may well have retained this behavior with the cleanup, but I thought I'd explain :)

PR #1677, opened 2023-07-25
Small fixes

Snowstar38:
- Cats will be properly sorted when you open the catlist screen for the first time (fixing a rare bug with medicine cats not being in order)
- The "next" and "previous" cat buttons now work properly for the StarClan/Dark Forest guide.
- Minor spelling and pronoun tagging fixes
- Fixed tagging issue with omen list for a few patrols

### DIFF
--- a/resources/dicts/patrols/beach/med/any.json
+++ b/resources/dicts/patrols/beach/med/any.json
@@ -156,7 +156,7 @@
         "chance_of_success": 50,
         "exp": 20,
         "success_text": {
-            "unscathed_common": "p_l is plucking some rosemary when {PRONOUN/p_l/poss} attention is suddenly grabbed by omen_list_emotional_smell_sound. The fur on {PRONOUN/p_l/poss} back begins to prickle slightly, the sound of {PRONOUN/p_l/poss} heartbeat like thunder in {PRONOUN/p_l/poss} ears. {PRONOUN/p_l/subject/CAP} {VERB/p_l/turn/turns} and hare back to camp, herbs forgotten. {PRONOUN/p_l/subject/CAP} {VERB/p_l/need/needs} to tell {PRONOUN/p_l/poss} mentor about this as soon as possible!"
+            "unscathed_common": "p_l is plucking some rosemary when {PRONOUN/p_l/poss} attention is suddenly grabbed by omen_list_sight. The fur on {PRONOUN/p_l/poss} back begins to prickle slightly, the sound of {PRONOUN/p_l/poss} heartbeat like thunder in {PRONOUN/p_l/poss} ears. {PRONOUN/p_l/subject/CAP} {VERB/p_l/turn/turns} and hare back to camp, herbs forgotten. {PRONOUN/p_l/subject/CAP} {VERB/p_l/need/needs} to tell {PRONOUN/p_l/poss} mentor about this as soon as possible!"
         },
         "fail_text": {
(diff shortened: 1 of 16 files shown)


scripts/screens/base_screens.py
```diff
@@ -169,6 +169,7 @@ def menu_button_pressed(self, event):
         elif event.ui_element == self.menu_buttons["camp_screen"]:
             self.change_screen('camp screen')
         elif event.ui_element == self.menu_buttons["catlist_screen"]:
+            Cat.sort_cats()
```
keyraven: We don't want to be resorting the entire cat list whenever we open the catlist screen. This has the potential to cause lag on very large clans, and is a good amount of unneeded duplicate work. 

The bug with medicine cat may by caused be small differences in the behavior of `list.sort` and `bisect.insort`.  For now, I'm not sure how to fix this issue, and it might be more of a pain to fix then it's worth.

scripts/screens/cat_screens.py
```diff
@@ -670,26 +670,28 @@ def determine_previous_and_next_cat(self):
...
         previous_cat = 0
         next_cat = 0
+        current_cat_found = 0
         if self.the_cat.dead and not is_instructor and self.the_cat.df == game.clan.instructor.df and \
                 not (self.the_cat.outside or self.the_cat.exiled):
             previous_cat = game.clan.instructor.ID
 
         if is_instructor:
-            next_cat = 1
+            current_cat_found = 1
```
keyraven, line 679: Out of curiosity, what was causing the next-cat issue? I thought it might have been the special starclan cat list, which is what is used for the page, while the next-cat function uses the general all_cat list.

keyraven (changes requested):

resources/dicts/patrols/beach/med/any.json
```diff
@@ -156,7 +156,7 @@
         "chance_of_success": 50,
         "exp": 20,
         "success_text": {
-            "unscathed_common": "p_l is plucking some rosemary when {PRONOUN/p_l/poss} attention is suddenly grabbed by a a_sign. The fur on {PRONOUN/p_l/poss} back begins to prickle slightly, the sound of {PRONOUN/p_l/poss} heartbeat like thunder in {PRONOUN/p_l/poss} ears. {PRONOUN/p_l/subject/CAP} {VERB/p_l/turn/turns} and hare back to camp, herbs forgotten. {PRONOUN/p_l/subject/CAP} {VERB/p_l/need/needs} to tell {PRONOUN/p_l/poss} mentor about this as soon as possible!"
+            "unscathed_common": "p_l is plucking some rosemary when {PRONOUN/p_l/poss} attention is suddenly grabbed by omen_list_emotional_smell_sound. The fur on {PRONOUN/p_l/poss} back begins to prickle slightly, the sound of {PRONOUN/p_l/poss} heartbeat like thunder in {PRONOUN/p_l/poss} ears. {PRONOUN/p_l/subject/CAP} {VERB/p_l/turn/turns} and hare back to camp, herbs forgotten. {PRONOUN/p_l/subject/CAP} {VERB/p_l/need/needs} to tell {PRONOUN/p_l/poss} mentor about this as soon as possible!"
```
Ryos00: I might be mistaken about how the snippet lists work, but originally the signs were something like what's in the sight section of the omen list so I think they should still be something like that.

keyraven (approved):

Ryos00 (approved):

PR #1686, opened 2023-07-29
added new gen thoughts

thyfrankie:


resources/dicts/thoughts/alive/general.json
```diff
@@ -224,7 +264,10 @@
         "thoughts": [
             "Is wondering what r_c is thinking",
             "Is hoping r_c notices {PRONOUN/m_c/poss} improvement",
-            "Is in awe of r_c's leadership"
+            "Is in awe of r_c's leadership",
+            "Thinks r_c could be leading the Clan better than {PRONOUN/r_c/subj} {VERB/r_c/are/is}",
```
keyraven: Incorrect pronoun tags - the tag should be {PRONOUN/r_c/subject}. Alternatively, the thought might be snappier without the last pronouns. IE., "Thinks r_c could be leading the clan better"

keyraven (changes requested):

Ryos00 (approved):

PR #1701, opened 2023-08-02
Update api url to clangen.io instead of arc's personal domain

lillithrosepup:
Dont want anyone to go snooping :p


archanyhm:
Looking fine and dandy. Thanks for reminding me that I forgot to adjust the URL—

PR #1708, opened 2023-08-06
Refactoring: Static-ification of the Events Module

keyraven:
Switched all functions in the events module classes to static functions, and all calls to these functions to be static as well. 

The calls to the events module were already essentially static, but without the proper decorators to allow it. When there were instance variables, only one copy of this variable was ever needed, so it worked better as a class variable. I believe this is more readable, and removes the inter-linked copies of event-module class instances, which existed as various instance variables of other event-related classes.

Also moved old_age deaths to general, to allow medicine cats and mediators to die of old age. 
And added randomness to known outside cats joining the clan.


scripts/events_module/new_cat_events.py
```diff
@@ -255,17 +250,20 @@ def handle_new_cats(self, cat: Cat, other_cat, war, enemy_clan, alive_kits):
...
         outside_cats = [i for i in Cat.all_cats.values() if i.status in ["kittypet", "loner", "rogue", "former Clancat"] and not i.dead and i.outside]
         return any(outside_cats)
 
-    def select_outside_cat(self):
+    @staticmethod
+    def select_outside_cat():
         outside_cats = [i for i in Cat.all_cats.values() if i.status in ["kittypet", "loner", "rogue", "former Clancat"] and not i.dead and i.outside]
         for cat in outside_cats:  # iterating over the generated list
             if cat.status in ["kittypet", "loner", "rogue", "former Clancat"] and not cat.dead:
                 return cat
```
Lixxis: Did I miss a shuffle function?
I'm not 100% sure, but it seems, that it will always favor the first cat in the list, meaning the time of the generation of the cat is the only thing that is important for the selection. Is this a wished behavior?

tests/test_relation_events.py
Lixxis, line 1: Thanks for changing the tests as well :D

Lixxis (approved):
Seems good to me, but only looked at the code and not test and run it.

scripts/events_module/new_cat_events.py
keyraven: I'm not sure. I haven't really looked at this function. I can quickly add some shuffle functionality!

tests/test_relation_events.py
keyraven: Of course!

PR #1716, opened 2023-08-07
Subcommand system + cat command

lillithrosepup:
Added subcommands to debug menu, added a command to manage cats

### DIFF
--- a/scripts/debugCommands/cat.py
+++ b/scripts/debugCommands/cat.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from scripts.debugCommands.command import Command
 
 from scripts.debugCommands.utils import add_output_line_to_log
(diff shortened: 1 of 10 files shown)


keyraven (approved):

scripts/debugCommands/cat.py
```diff
@@ -0,0 +1,83 @@
...
+        game.clan.add_cat(cat)
+        add_output_line_to_log(f"Added {cat.name} with ID {cat.ID}")
+
+class removeCatCommand(Command):
+    name = "remove"
+    description = "Remove a cat"
+    aliases = ["r"]
+    usage = "<cat name|id>"
+
+    def callback(self, args: list[str]):
```
keyraven, line 26: This type-hint syntax only works for python 3.9 and above, and causes crashes on the 3.8 builds.  I fixed it on the currently-existing debug commands here (https://github.com/Thlumyn/clangen/pull/1720).

keyraven (changes requested):

keyraven (approved):

PR #1720, opened 2023-08-09
Bugfixes: Various

keyraven:
- Various types and pronoun tag fixes
- Fixed a misplaced `:` in an event that was causing crashes. 
- Switched type hints in the debug commands to pre-3.9 syntax, to support 3.8 builds. 
- Added "debug_ensure_patrol_id" which helps debugging certain patrols. 
- Removed some unnecessary recursion in patrol filtering.


keyraven:
@ImLvna  -  This has some minor fixes for debug-screen content - can you take a look?

lillithrosepup:
Looks good ! Sorry I didn't catch that, I haven't worked with 3.8 much

keyraven:
> Looks good ! Sorry I didn't catch that, I haven't worked with 3.8 much

No problem! I didn't catch it as first either!

PR #1725, opened 2023-08-14
Feature: Special date patrol framework + refactoring

larkgz:
* Refactors date stuff to be in a separate module to prevent circular imports
* Switches to use `datetime` module to store dates so that it's easier to do date comparisons (particularly between months)
* Adds game setting to disable special date events
* Add support for special date events that are more than one day long
* Support date exclusive patrols by adding the tag `date_exclusive` and the specified tag for that special date. For Halloween, this is `halloween`.


Ryos00 (approved):

scripts/patrol.py
```diff
@@ -638,6 +639,12 @@ def filter_patrols(self, possible_patrols, biome, patrol_size, current_season, p
             if patrol.season not in [current_season, "Any"]:
                 continue
 
+            # filtering for dates
+            if 'date_exclusive' in patrol.tags:
```
keyraven: Would it be possible to remove the "date_exclusive" tag, and only check for the special date tags? The "date_exclusive" tag appears redundant.

PR #1741, opened 2023-08-21
Update cats.py

NyxieSagepark:


keyraven:
Hello! I'm glad you are interested in contributing to ClanGen! However, this PR only adds a single "accessory2" line to the cat save-dictionary, which duplicates the current "accessory" line. What was your intention here? 

Also, for future reference, be sure to add a descriptive title, and some content in the description.

keyraven:
Confirmed as an accident.

PR #1761, opened 2023-09-15
Experimental text change for main list screen

Snowstar38:
- Added a new text box with a font size of 28
- Changed cat name font size in list screen from 30 -> 28

Some names were getting cut off too soon, IMO. The text still seems quite readable, since it's only a small change in font size, but more characters can fit now. This is something that will be subjective so I fully understand if this isn't approved!

### DIFF
--- a/resources/theme/text_boxes.json
+++ b/resources/theme/text_boxes.json
@@ -177,6 +177,19 @@
             "text_vert_alignment_padding": "10"
         }
     },
+    "#text_box_28": {
+        "prototype": "#text_box",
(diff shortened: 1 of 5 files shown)


Snowstar38:
ex1

ZtheCorgi:
personally, i don't think this is necessary, and without my glasses on the smaller font definitely puts a bit more strain on my eyes. my eyesight is not that bad either, so i can only imagine the problem would be exacerbated for people who have worse eyesight than i do

PR #1763, opened 2023-09-16
Fix for fst_hunt_climbtree1

BlakeRipples:
So if I remember correctly, it's "CLIMBING" and not "CLIMB". I've been putting this off since I first found the bug and went to check and fix it, but the patrol hasn't shown up again since I identified the issue.


keyraven:
This patrol also only has a skill-based success outcome --> a full fix would include a non-skill success outcome.

keyraven:
This PR has had activity for a while. @GambolCalico , have you seen my comment?

PR #1767, opened 2023-09-29
perf: Use subsurfaces for sprites

larkgz:
By using `subsurface` instead of performing a blit to new surfaces, it creates a reference to the original data rather than creating a duplicate of it. The only thing to note is that if the original surface is modified so is the subsurface and vice-versa, but the spritesheets aren't directly modified anywhere as far as I know.

### DIFF
--- a/scripts/cat/sprites.py
+++ b/scripts/cat/sprites.py
@@ -44,23 +44,6 @@ class Sprites():
         """
         self.spritesheets[name] = pygame.image.load(a_file).convert_alpha()
 
-    def find_sprite(self, group_name, x, y):
-        """
-        Find singular sprite from a group.
-
-        Parameters:
-        group_name -- Name of Pygame group to find sprite from.
-        x -- X-offset of the sprite to get. NOT pixel offset, but offset of other sprites.
-        y -- Y-offset of the sprite to get. NOT pixel offset, but offset of other sprites.
-        """
-        # pixels will be calculated automatically, so for x and y, just use 0, 1, 2, 3 etc.
-        new_sprite = pygame.Surface((self.size, self.size),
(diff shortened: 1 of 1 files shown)


scripts/cat/sprites.py
```diff
@@ -81,26 +81,22 @@ def make_group(self,
...
-            (pos[0] * sprites_x * self.size, pos[1] * sprites_y * self.size,
-             (pos[0] + sprites_x) * self.size,
-             (pos[1] + sprites_y) * self.size))
+        new_group = pygame.Surface.subsurface(self.spritesheets[spritesheet],
+                                              pos[0] * sprites_x * self.size,
+                                              pos[1] * sprites_y * self.size,
+                                              self.size * sprites_x,
+                                              self.size * sprites_y)
 
         self.groups[name] = new_group
```
keyraven: Are you able to remove self.groups entirely? As far as I remember, it not used outside of this function.

keyraven (commented):
One small comment, but this is a great, necessarily change.

larkgz:
Looks like `self.groups` is only used in one function that is never called (which I removed). 

It seems like the `groups` subsurfaces are still hanging out in memory somewhere (since I can still grab them through `.get_parent()` from the sprites; I assume that that's the reference keeping it alive), but since they're subsurfaces rather than regular surfaces I think the memory usage is minimal.

keyraven:
> Looks like `self.groups` is only used in one function that is never called (which I removed).
> 
> It seems like the `groups` subsurfaces are still hanging out in memory somewhere (since I can still grab them through `.get_parent()` from the sprites; I assume that that's the reference keeping it alive), but since they're subsurfaces rather than regular surfaces I think the memory usage is minimal.

Ahh, I makes sense. Looks great, I'll merge it in. Thanks!

PR #1770, opened 2023-10-03
Add skill checkbox for mentor screen

Snowstar38:
- Added a checkbox to the mentor selection screen that filters for cats that share at least 1 skill with the chosen apprentice (with a tooltip to explain what it does)

### DIFF
--- a/scripts/screens/ChooseMentorScreen.py
+++ b/scripts/screens/ChooseMentorScreen.py
@@ -5,6 +5,7 @@ from .Screens import Screens
 
 from scripts.utility import get_text_box_theme, scale
 from scripts.cat.cats import Cat
+from scripts.cat.skills import SkillPath
 from scripts.game_structure import image_cache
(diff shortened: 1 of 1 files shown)


keyraven (commented):
Although having compatible skills does allow the mentor to increase the skill-level, that compatibility isn't just limited to same-skill.  This may invite an incorrect perception on how mentors affect skills. 

Design-wise, I believe using only a ToolTip to label a checkbox is non-optimal.  Tooltips should contain additional information, but having a lone checkbox without a label is confusing at-a-glace.  

Ultimately, I think this is a weird thing to add on it's own.  Maybe there should be more filtering options, and it can be added to it's own window?

Snowstar38:
Good point! I'll retract this for now and work on some sort of more refined filtering options for the mentor screen and see what you guys think of them. Apologies for taking so long to reply!

PR #1786, opened 2023-10-19
Reworded Cruel Season "Coming Soon" Message

akrazyartist:
Reworded Cruel Season Scene to provide clarity on the timing, so that it gives a more measurable amount of time. This would avoid confusion around the phrase "coming soon", as that means different things to everyone

### DIFF
--- a/scripts/screens/MakeClanScreen.py
+++ b/scripts/screens/MakeClanScreen.py
@@ -49,7 +49,7 @@ class MakeClanScreen(Screens):
                       "you'd like a challenge with a bit of brutality, then this mode is for you.<br><br>" \
                       "You heard the warnings... a Cruel Season is coming. Will you survive?" \
                       "<br> <br>" \
-                      "-COMING SOON-"
+                      "-WORK ON CRUEL SEASON WILL BEGIN ONCE DEVELOPMENT ON EXPANDED MODE IS FINISHED-"
 
     # This section holds all the information needed
     game_mode = 'classic'  # To save the users selection before conformation.


Thlumyn:
Yeah I don't know, this sort of implies that we will eventually finish developing expanded mode which isn't the case, as this game is constantly improving, maybe wording it something more along the line of "still in development" if enough people don't like "coming soon" but this is something sr devs would have to talk about

akrazyartist:
ah, i used those words becasue it is directly what i have been told before in the server by mods and devs /info /nm

archanyhm:
@jitteryHatter Yes, that's pretty much the clarification given in the chat about the specifics of why Cruel Season currently doesn't receive direct work (because there's still blockers that are currently being worked on)

As for that little `Coming soon` label on the bottom of the description page… I'm not too sure what we can change it to without turning this into a concrete promise. We avoid giving deadlines and making specific promises to deliver some feature at some concrete point in time because we have _no_ dedicated, full time contributors. Everyone is helping out voluntarily and that makes meeting deadlines a bit difficult. Saying that this feature will receive more work/attention once something else is done is basically like some sort of roadmap and we… don't really have one publicly for the same reasons.

Things I heard being proposed as an alternative to `Coming soon` are things like `Planned` (which is not accurate because it implies that it only exists on paper; that's not really true since clearing out the blocker issues is _directly_ related to Cruel Season), `(Still) in development` and other derivatives are _equally_ as unspecific as `Coming soon`. I personally don't really see a way to be more "clear" about when it will come without giving a concrete release date, a concrete timeframe or some other kind of "specification".

akrazyartist:
I understand not wanting to give direct deadlines, and i do understand that this is a volunteer based program! I do now understand why it cant be changed /gen 
 
im just, very very confused now. I have been told and seen others told that it would not be worked on until expanded was finished - those words exactly, but it seems like that is not the case? If it cant be said ingame, why is it being said in the sever? I fullly understand why it cant be changed, but I do hope you can understand my confusion. Ive been told this multiple times and seen it said multiple times, but its not the case? /lh /no bad tone intended /nm

archanyhm:
@jitteryHatter It kind of is the case; that's basically the status quo, but that might be subject to change. We can always give some little hints like that when asked in the moment, but if we basically put it in the game (in the buttons description) or on some publicly available roadmap, it turns into some sort of "promise". We'd have to make sure to diligently update those references whenever things change — for example when we shift priority on something else, people leave the team or go on hiatus  or when we encounter more blockers — and then it wouldn't be accurate anymore. Stable releases for example only receive bugfixes as updates, so the time between releases can be anywhere in between a few weeks to months — currently 0.9.4 is about two-three months old — and if we go "Hey, we have to focus on this feature for the time being because it's more urgent than Cruel Season" the description would still say "comes after [x] is done". If that makes sense.

So when devs/contributors give a little "update" on the state of things like that in the server or on social media, it's basically a "snapshot" of the status quo. That's what it's like in that moment. If we put it on the website, in the game (in this case, the description) or on some other kind of roadmap, it's sort of like a "promise".

I hope this makes sense; feel free to let me know if it doesn't. I'll gladly try to explain it differently then. /gen

akrazyartist:
I think it makes sense, I think it would be better if people in the server didnt say "this will happen after x /gen", because i thought it was genunine, aka a promise. I fully understand why it isnt! volunteer code, priorites, game development is so hard haha, but i feel like " we hope to develop cruel season after expanded has been fleshed out more, but this is subject to change /gen" would be better than a definte "after expanded mode.", as i know me and alot of people think that is what will 100% happen, due to the definte nature of the messages and /gen afterwards

/info /explanation /suggestion /gen /nm /no bad tone intended

archanyhm:
@jitteryHatter I wholeheartedly agree with you on that note! /gen

Therefore, this is feedback that I can definitely forward to the development team via an internal announcement. Would this be a solution for you? /gen

akrazyartist:
Yes that would be perfect thank you! 

As an explanation, I thought because of "this will be developed after expanded is done /gen" that i could put that into the game, as it is what the devs have said, i didnt realise all of this stuff, but i fully understand now! thank you for talking to me /gen

archanyhm:
Just for the record: I'll leave this PR open until I did the announcement. I plan to add a little SOP for the developers in the coming days which would cover this specific aspect. This is not forgotten or stale.

PR #1789, opened 2023-10-20
fix: leader loses litter when losing a life

### LINKED ISSUE #1788: Pregnant Leaders Lose Litter Upon Losing A Life
Reported by a player. As the title says, the pregnancy info gets cleared when a leader loses a life, which is not an intended effect, and we have decided that this needs to be fixed.

Ryos00:
- added an exception for leaders with more than 1 life left to not have their pregnancy wiped on death

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -414,9 +414,7 @@ class Cat():
         """
         if self.status == 'leader' and 'pregnant' in self.injuries and game.clan.leader_lives > 0:
             self.illnesses.clear()
-            for injury in self.injuries:
-                if injury != 'pregnant':
-                    del self.injuries[injury]
+            self.injuries = { key : value for (key, value) in self.injuries.items() if key == 'pregnant'}
         else:
             self.injuries.clear()
             self.illnesses.clear()
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -412,8 +412,14 @@ class Cat():
 
         May return some additional text to add to the death event.
         """
-        self.injuries.clear()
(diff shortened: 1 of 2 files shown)


scripts/cat/cats.py
```diff
@@ -412,8 +412,14 @@ def die(self, body: bool = True):
 
         May return some additional text to add to the death event.
         """
-        self.injuries.clear()
-        self.illnesses.clear()
+        if self.status == 'leader' and 'pregnant' in self.injuries and game.clan.leader_lives > 0:
```
keyraven, line 415: Shouldn't this be `game.clan.leader_lives >= 0` ? To match the check below. As it is, this will only trigger if the game.clan.leader_lives is -1 or less.

keyraven (changes requested):

scripts/cat/cats.py
```diff
@@ -412,8 +412,14 @@ def die(self, body: bool = True):
...
+            self.illnesses.clear()
+            for injury in self.injuries:
+                if injury != 'pregnant':
+                    del self.injuries[injury]
```
keyraven: Rather than using `del`, I suggest dictionary comprehension. This works, but it's very nested. 

replace lines 417, 418, and 419 with 
`self.injuries = { key : value for (key, value) in self.injuries.items() if key == 'pregnant'}

keyraven (changes requested):

scripts/cat/cats.py
Ryos00: oh thank you I had no idea what I was doing

keyraven (approved):

PR #1793, opened 2023-10-21
Halloween Patrols

Ryos00:
;3


drKhaki:
i believe the "scars": [] rows in the injury section are redundant since .get("scars", ()) makes the attribute optional
also, some contributions are written with "scar" instead of "scar**s**", most likely there was a mixup due to the history text and the old tagging system ^^

PR #1798, opened 2023-10-21
BugFix: Fix Do Not Proceed Crash

keyraven:
Fix Crashing on Do Not Proceed


keyraven:
Oh, there was a fix already, but it wasn't linked to the issue.

PR #1823, opened 2023-10-24
Outside cats gain EXP over time

sand-boa:
Someone in the dev-version-discussion channel on discord mentioned they thought it was weird that lost cats returned to the Clan with exactly the same EXP as they left with. I've had this code in my game for a while to give all outside cats EXP, simulating that they have a life outside the Clan. It's a lot simpler than the apprentice EXP code, but it works. The values I set are of course subject to change!

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -721,6 +721,7 @@ class Events:
         cat.one_moon()
         cat.manage_outside_trait()
             
+       self.handle_outside_EX(cat)
         cat.skills.progress_skill(cat)
         Pregnancy_Events.handle_having_kits(cat, clan=game.clan)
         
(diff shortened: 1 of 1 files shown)


scripts/events.py
```diff
@@ -1519,6 +1520,30 @@ def handle_apprentice_EX(self, cat):
 
             cat.experience += max(exp * mentor_modifier, 1)
 
+
+    def handle_outside_EX(self, cat):
+        if cat.status in ["loner", "rogue", "exiled", "kittypet", "lost"] and cat.age != 'kitten':
+            if cat.age == 'adolescent':
```
keyraven, line 1526: What is the purpose of reducing EXP for kittypets, or for having an age-dependence on the amount of experience gained? If possible, we'd like reduce the number of nested if-else chains of this variety. 

In addition, the mix and max values should reference back to some value in game.config, to allow for easier tuning of values.

keyraven (changes requested):

keyraven:
In addition, Pylint is failing. It appears to be an indentation issue.

scripts/events.py
Thlumyn: I can see the benefit in reducing exp for kittypets, they probably get less warrior useful experience living that way, but yeah, I think these values should be handled in game.config- perhaps a base value and a kittypet value for these that are calculated and used instead of these if-else chains with hardcoded values for each possibility

keyraven:
@sand-boa Have you seen the comments on this PR?

keyraven:
@sand-boa If there is not activity or response on this PR by the 17th, I will close this PR for inactivity.

PR #1829, opened 2023-10-26
Fixed broken mate check

hobohime:
one typo was responsible for the age gap of 40 moons not being consistently applied. Fix found by zabe, listed here as ZtheCorgi.

### DIFF
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -81,7 +81,7 @@
 	},
 	"mates":{
 		"age_range": 40,
-		"overide_same_age_group": false,
+		"override_same_age_group": false,
 		"chance_fulfilled_condition": 5,
 		"chance_friends_to_lovers": 170,
 		"confession": {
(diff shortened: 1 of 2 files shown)


scripts/cat/cats.py
```diff
@@ -2079,7 +2079,7 @@ def is_potential_mate(self,
             # if only one is aged up at this point, later they are more moons apart than the setting defined
             # game_config boolian "override_same_age_group" disables the same-age group check.
             if game.config["mates"].get("override_same_age_group", False) or self.age != other_cat.age:
-                if abs(self.moons - other_cat.moons) > game.config["mates"]["age_range"] + 1:
+                if abs(self.moons - other_cat.moons) or (other_cat.moons - self.moons) > game.config["mates"]["age_range"] + 1:
```
keyraven: Checking the negative of the difference should not change the results, since we are checking the absolute value.  In addition, the logic of this is incorrect. As it is, this will always evaluate to True is the cat's ages are different.  I assume this is the reason for the unit-test failing.

resources/game_config.json
```diff
@@ -81,7 +81,7 @@
 	},
 	"mates":{
 		"age_range": 40,
-		"overide_same_age_group": false,
+		"override_same_age_group": false,
```
keyraven, line 84: Good catch! Although this would not have resulted in the age cap being incorrectly applied, it does prevent the setting from working.

keyraven (changes requested):

keyraven (approved):

PR #1836, opened 2023-10-27
feat: Always halloween config

larkgz:
Adds `always_halloween` to `game_config.json` which tells all date functions that it's Halloween regardless of the actual date.

### DIFF
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -367,7 +367,8 @@
 		"april_fools": false,
 		"all_cats_are_newborn": false,
 		"newborns_can_roam": false,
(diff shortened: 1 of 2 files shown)


keyraven (approved):

archanyhm:
Looks good; @keyraven also approved this one. Can you @larkgz change the target branch of this PR to release-0.10? I think this would do well as a patch on the release version, given that the Halloween specific features are live.

Lixxis:
@archanyhm It was changed and we have to make sure this function is also available in the dev (+ future versions)

PR #1840, opened 2023-10-28
bugfix: loss of balance -> constantly dizzy

ZtheCorgi:
i believe "loss of balance" is an old name for "constantly dizzy", as the latter is a condition that exists while the former is...well, not. this should allow cats to actually get the "constantly dizzy" permanent condition from the "head damage" injury now


ZtheCorgi:
nvm PRing to release instead ig

PR #1848, opened 2023-10-29
Content: Basics of caretaker role

sedgestripe:
image
image
image
image
image


sedgestripe:
since there will no longer be a caretaker apprentice role, this pr will need a reworking

PR #1855, opened 2023-11-01
Fix debug mode from crashing on certain commands

tailhaver:
Fixes debug menu crashing when:
- clearing the console
- toggling / getting settings that don't exist
- changing the FPS cap


archanyhm:
@howlagon Bugfixes for bugs affecting the current release go to the release branch. In this case release-0.10 — unless this is dev exclusive. Please check & fix the target branch if necessary.

PR #1860, opened 2023-11-01
Fix debug mode from crashing on certain commands, PEP 8-ify

tailhaver:
- Fixes debug menu crashing when:
  - clearing the console
  - toggling / getting settings that don't exist
  - changing the FPS cap
- Makes negative numbers disable the FPS cap
- Makes debug files (debugCommands/*, debugMenu) follow the PEP 8 file naming convention


archanyhm:
I'll have a look at that tomorrow. Thanks for the re-submission!

Thlumyn:
@archanyhm @howlagon  Is this PR still active?

tailhaver:
it might be outdated, but I have no way of updating it as I am out of the country until next month. If it fixes the bugs with the debug mode in its current state, then it is active. Otherwise it can be closed.

tailhaver:
It would appear that this is still needed (these issues still appear on my fork of the latest commit)

Thlumyn (approved):

PR #1869, opened 2023-11-04
Feature: finish prey & nutrition system

Lixxis:
Activate the prey system for non-classic game.

* cats have nutrient information 
* cats now can get following conditions: **malnourished**, **starving**
* there is now a fresh-kill pile screen (e.g. clearing screen) with feeding tactics, feed single cats if they are hungry
* patrol filter for hunting, depending on season and biome
* special events around fresh-kill pile


resources/clansettings.json
```diff
@@ -119,6 +119,43 @@
             false
         ]
     },
+    "freshkill_tactics":{
+        "status wise":[
+            "status wise",
```
keyraven: I suggest the display names of some of these be changed.  When I first read this, I was confused what it meant, since I assumed "wise" to be it's own word, and not a suffix. 

I suggest "status-wise" or "by-status"

resources/clansettings.json
```diff
@@ -119,6 +119,43 @@
...
+            "The Clan will be fed according their status and within the status with age.",
```
keyraven: Small grammar fix:
"The Clan will be fed according their status and within the status ~~with~~by age."

scripts/screens/ClearingScreen.py
```diff
@@ -0,0 +1,683 @@
...
+from scripts.cat.cats import Cat
+from scripts.events_module.freshkill_pile_events import Freshkill_Events
+from scripts.game_structure.image_button import UISpriteButton, UIImageButton, UITextBoxTweaked
+from scripts.utility import get_text_box_theme, scale, shorten_text_to_fit
+from scripts.game_structure.game_essentials import game, screen_x, screen_y, MANAGER
+
+with open('resources/clansettings.json', 'r', encoding='utf-8') as f:
+    settings_dict = ujson.load(f)
+
+class ClearingScreen(Screens):
```
keyraven, line 15: Various notes on the clearing screen UI:

Some of the tab-buttons looks like they have been incorrectly sized, and the text is unclear. The size of buttons should always be set to 2x the pixel-dimensions of the image. 

Some text is being cut-off in the tactics list. The "g" in the "Feeding Tactic" title is also cut off.

scripts/cat_relations/inheritance.py
```diff
@@ -550,7 +550,7 @@ def get_not_blood_kits(self) -> list:
         """Returns a list of id's of kits, which are not related by blood to the cat."""
         return self.get_no_blood_relatives(self.kits)
 
-    def get_kits(self) -> list:
+    def get_children(self) -> list:
```
keyraven, line 553: Why was the name of this function changed?

scripts/patrol/patrol.py
```diff
@@ -793,64 +792,55 @@ def balance_hunting(self, possible_patrols: list):
...
+                    break
+
+            # now count the outcomes + prey size
+            prey_types = {}
+            for outcome in patrol.success_outcomes:
+                # ignore skill or trait outcomes
+                if outcome.stat_trait or outcome.stat_skill:
+                    continue
+                if outcome.prey:
+                    if outcome.prey[0] in prey_types:
```
keyraven, line 826: Will this break if more than one prey tag is ever used? At the moment, that isn't an issue, since there are only prey tags related to amount. But if other prey-tags are ever added, I imagine that might cause an issue on the balancing front.

scripts/patrol/patrol.py
```diff
@@ -793,64 +792,55 @@ def balance_hunting(self, possible_patrols: list):
...
-                    # if there is a tag set, check if this tag is not in the current patrol
-                    if not_allowed_tag and not_allowed_tag not in patrol.tags:
-                        filtered_patrols.append(patrol)
-                        break
-                    # when there is no tag set, add the patrol
-                    elif not not_allowed_tag:
-                        filtered_patrols.append(patrol)
-                        break
+            for adaption, needed_weight in PATROL_WEIGHT_ADAPTION.items():
+                if patrol.weight in range(needed_weight[0],needed_weight[1]):
```
keyraven: For speed and clarity, I suggest changing this line to: 
`if needed_weight[0] <= patrol.weight < patrol.needed_weight[1]`

`range(stop, end)` is sometimes unclear in this context, since it's inclusive of `stop` and exclusive of `end`. This syntax makes it the exclusivity/inclusivity clear, and it faster.

scripts/events.py
```diff
@@ -83,33 +84,22 @@ def one_moon(self):
...
-                    Single_Event(
-                        f"{game.clan.name}Clan doesn't have enough prey for next moon!"
-                    ))
-            if FRESHKILL_ACTIVE:
-                print(f" -- FRESHKILL: prey amount after feeding {game.clan.freshkill_pile.total_amount}")
-        
+                Freshkill_Events.handle_amount_freshkill_pile(game.clan.freshkill_pile, relevant_cats)
+            self.get_moon_freshkill()
+			# make a notification if the Clan has not enough prey
+            if not game.clan.freshkill_pile.clan_has_enough_food() and FRESHKILL_EVENT_ACTIVE:
```
keyraven: It might be worth moving the check for `FRESHKILL_EVENT_ACTIVE` first, so that `clan_has_enough_food()` doesn't need to run if `FRESHKILL_EVENT_ACTIVE` is `false`

resources/clansettings.json
```diff
@@ -119,6 +119,43 @@
...
+            false
+        ],
+        "less nutrition first":[
+            "less nutrition first",
+            "First the cats with the lowest nutrition will be fed, afterwards the Clan will be fed according to their status.",
+            false
+        ],
+        "hunter first":[
+            "hunter first",
+            "Firstly the hunter will be fed, afterwards the Clan will be fed according to their status.",
```
keyraven: From this explanation, I would assume that cats able to hunt (warriors and warrior apprentices) would be fed first.  However, this tactic works by feeding cats with the "HUNTER" skill first, than everyone else. I suggest rewording the hover-text blurb.

keyraven (changes requested):
I attempted to be thorough, although some of these comments may be nit-picky. Feel free to dispute any nit-picks! Overall, this is in pretty good shape.

Ryos00 (approved):

scripts/cat_relations/inheritance.py
Lixxis: I want it to be the same as in the cat class so it is more in sync / naming convention. When I implemented the feeding function I was confused why it is called get_kits in inheritance and get_children in the cat class.

scripts/patrol/patrol.py
Lixxis: I think this should break, if a new prey tag is added and the balancing dictionary + filtering is not updated. Otherwise, it might slip your eyes, and we can't find the issue of prey balancing. I'm not sure what tags you did have in mind, but when they change the amount of prey which can get caught, it should be added to the according dictionaries, otherwise balancing wouldn't make any sense.

scripts/patrol/patrol.py
Lixxis: A way better way 😂 I was too focused to see the confusing thing I made.

scripts/screens/ClearingScreen.py
Lixxis: Adapted the size.

The cutoff from Feeding Tactic is the same as with the Log, which is also in the medicine den, therefore I thought this is not as essential.

scripts/patrol/patrol.py
keyraven: I think this is OK for now, but we might consider rethinking the patrol-picking procedure to allow balancing to be built-in. I have a few idea, and I might start testing them out soon.

scripts/screens/ClearingScreen.py
keyraven: Okay, I think the cut-of title is fine, then. However, the sizes of the tab buttons are still slightly off. 

The "tactics" tab should be (70, 140), but it is currently set to (70, 128)
The "cats" tab should be (70, 150), but it is currently set to (68, 150)
The "logs" tab should be (70, 128), but it is currently set to (68, 128)
The "hungry" tab should be (160, 70), but it is currently set to (150, 70)
The "satisfied" tab should be (190, 70), bu is currently set to (170, 70)

keyraven (approved):

PR #1875, opened 2023-11-06
bugfix: is_potential_mate oversight

ZtheCorgi:
fixed an oversight within the is_potential_mate function that allowed mentors to gain romantic like with current apprentices over 14 moons old

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2098,6 +2098,10 @@ class Cat():
         is_former_mentor = (other_cat.ID in self.former_apprentices or self.ID in other_cat.former_apprentices)
         if is_former_mentor and not game.clan.clan_settings['romantic with former mentor']:
             return False
+        
+        #current mentor
+        if other_cat.ID in self.apprentice or self.ID in other_cat.apprentice:
+            return False
 
         return True
 


scripts/cat/cats.py
```diff
@@ -2098,6 +2098,10 @@ def is_potential_mate(self,
         is_former_mentor = (other_cat.ID in self.former_apprentices or self.ID in other_cat.former_apprentices)
         if is_former_mentor and not game.clan.clan_settings['romantic with former mentor']:
             return False
+        
+        #current mentor
+        if other_cat.ID in self.apprentice or self.ID in other_cat.apprentice:
+            return False
```
keyraven, line 2104: This same code is already present in lines 2094 - 2095.

keyraven (changes requested):

scripts/cat/cats.py
ZtheCorgi: oh. huh. then...i'm blind sorry. to be fair i was looking at lifegen code for like 5 hours

ZtheCorgi: i wonder why that issue happened then

PR #1876, opened 2023-11-07
Old age death chance rework

Snowstar38:
- Changed old_age_death_chance from a linear equation to an exponential one
- Replaced old_age_death_chance with old_age_death_curve, an adjustable value from 1-7, in game_config.json

Turns out that the average life expectancy was _less than 9 moons_ after hitting the old_age_death_start threshold of 150 moons!! This spreads out the death chances a little more and allows cats to potentially live past moon 200 (though the chance is very low).

It also helps with senior cats that join from events. Previously, if a cat that joined from an event was near or over 200, they would die on the first timeskip in which they were healthy. Now they have a decent chance of living a few moons within the Clan.

I'll be sharing more of the math and statistics in a comment below to explain the reasoning and balance behind this change.


Snowstar38:
Knowing that these are % chances that roll (and change) every single moon, I did a ton of statistical simulations on the different death rates, average death ages, standard deviations, etc, in sets of 50,000 each.

### Original (linear) equation, default settings:
chance $= \frac{moons - age start}{age chance}$
or
$y = \frac{x - 150}{51}$
#### Stats:
Average death age: 158.63 moons
Standard deviation of death age: 4.34 (about 2/3 of cats will die between 154.29 and 162.97 moons)
Oldest cat death age (out of 50,000, this was the oldest a cat lived to): 181 moons

| moon  | death chance |
|-------|--------------|
| 150   | 0            |
| 155   | 0.098039216  |
| 160   | 0.19607843   |
| 165   | 0.29411765   |
| 175   | 0.49019608   |
| 190   | 0.78431373   |
| 200   | 0.98039216   |
| 201   | 1            |

This means that, with an old_age_death_start of 150 and old_age_death_chance of 51 (the default), *50% of cats will die within 9 moons of hitting the age to qualify for death!* I know the number sounds absurd, but remember the cumulative risk of the chances being rolled every moon, and that by age 155, the chances are already 10% per moon (and 20% per moon by 160!).

### New (exponential) equation, default settings:
chance $= (1 + death curve)^{moons - age start} - 1$
or
$y = (1 + 0.0045)^{x - 150} - 1$
#### Stats:
Average death age: 168.11 moons
Standard deviation of death age: 9.13 (about 2/3 of cats will die between 158.98 and 177.24 moons)
Oldest cat death age: 216 moons

| moon | death chance |
|------|--------------|
| 150  | 0            |
| 155  | 0.022703413  |
| 160  | 0.045922272  |
| 165  | 0.069668277  |
| 175  | 0.11878987   |
| 200  | 0.25169078   |
| 225  | 0.40037897   |
| 250  | 0.56672982   |
| 275  | 0.75284145   |
| 300  | 0.96106127   |


Now things are a little more spread out. Cats still have a chance to die when they turn 150, but those chances increase more slowly. While the death chance at 200 moons is "only" 25%, we can see that the oldest cat out of 50,000 only lived to 216 due to the cumulative risk of rolling each moon.

Another positive about this new equation concerns newly recruited senior cats. Currently, cats over 200 moons can be recruited in events, but the linear death chance formula gives a 100% death chance for cats over 201 moons. This would lead to these new cats instantly dying on the next timeskip (if they were recruited injured, it will kill them the first timeskip after they heal). Now, if you recruit a 210-moon-old cat, the chance of them immediately dying is only 31% per moon instead of 100%! This way, new senior cats actually have a chance, albeit small, to hang on for a bit after joining the Clan.

<details> 
  <summary>For the curious, here are the statistics for the other values users can choose in game_config.json:</summary>

	Stats for exponential old_age_death_curve of 1 (rate 0.001):
	Average death age: 189.02 moons
	Standard deviation of death age: 19.98
	Oldest cat death age: 297 moons

	Stats for exponential old_age_death_curve of 2 (rate 0.002):
	Average death age: 177.41 moons
	Standard deviation of death age: 14.02
	Oldest cat death age: 248 moons

	Stats for exponential old_age_death_curve of 3 (rate 0.003):
	Average death age: 172.19 moons
	Standard deviation of death age: 11.32
	Oldest cat death age: 227 moons

	Stats for exponential old_age_death_curve of 4 (rate 0.0045):
	Average death age: 168.11 moons
	Standard deviation of death age: 9.13
	Oldest cat death age: 216 moons

	Stats for exponential old_age_death_curve of 5 (rate 0.008):
	Average death age: 163.46 moons
	Standard deviation of death age: 6.73
	Oldest cat death age: 193 moons

	Stats for exponential old_age_death_curve of 6 (rate 0.02):
	Average death age: 158.29 moons
	Standard deviation of death age: 4.03
	Oldest cat death age: 178 moons

</details>

scripts/events.py
```diff
@@ -1671,13 +1671,27 @@ def handle_injuries_or_general_death(self, cat):
...
+        death_curve_setting = game.config["death_related"]["old_age_death_curve"]
+        death_curve_multipliers = {
+            1: 0.001,
+            2: 0.002,
+            3: 0.003,
+            4: 0.0045,
+            5: 0.008,
+            6: 0.02
+        }
+        death_curve_value = death_curve_multipliers[death_curve_setting]
```
keyraven: I think limiting values to only 6 possible settings is non-optimal. game.config can be used by players, but it's primary purpose to allow easy re-balancing without changing any code.  Limiting the death_curve_setting to only six settings means that any small adjustments need to be made in the code itself.   

This particular line also introduces a crash if death_curve_setting is not an integer 1-6.  That can be fixed by using `get()` with an appropriate default value, but I would suggest the storing the actual value of the multiplier in config.

keyraven (commented):
I think these are some really good changes, and I appreciate the statistical calculations! Overall, I think this is a needed change. My comments are limited to the use of death_curve_setting.

scripts/events.py
Snowstar38: Good point! I have changed it to a more direct value now. It is multiplied by 0.001 in the actual formula, just to make it look neater, since the only viable ranges for the formula are between ~0.001 and ~0.02.

```python
        age_start = game.config["death_related"]["old_age_death_start"]
        death_curve_setting = game.config["death_related"]["old_age_death_curve"]
        death_curve_value = 0.001 * death_curve_setting
        # made old_age_death_chance into a separate value to make testing with print statements easier
        old_age_death_chance = ((1 + death_curve_value) ** (cat.moons - age_start)) - 1
        if random.random() <= old_age_death_chance:
            Death_Events.handle_deaths(cat, other_cat, game.clan.war.get("at_war", False), enemy_clan, alive_kits)
            return True
``` 

```python
		"old_age_death_start": 150,
		"old_age_death_curve": 4.5,
		"comment": [
			"old_age_death_curve is a multiplier that affects the chances of cats dying once they are over old_age_death_start moons.",
			"Average life expectancy with default old_age_death_start of 150 moons:",
			"1 = 189 moons, 2 = 178, 3 = 172, 4.5 (default) = 168, 8 = 163, 20 = 158"
		]
```

keyraven: Awesome, this looks great!

keyraven (approved):

PR #1883, opened 2023-11-08
feat: Remove accessory button

### LINKED ISSUE #1866: Add back “remove accessory” button
Apparently, there was previously a remove accessory button, but it was removed due to lack of space. It would be nice if it could be added back, since there seems to be more places the button could be put now.

larkgz:
Adds a button to cat toggles that removes the accessory of the selected cat. I'm not sure if this is the best place to put the button, but I also don't know where else it would go.

image

It's disabled on cats that have no accessory:
image


larkgz:
#1866

keyraven:
There is a pylint failure. In addition, I don't think this is the correct place for the replace accessory button, since cat toggles are meant for cat-behavior toggles, not single-use buttons. Maybe the button can be placed on the sprite inspect screen?

larkgz:
Don't know why Pylint failed? It passed on my fork and it also passes here after I forced a rerun with an empty commit. I guess the runner just crashed, but there wasn't an outage reported on that day or anything...

I'll see about moving the button. Hopefully it's not too confusing with the existing "hide accessory" button that's already on that screen?

EDIT: Actually, I think it might work under the "dangerous" tab, because it's an irreversible action like the other two.

keyraven (approved):

PR #1898, opened 2023-11-16
Typo Correction and button alignment fix

selkirks:
added various dark forest thoughts for apprentices, leaders, and warriors

fixed the 'retire' button not being aligned with the 'switch to warrior' button, reported here https://github.com/ClanGenOfficial/clangen/issues/1889


keyraven:
Reverted added thoughts (with permission), since this is going into release.

keyraven (approved):

PR #1931, opened 2023-11-29
Small fixes

Snowstar38:
- Pronoun tagging fixes for some leader deaths
- Fix for a rogue thought, now makes sure Clan cat is alive
- Fix for current/needed prey text covering the bottom half of the clickable element of the last cat on the patrol screen
- Added relationship gain tags to a specific event (saving a young cat from a river)


scripts/screens/PatrolScreen.py
Thlumyn, line 1: Can you talk about this change in your summary? Thanks (specifically the changes to PatrolScreen.py)

Thlumyn (changes requested):

Thlumyn (approved):

PR #1941, opened 2023-12-01
Fix to allow constraints in neutral interactions

Koriiy:
This is to fix constraints not applying in non-group neutral relationship interactions. It still allows neutral group interactions to work correctly as well.

### DIFF
--- a/scripts/cat_relations/relationship.py
+++ b/scripts/cat_relations/relationship.py
@@ -99,9 +99,10 @@ class Relationship():
         all_interactions = NEUTRAL_INTERACTIONS.copy()
         if in_de_crease != "neutral":
             all_interactions = INTERACTION_MASTER_DICT[rel_type][in_de_crease].copy()
-            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode, False)
+            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode)
         else:
-            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode, True)
+	    intensity = None
+            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode)
 
         if len(possible_interactions) <= 0:
             print("ERROR: No interaction with this conditions. ", rel_type, in_de_crease, intensity)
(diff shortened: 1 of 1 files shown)


scripts/cat_relations/relationship.py
```diff
@@ -413,7 +413,9 @@ def get_relevant_interactions(self, interactions : list, intensity : str, biome
             if len(in_tags) > 0:
                 continue
 
-            if interact.intensity != intensity:
+            if neutrality:
+                pass
```
Lixxis: You can make it as a one-liner, which would also increase the readability. (The "pass" might be confusing, even it is correct)
```
if not neutrality and interact.intensity != intensity:
    continue
```

scripts/cat_relations/relationship.py
```diff
@@ -99,9 +99,9 @@ def start_interaction(self) -> None:
         all_interactions = NEUTRAL_INTERACTIONS.copy()
         if in_de_crease != "neutral":
             all_interactions = INTERACTION_MASTER_DICT[rel_type][in_de_crease].copy()
-            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode)
+            possible_interactions = self.get_relevant_interactions(all_interactions, intensity, biome, season, game_mode, False)
```
keyraven: Rather than creating a new parameter into this function, I think it would be better to pass `None` for intensity in the case of neutral interactions. In the function itself, simply ignore the intensity check if intensity is `None`.

scripts/cat_relations/relationship.py
```diff
@@ -413,7 +413,9 @@ def get_relevant_interactions(self, interactions : list, intensity : str, biome
             if len(in_tags) > 0:
                 continue
 
-            if interact.intensity != intensity:
```
keyraven, line 416: In `None` is passed for intensity,  this line would be changed to `if intensity is not None and interact.intensity != intensity:`
We should avoid using `pass` if possible.

keyraven (changes requested):

scripts/cat_relations/relationship.py
Koriiy: Fixed for both! ty :)

Koriiy: Also sorry, still new to github so I had to fix formatting! Was having trouble editing directly on site so I cloned the folder - sorry about the amount of commits :)

Lixxis (approved):

keyraven (approved):

PR #1945, opened 2023-12-05
fix error in gen_border_otherclan_meeting1  text referencing nonexistant cat

### LINKED ISSUE #1940: r_c, error2, and missing "Clan" in scar text
I sent a cat on a solo border patrol and got gen_border_otherclan_meeting1; it ended up failing and my cat got a claw wound. Upon healing she got scarred and this was the text in her history:

Bug25

Apologies if I did this wrong, this is my first time making a bug report!

dreamboats95:
fixes https://github.com/ClanGenOfficial/clangen/issues/1940

(does not address o_c_n mismatch i haven't dug that deep into it yet, but it fixes the issue requester's error)


Thlumyn (approved):

Thlumyn (changes requested):
I think that that section was good to change; but I think this section in the other outcome needs changing as well (especially with the mismatch between s_c and r_c in this other instance; probably should use p_l in both cases, but the s_c/r_c issue might be why its fully throwing an error on the history)
image

Thlumyn (approved):

PR #1948, opened 2023-12-06
permanent condition thoughts for warriors

selkirks:
thoughts added for warriors with recurring shock and lasting grief conditions


Thlumyn (approved):

Lixxis:
I think some thought might also be suiting in the "general.json", just a recommendation.

PR #1961, opened 2023-12-12
Custom Pronouns & Presets

DeadGirlRisen:
Custom Pronouns & Presets. 

I messed up on the last one by accident, somehow merging my mod into it when I didn't mean to. This should be all the files on the clean clangen.

Its saying I changed stuff to patrols when I didn't, um, I think its becuase this copy is from a few commits behind? Im not sure how to get rid of those changes to the specific files...


DeadGirlRisen:
Okay, fixed, once more. This time, it should only be the actual files that were changed.

scripts/screens/ChangeGenderScreen.py
```diff
@@ -0,0 +1,485 @@
...
+        self.back_button = UIImageButton(scale(pygame.Rect((50, 120), (210, 60))), "", object_id="#back_button"
+                                         , manager=MANAGER)
+        self.update_selected_cat()
+    
+    '''def on_remove_button_click(self, pronoun_name):
+        print(f"Remove button clicked for pronoun: {pronoun_name}")'''
+        
+    def get_sample_text(self, pronouns):
+        text = ""
+        text += f"Demo: {pronouns['name']} <br>"
```
keyraven, line 190: Game in currently crashing on this line with a throw `Keyerror` when this page attempts to open. This only seems to occur on load-in clans.

keyraven (changes requested):
Thanks for all your work! This is great progress! 

General notes: 

- We would like to re-use button assets as often as possible. I noticed that you created new x-box and check-box button assets. However, we already have assets created for that type of button. Was there a reason those per-existing assets would not work here? If you can make them work, that would help keep down the total number of button assets. 
- What is the purpose of the "name" property for pronouns? This seems unnecessary, since pronouns are generally referred to with `subject/object`.   At the moment, it appears redundant.
- The general process to add a custom pronoun is confusing. I don't think it's finished yet, but you must open the "customize" dialogue, enter your pronouns, then click a button outside of the customize window to save the pre-set.  I also don't think it's clear that the "customize" window add a preset pronoun. As it's currently set up, I would expect it to add the pronoun directly.  The option directly above the pronouns (to change the gender) directly changes the cat's gender value. 
- There is a lot of unused blank space on the gender screen. It should be rearranged to be more balanced. Optional pop-up windows do not need their own dedicated white-space, since they can appear over other elements. 
- In general, I am struggling with the process to add custom pronouns. I am finding the process fairly unintuitive.  I know this is difficult nut to crack. In general, I would suggest grouping together all related function. At the moment, things are spread out. For example, to add a custom pronoun you need to use the customize dialogue, but removing or loading them is in a entirely separate place. 
- The option to "disable custom pronouns" should still allow the player to set pronouns for an individual cat.  The only thing that setting should do assign all cats they/them at generation. 
- Are you planning to store user-entered pronouns in pronouns.json? User data should never be placed in resources. Resources files are read-only data.  User data should always be placed in `saves`
- Custom pronouns should only be set by the user. The include `xhe` preset should be removed.

DeadGirlRisen:
> We would like to re-use button assets as often as possible. I noticed that you created new x-box and check-box button assets. However, we already have assets created for that type of button. Was there a reason those per-existing assets would not work here? If you can make them work, that would help keep down the total number of button assets.

the X button is still the same exit button to my knowledge, I didn't recreat the exit_window_button.

The checkmark coding for how to use it was confusing for us admittadly, on how to make it work. 

> What is the purpose of the "name" property for pronouns? This seems unnecessary, since pronouns are generally referred to with subject/object. At the moment, it appears redundant.

The name purpose is for identification purposes. In the code, the way you remove the pronouns is by identifying the name it has. So it's also a rather crucial part of the code. It's also an easy way to recognize which set of pronouns it is, if you have two similiar pronouns, or pronouns that might have the same subject/object but different poss and inposs for example.

> The general process to add a custom pronoun is confusing. I don't think it's finished yet, but you must open the "customize" dialogue, enter your pronouns, then click a button outside of the customize window to save the pre-set. I also don't think it's clear that the "customize" window add a preset pronoun. As it's currently set up, I would expect it to add the pronoun directly. The option directly above the pronouns (to change the gender) directly changes the cat's gender value.

To Bingle and I's understanding of the code, you can't input text outside of a window. So we had to make it a window, hence the customize button that we have.

We might be able to add those buttons inside the window, unless we can figure a way to input text outside of a window, but that was our bottleneck in this area.

> There is a lot of unused blank space on the gender screen. It should be rearranged to be more balanced. Optional pop-up windows do not need their own dedicated white-space, since they can appear over other elements.

Part of the issue above again was us being unable to have the text boxes outside of a window, I can see if Bingle and I can take a look at it again though to see if we can figure it out.

> In general, I am struggling with the process to add custom pronouns. I am finding the process fairly unintuitive. I know this is difficult nut to crack. In general, I would suggest grouping together all related function. At the moment, things are spread out. For example, to add a custom pronoun you need to use the customize dialogue, but removing or loading them is in a entirely separate place.

I'll talk to bingle, see what we can do for that. I think our issue there would be needing a second button in the containor area. But we might be able to figure it out.

> The option to "disable custom pronouns" should still allow the player to set pronouns for an individual cat. The only thing that setting should do assign all cats they/them at generation.

We have two options, assign cats they/them at generation which starts as true, and you can disable it so new cats come with pronouns.

The custom pronouns needs to be enabled to allow you to set specific pronouns, otherwise its basically just the previous gender alignment system.

> Are you planning to store user-entered pronouns in pronouns.json? User data should never be placed in resources. Resources files are read-only data. User data should always be placed in saves
Custom pronouns should only be set by the user. The include xhe preset should be removed.

Alright, I had the xhe included as an example of a custom preset, but that can easily be removed.

We will see if we can move it into the saves.

Thank you for reviewing it, and I'll let bingle know so we can figure out where to go from here.

DeadGirlRisen:
Alright, after winter holidays slowed down progress, I think Bingle and I are ready for another check.

1. Managed to get the checkmarks working.
 - The x key is the same as in game, however their was no disable exit_window button, which is what was added.
2. Name is for identification purposes in the code, as well as for having similiar but slightly different pronouns. Can not be changed, as the name works for the code to be able to remove the correct pronoun.
3. We managed to rework it so everything is a lot neater and less confusing now.
4. I believe this was also fixed when we reworked UI.
5. User entered custom pronouns are now saved in the saves within the clan file. Pronoun file is now purely for having default pronouns that are always available.

I feel we made some good leaps forward. Because of the winter break we may have missed something, but I reread the comments, just let me know what we can do and all ^^.

PR #1986, opened 2024-01-04
Bugfixing: set weight when creating PatrolEvent

sedgestripe:


Thlumyn (approved):

Lixxis (approved):

scripts/patrol/patrol.py
```diff
@@ -618,6 +618,7 @@ def generate_patrol_events(self, patrol_dict):
                 biome=patrol.get("biome"),
                 season=patrol.get("season"),
                 tags=patrol.get("tags"),
+                weight=patrol.get("weight"),
```
Lixxis:
```suggestion
                weight=patrol.get("weight", 20),
```

The default one is 20 so it should that and not 'None'

Lixxis (changes requested):

Lixxis (approved):

PR #1999, opened 2024-01-09
Bugfix: Name clan screen hover button

### LINKED ISSUE #1997: Clan naming screen buttons do not display hover state
All buttons on that screen that are overlayed on top of the image do not show a hover state when hovered.  Those buttons should have hover assets in file, so this is likely an issue relating to the layering of buttons over images.  As we don't have this issue on other screens of the game, this likely isn't a bug with pygame gui itself, but an issue in the coding of that screen.

Thlumyn:
Changes the name clan screen to use the name clan image as a background image instead of a GUI element to avoid GUI layering issues that caused the buttons to not recognize hovering. Fixes: #1997

### DIFF
--- a/scripts/screens/MakeClanScreen.py
+++ b/scripts/screens/MakeClanScreen.py
@@ -465,7 +465,7 @@ class MakeClanScreen(Screens):
                 self.elements["error"].hide()
                 self.elements['next_step'].enable()
             
-            # Set the background for the name clan page - done to here to avoid GUI layering issues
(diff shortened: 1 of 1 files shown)


Thlumyn:
Huh, no clue why the mac build here has taken _two hours_ to try and build... Not even sure why it's trying to build it yet. (edit: I forced it to rerun and it ran fine this time, so who knows)

scribblecrumb (approved):

PR #2014, opened 2024-01-12
Content: Dead Exiled Cat Thoughts

Thlumyn:
Adds a couple thoughts specific to dead exiled cats.


Tiritirimatangi (approved):
Love these, they'll make great additions to the game!

PR #2020, opened 2024-01-15
Fix #2004, pronoun tagging to new_cat_welcoming.json

### LINKED ISSUE #2004: loner kit generates with warrior name
**Type: ** text
**Summary:** a loner newborn generates with a warrior name. also queen is refered to with feminine prns
**Grade:** Unimportant
**Occurrence:** Permanent for this patrol
**Reproduce:** send out border patrols till you get gen_bord_welcomingsparrowgift1. proceed. this outcome generates a loner parent, clan newborns, and a loner newborn, with the loner newborn being given a warrior name
**Patrol ID + Outcome Print (if applicable):** gen_bord_welcomingsparrowgift1 | SUCCESS: True
**Commit #:** d2685a540cea30f1890abe4e8cdf5abd9374ba5c

image
image

tailhaver:
Fixes #2004 by manually appending "kit" or "paw" to loners if they are under a certain age. This fixes an edge case where a kit or app is a loner, giving them a proper name instead of a warrior name. This is implemented _after_ the standard suffix check for kits/apps, just as a fallback.

This also adds pronoun tagging to [new_cat_welcoming](https://github.com/ClanGenOfficial/clangen/blob/development/resources/dicts/patrols/new_cat_welcoming.json), my first pronoun tags! I'd prefer if someone looked over it just as a double check, although worst case someone will find it and (hopefully) report it.

### DIFF
--- a/resources/dicts/patrols/new_cat_welcoming.json
+++ b/resources/dicts/patrols/new_cat_welcoming.json
@@ -150,7 +150,7 @@
         ],
         "antag_success_outcomes": [
             {
-                "text": "p_l approaches the loner and tells themfirmly that newcomers aren't being taken in right now. The loner seems disappointed, but leaves anyway.",
+                "text": "p_l approaches the loner and tells them firmly that newcomers aren't being taken in right now. The loner seems disappointed, but leaves anyway.",
                 "exp": 10,
                 "weight": 20,
                 "outsider_rep": -2
--- a/resources/dicts/patrols/new_cat_welcoming.json
+++ b/resources/dicts/patrols/new_cat_welcoming.json
@@ -10,7 +10,7 @@
         "max_cats": 6,
         "min_max_status": {},
         "weight": 20,
-        "intro_text": "Your patrol finds a loner who is interested in joining the Clan. They banter with the cat a bit, making them feel more welcome.",
+        "intro_text": "Your patrol finds a loner who is interested in joining the Clan. They banter with the cat a bit, making {PRONOUN/n_c:0/object} feel more welcome.",
         "decline_text": "Your patrol decides to leave the loner be.",
         "chance_of_success": 60,
         "success_outcomes": [
@@ -32,7 +32,7 @@
                 "outsider_rep": 1
             },
             {
-                "text": "The patrol doesn't have to do much convincing. The loner seems happy to join the Clan with their litter of kits.",
+                "text": "The patrol doesn't have to do much convincing. The loner seems happy to join the Clan with {PRONOUN/n_c:0/poss} litter of kits.",
                 "exp": 10,
                 "weight": 5,
                 "new_cat": [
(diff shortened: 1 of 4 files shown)


Tiritirimatangi (commented):
Tiri has reviewed the typo and pronoun tagging fixes and would like to fall at your feet in gratitude - I'm not qualified to review the coding stuff though

resources/dicts/patrols/new_cat_welcoming.json
```diff
@@ -150,7 +150,7 @@
         ],
         "antag_success_outcomes": [
             {
-                "text": "p_l approaches the loner and tells them firmly that newcomers aren't being taken in right now. The loner seems disappointed, but leaves anyway.",
+                "text": "p_l approaches the loner and tells themfirmly that newcomers aren't being taken in right now. The loner seems disappointed, but leaves anyway.",
```
Thlumyn: Spelling (themfirmly)

scripts/cat/names.py
```diff
@@ -173,12 +175,16 @@ def give_suffix(self, pelt, biome, tortiepattern):
...
 
     def __repr__(self):
+        # Handles predefined suffixes (such as newborns being kit), then suffixes based on ages (fixes #2004, just trust me)
         if self.status in self.names_dict["special_suffixes"] and not self.specsuffix_hidden:
             return self.prefix + self.names_dict["special_suffixes"][self.status]
-        else:
-            if game.config['fun']['april_fools']:
-                return self.prefix + 'egg'
-            return self.prefix + self.suffix
+        if self.status in ['loner'] and self.moons is not None and self.moons < 12:
```
Thlumyn: Can we add `and not self.specsuffix_hidden` here? So that a loner with no suffix doesn't get paw or kit attached

Thlumyn: Moons <12 doesn't necessarily line up with being an apprentice; a lot of cases here where that wouldn't line up. Such as when players have exp based graduation turned on and the cats graduate based on exp. If you really wanted to, you could check for exp based graduation and check the cat's exp to see if they are really meant to be an apprentice. But that leaves a bit of code redundancy - if anything changes with the app/exp code it could mess up here. Maybe the apprentice graduation already has a function you could borrow for that. 
Though that might not fix things in the case of, like, Ravenpaw, where he kept his apprentice name his whole life after leaving. Not sure there's a good way to approach this.

Thlumyn (changes requested):

scripts/cat/names.py
tailhaver: This will only be used for loners outside the clan though, such as in one of the outcomes for `gen_bord_welcomingsparrowgift1` where you meet a loner kitten. It would appear as if loners don't gain any experience, and if they join the clan then they have their loner status stripped.
I personally don't see another way to fix this outside of potentially adding an `is_loner` flag to each cat, but that sounds unneeded.

Thlumyn (approved):

PR #2021, opened 2024-01-15
Cat List UI Adjustments

scribblecrumb:
- Cat List UI has been adjusted to add clarity for the user.
- All List screens are now controlled by one class instead of being spread across multiple.  It's now far far easier to change elements on all list screens at once, instead of having to change the same thing in 5 separate scripts (hell)
- Age Ascending filter option added to go along with old Age Descending filter.
- Filter button now clearly displays which filter is currently being used.
- Search input now transfers across list screen changes, i.e. if you search for a cat name on your clan list, you can switch to starclan and the search input will not reset as it used to.
- Favorite cat toggle is now displayed and functional on every single cat list screen (originally it only appeared on your clan list)
- If the user clicks on a cat in the cat list to view their profile and then clicks 'back' to return to the cat list, the 'back' button will now return them to the list they were previously viewing.


Thlumyn:
Unit test doesn't seem happy about this; I can take a closer look about that later this evening

scribblecrumb:
> Unit test doesn't seem happy about this; I can take a closer look about that later this evening

Yeah I'm trying to figure out why rn, if I don't manage it by the time you're free then definitely see if you can find the issue

scribblecrumb:
I'm at a bit of a loss, it seems to be complaining about the `insert_cat()` method, which I didn't even touch?  I'm not sure what the issue is.  I'm also not sure what's going on with the TypeError in `sort_cats()` as I can see any arguments that are missing.

Thlumyn:
@scribblecrumb
I made a little change to try and fix where it was failing the unit test

scribblecrumb:
> @scribblecrumb I made a little change to try and fix where it was failing the unit test

The tests are succeeding now, thank you!

Thlumyn (approved):

PR #2023, opened 2024-01-17
QOL: same sex setting update

Lixxis:
As discussed and voted, if the same sex kittens are turned on:
* single cats can get pregnant as well, 50/50 if they bring back a litter or not
* if it is a pair, gender doesn't define which partner gets pregnant, it is always the main cat, which had a successful "get pregnant" check

_sorry it took so long, though I would make other bug fixes_


Thlumyn:
I will look through this when I get home

Thlumyn (approved):

PR #2028, opened 2024-01-19
Custom pronouns & presets

DeadGirlRisen:
Scribble recommended making a draft with the branch.

- Adds pronouns into the game
- Allows users to pick which pronouns their cat uses, including multiple pronouns
- Allows user to create pronouns

Right now were mostly in the final phase, of needing to implement more pronoun tags into the text.

### DIFF
--- a/scripts/cat_relations/inheritance.py
+++ b/scripts/cat_relations/inheritance.py
@@ -534,7 +534,7 @@ class Inheritance():
                         "type": rel_type,
                         "additional": [add_info]
                     }
-                    self.all_but_cousins.append(inter_cat)
+                    self.all_but_cousins.append(inter_id)
                     self.all_involved.append(inter_id)
                 
 


scripts/cat/cats.py
```diff
@@ -93,12 +95,22 @@ class Cat():
             "conju": 2
         },
         {
+            "name": "Masculine",
```
keyraven: Please remove the "name" property on all pronouns, as it's not necessary, or displayed to the user.

scripts/cat/cats.py
```diff
@@ -93,12 +95,22 @@ class Cat():
...
             "subject": "he",
             "object": "him",
             "poss": "his",
             "inposs": "his",
             "self": "himself",
             "conju": 2
+        },
+        {
+            "name": "Custom",
```
keyraven: Custom pronouns are not default, and should not appear is a list of default pronouns.  Please remove the "custom" option from this list.

scripts/game_structure/windows.py
```diff
@@ -458,65 +459,312 @@ def process_event(self, event):
                 self.kill()
 
 
-class SpecifyCatGender(UIWindow):
-    """This window allows the user to specify the cat's gender"""
+class PronounCreation(UIWindow):
```
keyraven: If possible, I would like to prevent people from creating duplicate custom pronouns (or creating a custom pronoun that is the same as the defaults), by throwing a error message if attempted. This helps prevent some weird edge-cases with removing pronouns. It can be added later if need --> I wouldn't consider it a necessity for testing.

scripts/game_structure/windows.py
```diff
@@ -458,65 +459,312 @@ def process_event(self, event):
...
+        # Row 1
+        self.boxes["subject"] = pygame_gui.elements.UITextEntryLine(
+            scale(pygame.Rect((180, 290), (200, 60))),
+            placeholder_text=self.the_cat.pronouns[0]['subject'],
+            manager=MANAGER,
+            container=self)
+
+        self.boxes["object"] = pygame_gui.elements.UITextEntryLine(
+            scale(pygame.Rect((430, 290), (200, 60))),
+            placeholder_text=self.the_cat.pronouns[0]['object'],
```
keyraven: This is a minor issue, but I would suggest using `initial_text` rather than `placeholder_text`. `placeholder_text` looks identical to entered text, but the box is empty. If you click the "add" with an empty box, the task will quietly fail and not add a pronoun. This may be confusing to players, since it appears as if all boxes are full. `initial_text` actually places the text in the box. 

Additionally, I suggest always using they/them as the initial text, rather than the cat's current first pronoun. This isn't a huge deal, I just think it's better to be consistent.

scripts/game_structure/windows.py
```diff
@@ -458,65 +459,312 @@ def process_event(self, event):
...
+            container=self)
+
+        self.boxes["self"] = pygame_gui.elements.UITextEntryLine(
+            scale(pygame.Rect((550, 470), (200, 60))),
+            placeholder_text=self.the_cat.pronouns[0]['self'],
+            manager=MANAGER,
+            container=self)
+        
+        # Save Confirmation
+        self.pronoun_added = pygame_gui.elements.UITextBox(f"Pronoun saved and added to {self.the_cat.name}!",
```
keyraven: When you click the "add new pronoun" button, I would prefer if it only added it as a pronoun option, rather than also adding it to the cat.  This allows people to add a bunch of custom pronouns that they wish to use later without having to go through the remove it from the cat. Additionally, it would mean you no longer have to pass the cat object as a parameter into the window constructor.

resources/gamesettings.json
```diff
@@ -65,6 +65,11 @@
             "Randomize relationship values when creating Clan",
             "Clan founder cats will start the game with established relationships.",
             true
+        ],
+        "they them default": [
+            "Use they/them as default pronouns",
+            "If this setting is on new cats will generate with they/them pronouns, regardless of gender.",
+            true
```
keyraven: The default for this setting should be `false`. This is a huge project -- it shouldn't be hidden behind a non-default toggle! :)

scripts/cat/cats.py
```diff
@@ -279,66 +291,98 @@ def __init__(self,
...
-                    self.genderalign = self.gender
-            elif self.gender == "male" and not self.status in ['newborn', 'kitten']:
-                if trans_chance == 1:
-                    self.genderalign = "trans female"
-                elif nb_chance == 1:
-                    self.genderalign = "nonbinary"
+                    they_them_default_setting = False
+                    print("Game is not using they/them pronouns")
+
+                if they_them_default_setting is True:
```
keyraven: This works well, but I would like to clean it up and make it a bit more readable, and less of a maze of if-statements. Not high priority.

keyraven (commented):
Looking really, really nice! Mostly minor comments.

scripts/screens/ChangeGenderScreen.py
```diff
@@ -0,0 +1,631 @@
...
+                "subject": "",
+                "object": "",
+                "poss": "",
+                "inposs": "",
+                "self": "",
+                "conju": 1
+            }
+        ]
+        self.remove_button = {}
+        self.checkboxes_text = {}
```
scribblecrumb: Since this screen no longer utilizes checkboxes, it would be good to remove all references to checkboxes.  Otherwise it becomes difficult for others to easily understand what your code is referencing.

PR #2036, opened 2024-01-20
Guide Now Appears First In List

### LINKED ISSUE #2045: New sorting causes weirdness with switching between cats
image
image
image
Type: Code
Summary: Guide shows where they are in the ranks on the afterlife pages but when you click on the guide and press next cat, the game treats them like they're still at the top of the list.
Grade: Non-vital
Occurrence: Permanent
Reproduce: Have your guide be anything other than a leader and sort by ranks. Then click on the guide.
Commit #: 272eac575c0c16cc1f253476183b71d2fef24197

scribblecrumb:
- Adjusted cat list sorting to ensure that the guide is ignored by the sorting.  Guide now always appears first in their respective list.


scripts/screens/ListScreen.py
```diff
@@ -538,6 +538,19 @@ def update_search_cats(self, search_text):
         """Run this function when the search text changes, or when the screen is switched to."""
         self.current_listed_cats = []
         Cat.sort_cats(self.full_cat_list)
+
+        # adding in the guide if necessary, this ensures the guide isn't affected by sorting as we always want them to
+        # be the first cat on the list
+        if self.current_group == 'df' and game.clan.instructor.df:
```
Thlumyn, line 544: You can probably turn this into one if statement; like if (group == df and instructor.df) or (group == sc and not instructor.df) 
Instead of repeating the same code there twice

Ryos00 (approved):

PR #2049, opened 2024-01-23
Alternate Residence Toggle

DeadGirlRisen:
Allows a cat to be sent to the opposite of where the instructor is.


keyraven (commented):
I am wary of adding this toggle. You are already able to send a cat to the Dark Forest once they die, so this is not adding any additional functionality.  We only have so much room to add additional cat toggles, and this takes up another slot. 

I also believe the implementation is a bit confusing to the player.  If we want to give players direct control over which afterlife cats go to before the cat dies, then that choice should be clear - for example, letting them select from a drop-down.  However, my general opinion is that this functionality is not necessary.

PR #2060, opened 2024-01-25
Bugfix: Add schema to poetry

### LINKED ISSUE #2057: Schema not in builds
Hi! Seems schema is not bundled in builds currently:
schema
This was noticed by a few users that regularly use dev builds, so I tested it myself on Linux as well

Thlumyn:
Attempting to fixes: #2057 

Not sure why poetry is removing the main & dev tags - is that because of the newer version of poetry? 
I'd appreciate advice on it before merging


archanyhm:
Not quite sure about the category fields there — however, the lockfile is automatically generated by Poetry and should not be manually modified in the first place, so I think it's fine if we treat Poetry's output as single source of truth here.

archanyhm (approved):
Not quite sure about the category fields there — however, the lockfile is automatically generated by Poetry and should not be manually modified in the first place, so I think it's fine if we treat Poetry's output as single source of truth here.

For the record: Both source and binary versions of ClanGen run again with this fix.

PR #2065, opened 2024-01-26
fix dead cat relationships + patrol fixes

selkirks:
a fix in gen_train_deputydoubt1 in which a leader was regarded as a deputy. reported here: https://github.com/ClanGenOfficial/clangen/issues/2001

a fix in which you were able to view relationships of dead cats. reported here: https://github.com/ClanGenOfficial/clangen/issues/1971

a fix in which a loner medicine cat generated as an adolescent. reported here: https://github.com/ClanGenOfficial/clangen/issues/2015


Tiritirimatangi (commented):
The patrol fixes look good! I can't comment on the coding fix, sorry

sedgestripe (approved):

resources/dicts/patrols/new_cat_welcoming.json
```diff
@@ -992,7 +992,7 @@
                 "weight": 20,
                 "outsider_rep": 0,
                 "new_cat": [
-                    ["meeting", "loner"]
+                    ["status:medicine cat","meeting", "loner"]
```
Lixxis: This line will not change the status of the cat. A cat outside the clan will get the status "loner", even with that added tag. I tested it out and looked into the save of the cat. To assure that this cat get the "medicine cat" after deciding to join the clan we would have to look at the background and assigning a status according to that.

selkirks: do i assign one of the med cat backstories?

Lixxis: In my test run it assigned the "wandering_healer2" backstory.

Ryos00 (approved):

PR #2098, opened 2024-02-17
rat king text fixes + others

yippeecore:
fixed a typo ('being' to 'bring'), and capitalized all mentions of Rat King. also changed the death results to say that the cat was killed by 'a' rat king rather than 'the' rat king, since the scarred results use 'a'. | originally reported here: https://github.com/ClanGenOfficial/clangen/issues/1818#issuecomment-1874356240

### DIFF
--- a/resources/dicts/events/ceremonies/ceremony-master.json
+++ b/resources/dicts/events/ceremonies/ceremony-master.json
@@ -2269,7 +2269,7 @@
       "grumpy",
       "gloomy"
     ],
-    "(old_name) remains completely quiet as l_n names {PRONOUN/m_c/object} m_c after {PRONOUN/m_c/poss} r_h. Before anyone can congratulate {PRONOUN/m_c/object}, however, {PRONOUN/m_c/subject} {VERB/m_c/turn/turns} and walk away without a word to do {PRONOUN/m_c/poss} silent vigil."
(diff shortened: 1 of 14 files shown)


resources/dicts/events/ceremonies/ceremony-master.json
```diff
@@ -2269,7 +2269,7 @@
       "grumpy",
       "gloomy"
     ],
-    "(old_name) remains completely quiet as l_n names {PRONOUN/m_c/object} m_c after {PRONOUN/m_c/poss} r_h. Before anyone can congratulate {PRONOUN/m_c/object}, however, {PRONOUN/m_c/subject} {VERB/m_c/turn/turns} and walks away without a word to do {PRONOUN/m_c/poss} silent vigil."
+    "(old_name) remains completely quiet as l_n names {PRONOUN/m_c/object} m_c after {PRONOUN/m_c/poss} r_h. Before anyone can congratulate {PRONOUN/m_c/object}, however, {PRONOUN/m_c/subject} {VERB/m_c/turn/turns} and walk away without a word to do {PRONOUN/m_c/poss} silent vigil."
```
Ryos00: I think "walk" should actually be verb tagged as well

Ryos00 (changes requested):

resources/dicts/events/ceremonies/ceremony-master.json
yippeecore: done!

Ryos00 (approved):

Thlumyn:
@Tiritirimatangi  Are we using themself or themselves as the object form of they? I personally think themselves is a bit more grammatically correct but it's more important that we're just consistent with the use throughout the rest of the game.

Tiritirimatangi:
> @Tiritirimatangi Are we using themself or themselves as the object form of they? I personally think themselves is a bit more grammatically correct but it's more important that we're just consistent with the use throughout the rest of the game.

@Thlumyn Themself is the current version - but obviously we can easily edit that in utility.py via pronoun tagging if we'd like to change that

PR #2101, opened 2024-02-18
Enforce pypoetry

### LINKED ISSUE #2061: Making Poetry mandatory
Hi folks,

ever since I introduced the compatibility fix for macOS 10.13 back when I first joined the team, we've started to rely on external package/environment managers for Python — which eventually ended up including Poetry for virtual environment management.

As of now, we have both a `requirements.txt` and a `pyproject.toml` which _should_  in theory contain the same dependencies, but in practice they go out of sync if one developer forgets about this "constraint". To match our development and testing environments with our build environment, I'm phasing out the `requirements.txt` file and make using Poetry mandatory.

Also see: #2057

archanyhm:
closes #2061


poetry.lock
```diff
@@ -2,24 +2,24 @@
 
 [[package]]
 name = "altgraph"
-version = "0.17.3"
+version = "0.17.4"
```
Lixxis, line 5: version update on that package was intended?

Lixxis (commented):
In general, are all the version changes are intended? (safety proof)

poetry.lock
archanyhm: All updates are according to the specifications in the pyproject.toml; just the lock file got updated and that on purpose, yeah. I think that hasn't happened in a while.

larkgz:
Can you also modify the CONTRIBUTING.md so it’s consistent?

archanyhm:
@larkgz 
> Can you also modify the CONTRIBUTING.md so it’s consistent?

Done!

CONTRIBUTING.md
Thlumyn, line 1: Extra comma

Lixxis (approved):
I don't see any issues, but didn't run the workflows. 
I saw you tested with another developer in discord.

PR #2103, opened 2024-02-18
small addition - game_config.json

ZtheCorgi:
adds a comment that clarifies what override_same_age_group does


Thlumyn:
Can you PR this into the development branch instead of the release branch? We are getting close to another release and I'd like to avoid having to merge the old release branch back into development if possible.

PR #2125, opened 2024-03-04
Update LICENSE.md

Thlumyn:
Updating the license description to remove some unclear wording - 

Defining which license writing falls under. (When the license was originally written, these were hardcoded in and not separated into resource files). 

Clarifying who gets attributed in our CC license.


Thlumyn:
Do not merge yet, this is still under discussion in development channels

PR #2127, opened 2024-03-05
Remove reference to requirements.txt from main.py

Thlumyn:
Reinstalling all my dependencies on a new computer, tried to run it without and got this error message that gives outdated instructions for installing using requirements.txt.

### DIFF
--- a/main.py
+++ b/main.py
@@ -44,7 +44,7 @@ if not getattr(sys, 'frozen', False):
     if isMissing:
         print("""You are missing some requirements to run clangen!
                 
-                Please look at the "readme.md" file for instructions on how to install them.
+                Please look at the "README.md" file for instructions on how to install them.
                 """)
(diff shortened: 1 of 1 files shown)


main.py
```diff
@@ -42,18 +42,10 @@
...
-                  """)
-        else:
-            print("""You are missing some requirements to run clangen!
-                  Please run the following command in your terminal to install them:
-                  
-                  python3 -m pip install -r requirements.txt
-                  """)
+        print("""You are missing some requirements to run clangen!
+                
+                Please look at the "readme.md" file for instructions on how to install them.
```
keyraven: README.md is in all caps - it is probably best to match that here.

keyraven (changes requested):

keyraven (approved):

PR #2135, opened 2024-03-11
Thought Fixes and Some New Thoughts

DeadGirlRisen:
This fixes up the traits in the kitten file and a few others that were incorrect after the trait changes.

Also removes altruistic, empathetic, and patient from the ceremony traits. Patient traits were added to calm, and Altruistic and Empathetic were added to Compassionate.

### DIFF
--- a/resources/dicts/thoughts/alive/apprentice.json
+++ b/resources/dicts/thoughts/alive/apprentice.json
@@ -213,19 +213,14 @@
         ]
     },
     {
-        "id": "patient_warrior_apprentice",
+        "id": "calm_warrior_apprentice",
         "thoughts": [
-            "Is busy mastering a fighting technique",
-            "Spent all morning mastering and practicing the hunting crouch",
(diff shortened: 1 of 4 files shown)


resources/dicts/thoughts/alive/apprentice.json
```diff
@@ -263,12 +258,35 @@
...
             "Is bored during a hunting patrol",
+            "Is busy mastering a fighting technique",
             "Is in trouble after scratching r_c during an argument"
         ],
         "main_trait_constraint": [
             "fierce"
         ]
     },
+    {
+        "id": "competitive_warrior_apprentice",
```
Thlumyn, line 269: These are new thoughts, rather than bug fixes? If so we might want to wait until after the next release to add them.

DeadGirlRisen: The first one isn't new, I think the second two might have been though. The first one I believe was shifted from another area to competitive. I can remove the other two.

Ryos00 (approved):

PR #2142, opened 2024-03-13
Add missing module list to missing requirement message

tailhaver:
Created to move the discussion from discord#dev-game-talk to Github so more people can talk on the topic, if desired. Tested on 3.11[.8], 3.12[.2]
This simply adds a couple extra lines to show which modules specifically are missing, as well as renames variables relating to the check to follow PEP 8 standards
image
[start of original thread](https://canary.discord.com/channels/1003759225522110524/1005586496142704800/1217329065086947411)

### DIFF
--- a/main.py
+++ b/main.py
@@ -53,7 +53,6 @@ Please look at the "README.md" file for instructions on how to install them.\n""
 
     del required_modules
     del is_missing
-    del new_line
 del find_spec
 
 from scripts.housekeeping.log_cleanup import prune_logs
--- a/main.py
+++ b/main.py
@@ -24,7 +24,7 @@ import threading
 from importlib.util import find_spec
 
 if not getattr(sys, 'frozen', False):
-    requiredModules = [
+    required_modules = [
         "ujson",
         "pygame",
         "pygame_gui",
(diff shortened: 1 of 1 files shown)


Thlumyn:
I just read through the discord discussion.

Honestly, I think this is a good idea to be more specific in these errors. I know that we only intend for contributors to run off source code, but even then they might not all know too well how to do things - everyone starts somewhere and I like to think of ClanGen as a project that is beginner friendly. Maybe add another note before "Please look at README.md" saying something along the lines of "You should not be seeing this if you are running the code through Poetry, which is the only supported way to run this code from source." Or something like that. Just as an explanation before people just go and try and install those dependencies globally without reading the README

Some more detail about getting things to run through Poetry might be helpful as well, I know we are trying to stay IDE agnostic and that Thonny isn't really good for a development platform, but even something like "how to change environment in Visual Studio Code" might be helpful for some folk. 

Hopefully this all isn't as much of an issue once the .bat files work more consistently. 

@archanyhm

Thlumyn:
If you can add another note before "Please look at README.md" saying something along the lines of "You should not be seeing this if you are running the code through Poetry, which is the only supported way to run this code from source." Then I would be fine with approving this

Thlumyn:
I'm closing this as stale, please reopen if you are interested in continuing this.

PR #2149, opened 2024-03-14
Bugfixes: FKP, History text issues, Inheritance, settings description

### LINKED ISSUE #2132: error2 in history
image
upon a cat getting scarred from proceeding on bch_bord_rogue1 | SUCCESS: False, their history displays error2

### LINKED ISSUE #2133: error2s in history
image
upon proceeding on PATROL ID: gen_bord_smalldog3 | SUCCESS: False and having a cat get scarred, their history shows error2

### LINKED ISSUE #2140: Kits dying of hunger when they should have priority
**Type:** Code
**Grade:** Important
**Commit #:** 44699b314
**Summary:**
I have the feeding tactic set to "Hunters first", which in theory should give priority to cats with hunter skill first, and then to kittens (according to status). Current prey amount (before timeskip) is 29, needed amount is 32. My Clan has 3 cats with the hunter skill (one of them is the medicine cat) and 10 kits (whose current nutrition status is "hungry"). Upon timeskip, all 10 of the kits die of hunger, consistently (I tried a few times with the same result). My impression is that, under this tactic, kits are not being fed at all.

### LINKED ISSUE #2144: Adopted kitten isn't shown in their adoptive parents' family trees
when playing with 'allow same-sex breeding' toggled off, a same-sex pair who adopted a kit via timeskip don't show the kit in either of their trees - however, the kit's tree shows that they have parents just fine.

discovered while playing this commit - 44699b314005049f2c09e60cc315e4115e8b0b22
image
image
image
image

### LINKED ISSUE #2148: change name of same sex bio kits setting
as i recall, it was desired to change the name of the setting to reflect that it will also enable male cats in m/f relationships to become pregnant, not just m/m relationships. i believe the proposed setting name was smth like "pregnancy ignores biology"?

Lixxis:
FKP:
fix #2140

History text:
fix #2133 
fix #2132

Inheritance:
fix #2144 

Settings:
fix #2148

### DIFF
--- a/resources/clansettings.json
+++ b/resources/clansettings.json
@@ -90,8 +90,8 @@
             false
         ],
         "same sex birth": [
-            "Allow same-sex breeding",
-            "Allow couples to birth kittens despite same-sex status.",
+            "Pregnancy ignores biology",
+            "Allow all cats to get pregnant despise their gender and all couples to birth kittens despite same-sex status.",
             false
         ],
         "same sex adoption": [
--- a/scripts/clan_resources/freshkill.py
+++ b/scripts/clan_resources/freshkill.py
@@ -81,6 +81,7 @@ class Freshkill_Pile():
             self.total_amount = game.prey_config["start_amount"]
         self.nutrition_info = {}
         self.living_cats = []
+        self.already_fed = []
         self.needed_prey = 0
 
     def add_freshkill(self, amount) -> None:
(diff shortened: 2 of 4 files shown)


ZtheCorgi:
for the ignore biology setting, i suggest the description to be "Allow cats to become pregnant or get others pregnant regardless of sex."

if the point about the relationships is important, maybe something like "Allow cats of either sex to become pregnant, even in same-sex relationships."

Thlumyn (approved):

Ryos00 (approved):

PR #2154, opened 2024-03-20
credits list update

Ryos00:
just updated the credits list to the current accurate info


Thlumyn:
Two things: 

1. Did you mean to remove Eshame & Mojito?
2. This is more of a personal request, but can you put (thlumyn) after my name in the credits? Just so it might be less confusing for people

Ryos00:
> Two things:
> 
>     1. Did you mean to remove Eshame & Mojito?
> 
>     2. This is more of a personal request, but can you put (thlumyn) after my name in the credits? Just so it might be less confusing for people

eshame and mojito were changed to the names they wrote as preferred on the credits sheet ^^ they're not gone, just different names

and yes, added thlumyn beside yours

Thlumyn (approved):

PR #2173, opened 2024-03-30
Fix "not_pl_rc"

### LINKED ISSUE #2171: Patrol Constraint not working
"can_have_stat": ["not_pl_rc"], does not work. 

Because there is no current "o_c1" for can have stat, I was testing out the can have stats. I discovered that the not_pl_rc tag doesn't work at all.

(I wanted to be able to specify that either r_c or o_c1 could have the stat, but the patrol leader couldn't, and learned there was no o_c1 tag for the can have stat)

image
image
image

Another thing in my testing that needs looking into is when you have a cat that can fit into two tags. In this example, instead of Curlycatcher being selected as Random Cat, it used Spiderpaw twice, so there should probably be a check that if they can fit into another role like app1, that they can't be used for the random cat.

image
image

keyraven:
Fixes #2171 . Typo Issue

### DIFF
--- a/scripts/patrol/patrol_outcome.py
+++ b/scripts/patrol/patrol_outcome.py
@@ -208,7 +208,7 @@ class PatrolOutcome():
         
         # With allowed_specfic empty, that means the stat can can be anyone that's not patrol leader
         # or stat cat. This can
-        if not allowed_specfic or "no_pl_rc" in allowed_specfic:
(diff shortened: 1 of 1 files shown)


Tiritirimatangi (approved):
Even I understand what Key did to fix this, even with my limited knowledge

PR #2253, opened 2024-04-25
Add support for 'not_working' thoughts (#2233)

### LINKED ISSUE #1956: Medicine cat apprentice caught a mouse in thought
Type:  Text
Summary: Marigoldpaw "caught the biggest mouse ever!" despite being a medicine cat apprentice so not going hunting. She's also sick with whitecough and as such probably shouldn't be going out of camp. Marigoldpaw please calm down. 
She's also the only medcat in the clan right now. I don't think that's what's causing this but I thought I'd mention it. 
 Grade:  Unimportant
Occurrence: Rare
Reproduce: Check a medicine cat apprentice's thoughts until you see this one
Commit: 7a10bb4e

Screenshot 2023-12-09 at 12 57 36 AM

### LINKED ISSUE #2233: Add "Healthy" Parameter to Thoughts
We need a way to distinguish between thoughts for healthy cats and thoughts for sick/injured cats.  The proposed solution is adding a "Healthy" parameter to the thought format that would be a bool.  True would be the default and would mean the thinker is healthy.  False would mean the thinker is sick/injured.

A coder that takes this on should coordinate with a writer to have at least a few "sick/injured" thoughts prepared for use with this feature.

The end goal here is to prevent cats who are sick/injured from having thoughts about sparring, going on patrol, or other activities that they would not realistically be doing.

jdowner:
This change allows thoughts to be selected based upon a boolean attribute in the thought format called 'not_working'. The 'not_working' status of a cat is an existing feature that returns `true` if the cat has anything more than a minor illness/injury and `false` otherwise.

### DIFF
--- a/resources/dicts/thoughts/alive/general.json
+++ b/resources/dicts/thoughts/alive/general.json
@@ -82,6 +82,59 @@
         ]
     },
     {
+        "id": "gen_alive_not_working_insteadinthemedden",
+        "thoughts": [
+            "Misses {PRONOUN/m_c/poss} own nest, despite the ones in the medicine den being... fine",
+            "Can't get the sharp smell of the herb stores out of {PRONOUN/m_c/poss} nose",
+            "Feeling confined and cooped up",
+            "Just wants to sleep",
+            "Why is recovering so exhausting!",
+            "Basking in the sun at the medicine den entrance",
+            "Feeling frustrated at {PRONOUN/m_c/poss} own exhaustion",
+            "Doesn't think {PRONOUN/m_c/subject} should have to sleep in the medicine den",
+            "Eats some prey because {PRONOUN/m_c/suject} know {PRONOUN/m_c/subject} should, not because of hunger",
+            "Stands up for a second and feels lightheaded",
+            "Isn't having a fun time"
+        ],
+        "main_status_constraint": [
+            "not_working"
+        ]
+    },
+    {
+        "id": "gen_alive_not_working_insteadinthemedden_withmedcat",
+        "thoughts": [
+            "Watches r_c fuss around the herb stores, half-asleep",
+            "Isn't goign to argue with r_c about {PRONOUN/m_c/poss} medical care",
+            "Is going to argue with r_c about {PRONOUN/m_c/poss} medical care",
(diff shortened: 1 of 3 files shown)


scribblecrumb:
This looks good to me!  Feel free to ping the Writing Contributors role, or just Tiri herself, to find a volunteer writer to add some not_working thoughts.

Tiritirimatangi (changes requested):
I'll get right on to generating some thoughts for you - but i do have a question. The "not_working" tag means the cat has to be in the medicine den, right? Have you coded in an exception for pregnant cats, who stop working to move into the nursery and whose not_working thoughts should probably read different to an injured or ill cat?

Tiritirimatangi:
I also think this paramter would be exceptionally helpful to have active in the events code too, to encourage writer to give us relationship events that occur between a medicine cat (or med cat app) and their patients, that'd be super cool

scribblecrumb:
> I'll get right on to generating some thoughts for you - but i do have a question. The "not_working" tag means the cat has to be in the medicine den, right? Have you coded in an exception for pregnant cats, who stop working to move into the nursery and whose not_working thoughts should probably read different to an injured or ill cat?

There's actually an annoying amount of exceptions to the "major+ severity stays in med den" rule (the med den screen is annoying in having to track them all, ugh).  If we're going to have that going on in multiple scripts, then we need to change how it's tracked in the med den screen as well, otherwise we're setting ourselves up for annoyance later.  Personally, I'd prefer if writers simply treated not_working thoughts as "cat is staying in camp, but not necessarily immobile or in the med den".  If you want to get more specific than that, the thought format already has "has_injury", which lets you specify that the cat must have a certain condition in order to get a thought.  Not the most flexible, but it might be the simplest.

scribblecrumb:
> I also think this paramter would be exceptionally helpful to have active in the events code too, to encourage writer to give us relationship events that occur between a medicine cat (or med cat app) and their patients, that'd be super cool

Since I'm already reformatting moon events, I can look into including this there.  Though, they would be misc moon events that affect relationships, not specifically relationship events.  If you'd like that for relationship events specifically, then that'd be best made as a new enhancement issue as I'm not likely to get into that end of the code.

Lixxis (approved):

Tiritirimatangi (approved):

PR #2269, opened 2024-04-27
Specify death in mass extinction events

KiraDragoness:
Added specific lines in history tab when a cat dies because of a mass extinction event.


scripts/events.py
```diff
@@ -2129,7 +2129,28 @@ def handle_mass_extinctions(self, cat):  # pylint: disable=unused-argument
             for poor_little_meowmeow in dead_cats:
                 poor_little_meowmeow.die()
                 # this next bit is temporary until we can rework it
-                History.add_death(poor_little_meowmeow, 'This cat died after disaster struck the Clan.')
+                if event_string == f'{names} drown after the camp becomes flooded.':
```
Lixxis: All those checks will work if there are multiple cats dying, but not if a single one dies. In line 2118-2123 some words are replaced, so the event string will work for a single cat dying. 

I would suggest to work with regex, there you can cover both ways at the same time. If you want to use it, you have to add "import re" at the top and this would be an example for the if statement:
```suggestion
                if re.search("(drown|drowns) after the camp becomes flooded.", event_string):
```

Lixxis (changes requested):

scripts/events.py
KiraDragoness: I didn't even realize there was the possibility for just one cat to die! I made the changes you suggested and tested it to make sure I didn't break anything.

KiraDragoness: Wait, I've messed up something merging to my personal branch, let me fix it

PR #2277, opened 2024-04-28
Fix: Prevent duplicate prefixes in the same litter

### LINKED ISSUE #2268: [BUG] 2 kits from the same litter get the same name
**Type**: Code

**Describe the bug**
Two kits that have been born within the same litter receive the same Prefix.

**Grade**: Non-vital

**Reproduce**
Skip moons until a new litter has 2 kits with the same name.

**Screenshots**
image
image
image

joesandmeyer:
Prevent kits in the same litter from being born with the same name.


Lixxis (approved):

Lixxis:
@joesandmeyer you can add the issues in the PR section "Development", then they will be closed automatically and you don't have to mention it as a comment at the issue ticket \info
image

PR #2278, opened 2024-04-28
Specify death in mass extinction events

KiraDragoness:
Added specific lines in history tab when a cat dies because of a mass extinction event. 
Incorporated Lixxis suggestion from my previous PR.

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -7,10 +7,10 @@ TODO: Docs
 """
 
 # pylint: enable=line-too-long
+import re
 from collections import Counter
 import random
 import traceback
-import re
 
 from scripts.cat.history import History
 from scripts.patrol.patrol import Patrol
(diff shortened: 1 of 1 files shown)


scribblecrumb (approved):

scripts/events.py
```diff
@@ -2129,7 +2129,29 @@ def handle_mass_extinctions(self, cat):  # pylint: disable=unused-argument
             for poor_little_meowmeow in dead_cats:
                 poor_little_meowmeow.die()
                 # this next bit is temporary until we can rework it
-                History.add_death(poor_little_meowmeow, 'This cat died after disaster struck the Clan.')
+                import re
```
Lixxis: imports should be done in the first lines of a file (cleaner structure) just put it at the top to the other imports \info

Lixxis (changes requested):

scripts/events.py
KiraDragoness: Done!

Lixxis (approved):

PR #2303, opened 2024-05-02
On-death thought rework

Snowstar38:
- Death thoughts have their own .json and are now no longer in cats.py
- Now others can introduce additional variants for each type of death thought
- Also adds pronoun tagging for ondeath thoughts as this was not possible before


scribblecrumb:
Leader death thoughts don't seem to be working.  I tried having my leader die both naturally via timeskip and manually via profile, either one gave a death thought.  Regular cats do seem to work fine, though.

scribblecrumb (changes requested):
Leader death thoughts don't seem to be working. I tried having my leader die both naturally via timeskip and manually via profile, neither one gave a death thought. Regular cats do seem to work fine, though.

Snowstar38:
Ah, I just saw this. When I get home I will take a look and try to fix this. Thanks for telling me!

Snowstar38:
@scribblecrumb Hey, I apologize for taking so long to look into this - life stuff. But I just tried to do this and leader thoughts work fine for me? I cloned the dev branch and merged in only these file changes, but my leader death thoughts appear properly for me under both circumstances 🤔

scribblecrumb:
Hmmm very odd, this is the error I get when my leader dies.

image

It looks like your branch is pretty far behind dev rn too, you may want to sync it.  Perhaps that's making a difference in what I'm experiencing?

image

Snowstar38:
@scribblecrumb 🤦‍♀️ I have no idea how I managed to not notice I hadn't updated the branch, haha!

In any case, once I updated it, I ran into the same issues you'd mentioned. I think they should be fixed now. Thanks for being patient with me about this 😅

scripts/cat/cats.py
```diff
@@ -558,12 +563,10 @@ def die(self, body: bool = True):
         else:
             self.dead = True
             game.just_died.append(self.ID)
-            self.thought = (
-                "Is surprised to find themself walking the stars of Silverpelt"
-            )
+            death_thought = Thoughts.new_death_thought(self, darkforest, isoutside)
+            final_thought = event_text_adjust(self, death_thought, self)
```
scribblecrumb: make sure that you specify `main_cat=self` in the event_text_adjust arguments, otherwise it'll think `self` is the patrol leader, not the main cat.

scripts/cat/thoughts.py
```diff
@@ -330,3 +330,57 @@ def get_chosen_thought(main_cat, other_cat, game_mode, biome, season, camp):
...
         return chosen_thought
+    
+    def create_death_thoughts(self, inter_list) -> list:
+        #helper function for death thoughts
+        created_list = []
+        for inter in inter_list:
+            created_list.append(inter)
+        return created_list
+    
+    def leader_death_thought(self, lives_left, darkforest):
```
scribblecrumb, line 341: lets add some docstrings to these new functions!  Make sure to include the params with `:param name_of_param: explanation of param`

scripts/utility.py
```diff
@@ -1894,7 +1894,10 @@ def event_text_adjust(
 
     # main_cat
     if "m_c" in text:
-        replace_dict["m_c"] = (str(main_cat.name), choice(main_cat.pronouns))
+        if main_cat:
+            replace_dict["m_c"] = (str(main_cat.name), choice(main_cat.pronouns))
+        else:
+            replace_dict["m_c"] = (str(Cat.name), choice(Cat.pronouns))
```
scribblecrumb: I don't believe this quite works as intended (nor do I think it's a necessary change, you just need to pass the cat as `main_cat`) `Cat` is meant to be the Cat class, not a cat object.  So it looks like you're calling for the Cat class `.name` and `.pronouns`, not a cat object's `.name` and `.pronouns`

scribblecrumb (changes requested):

Snowstar38:
@scribblecrumb I have followed your suggestions! The edit to utility.py was just because I didn't know why it kept crashing aha; you figured out it was the main_cat=self part so now it's unnecessary and I removed it. :)

scribblecrumb (approved):

PR #2304, opened 2024-05-02
Mentor filters

Snowstar38:
Adds filters for the choose mentor screen:
- cats with no current apprentice
- cats with no former apprentices

This is very useful in Clans with large numbers of cats, to save time having to manually click on and check each cat for if they do or don't have an apprentice.

image


Lixxis:
Good idea! Love it. 
@scribblecrumb can you look over it, maybe the UI can be improved a bit, so the information what the buttons do is already visible when not hovering it?

scribblecrumb:
> Good idea! Love it. @scribblecrumb can you look over it, maybe the UI can be improved a bit, so the information what the buttons do is already visible when not hovering it?

Sure, I'll have a look and see what i can come up with.

scribblecrumb:
Can we get the UI styled to match the mate-choosing screen?  It's already got filters, so I think it'd be good to match it. 

image

Snowstar38:
I hadn't even thought of that, Scribble! Good idea. Will do this later tonight hopefully. :)

Snowstar38:
Sorry that took me so long! This was a headache for someone who's never done UI before, but I've changed it!

https://github.com/ClanGenOfficial/clangen/assets/104292791/a4051ebd-d5d0-471b-9400-40ae43513ccc

scribblecrumb (approved):

PR #2309, opened 2024-05-03
Matrix c/i Rewrite

lillithrosepup:
Rewrites the build.yml to be two jobs.

- Configure release
- Build

Configure release just sets variables

Build builds it for all platforms. The only platform specific logic is on macos running appdmg, and on windows building the self updater. All important things are defined in the matrix to keep the steps as simple as possible

[Here is a working test of it (granted, it doesnt have the update api but it should work) on my repo](https://github.com/ImLvna/clangen/actions/runs/8942525575)

Requesting review from @archanyhm as always

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,6 @@ env:
 
 
 jobs:
-
   configure-release:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


lillithrosepup:
oh yeah forgot to add, this also adds /utils/util.py and /utils/version.py

util.py contains a function to run a command and return stdout, stderr, and the exit code
version.py is a tool that either takes arguments and writes that to version.py, or if no arguments are specified it gets them from git cli

archanyhm:
I’ll have a look at this later today. Thanks for the contribution!

archanyhm (approved):

PR #2312, opened 2024-05-03
Fix attempt for issues #2290 and #2292

### LINKED ISSUE #2290: [BUG] Leader keeps "losing lives to weak leg" but doesn't actually lose them
**Type**: Code

**Describe the bug**
I have a Clan leader with the weak leg permanent condition. After the last update, every moonskip I get the message she "lost a live to weak leg". It gets recorded on her history text, but she doesn't actually lose any lives.

**Grade**: Regular/Important

**Reproduce**
Get a leader with the weak leg permanent condition, then skip moons and check the events and their death history text.

**Commit # or Game Version Number**: f586b2462

**Screenshots**
image
image

### LINKED ISSUE #2292: [BUG] Warrior dies while kitting but text says they lost a life
**Type**: Code

**Describe the bug**
A cat who's a warrior dies while giving birth, and the event text says she "lost one of her nine lives".

**Grade**: Non-vital

**Reproduce**
Get non-leader cats to get pregnant and skip moons until in one birth this text appears.

**Commit # or Game Version Number**: f586b2462

**Screenshots**
image
image

### LINKED ISSUE #2300: [BUG] Leader lost life to a weak leg, not actually losing life, happens multiple moons in a row
**Type**: (only select one)
Code

**Describe the bug**
I get a continuous moon event saying my leader has lost a life to a weak leg. The life counter doesn't go down and the event doesn't clear the next moon.

**Grade**: (only select one)
Important

**Reproduce**
Make a leader, give them the weak leg condition, see if the error occurs.

**Patrol ID + Outcome Print (if applicable)**:
N/A
0.9.0 (Latest Dev released yesterday, May 1st 2024)

**Additional context**

**Screenshots**
Screenshot 2024-05-02 124659
Screenshot 2024-05-02 125308
Screenshot 2024-05-02 125320
Screenshot 2024-05-02 125333
Screenshot 2024-05-02 125349

KiraDragoness:


scripts/events_module/condition_events.py
```diff
@@ -678,11 +678,11 @@ def handle_already_disabled(cat):
             status = cat.moon_skip_permanent_condition(condition)
 
             # if cat is dead, break
-            if cat.dead or cat.status == 'leader':
+            if cat.dead:
```
Lixxis: If the leader has more than one live left, this if statement will not pas.
Maybe you can add a variable to check how many lives are left before the function call "moon_skip_permanent_condition" and check if the number decreases?

Lixxis (changes requested):

scripts/events_module/condition_events.py
KiraDragoness: Did you mean something like what I changed now? Sorry if I got it wrong, I'm basically learning from scratch through deduction/trial and error!

Lixxis: Not quite. Sorry if I didn't explain it good enough, here is another try:

The "boolean" cat.dead (boolean contains information "true" or "false") is only set to true, when the cat is really dead. The thing with the leader is, that their life is tracked by `game.clan.leader_lives` and the `.dead` value will be set to true if the leader does not have any lives left. 

To see if the permanent conditions result in one death of the leader, we have to compare the live amount before and after the `cat.moon_skip_permanent_condition` call. If there is a difference, we have to be in that part of code. 

The change in line 685 is correct, but you have to adapt line 681.
My idea: add in line 677 (right after the comment "checking if the ...") 
`previous_lives = game.clan.leader_lives` and then adapt line 681 to this: `if cat.dead or (previous_lives != game.clan.leader_lives and cat.status == "leader"):`

Do you understand what I mean? /genq

scripts/events_module/condition_events.py
```diff
@@ -675,14 +675,15 @@ def handle_already_disabled(cat):
...
             status = cat.moon_skip_permanent_condition(condition)
 
             # if cat is dead, break
-            if cat.dead or cat.status == 'leader':
+            if cat.dead or game.clan.leader_lives < prev_lives:
                 triggered = True
                 event_types.append("birth_death")
                 event = f"{cat.name} died from complications caused by {condition}."
-                if cat.status == "leader" and not cat.dead:
+                if cat.status == "leader" and game.clan.leader_lives > 1:
```
Lixxis:
```suggestion
                if cat.status == "leader" and game.clan.leader_lives >= 1:
```

Lixxis:
@KiraDragoness 
I think my last suggestion got lost, so I ping. 
Only a small fix to change the line 686 to "game.clan.leader_lives >= 1:" so the case of one life is left is also covered \info

KiraDragoness:
Sorry, my bad! I somehow didn't get a notification for that. It's changed now!

scripts/events_module/condition_events.py
```diff
@@ -675,14 +675,15 @@ def handle_already_disabled(cat):
...
 
             # if cat is dead, break
-            if cat.dead or cat.status == 'leader':
+            if cat.dead or game.clan.leader_lives < prev_lives:
                 triggered = True
                 event_types.append("birth_death")
                 event = f"{cat.name} died from complications caused by {condition}."
-                if cat.status == "leader" and not cat.dead:
+                if cat.status == "leader" and game.clan.leader_lives >= 1:
                     event = f"{cat.name} lost a live to {condition}."
```
Thlumyn: This isn't something new for the PR, but if you want to change it to "lost a life to" instead of "lost a live to" while you have this file open
I'd really appreciate it

KiraDragoness: Oh yeah I was aware of the typo but totally forgot to fix it too, done!

Lixxis (approved):

PR #2327, opened 2024-05-08
fix: adjust path to self_updater binary after PyInstaller bump

archanyhm:


archanyhm:
Since PyInstaller now stores all the bundled files in the `_internal` subfolder, the self_updating capabilities are a bit broken at the moment.

Thlumyn (approved):

Issue #2344, opened 2024-05-15
[BUG] Lost Cats Returning with Faded Kits


jaysoleson:
**Type**: Code

**Describe the bug**
This is a bug found by kitty051336 in the Lifegen server that occurs when lost/outside cats with faded kits attempt to return to the Clan. The code that brings their kits along with them gives a keyerror for the faded kit's ID and crashes the game.

**Grade**:  Non-vital (? WILL cause a crash, but it's rare to occur)

**Reproduce**
Steps to reproduce the behavior (with default fading time):
1. Have a cat under 15-ish moons with a 1-2 moon old kit in the Clan
2. Have this parent cat become lost, and their kit dies shortly after.
3. Have the parent stay alive outside until the kit fully fades.
4. Now, when the cat attempts to return to the Clan, the faded kit will cause a crash.

**Commit # or Game Version Number**: 
This was actually found in Lifegen, so I'm not actually sure (we double checked it wasn't our issue lol). The last Clangen commit we have merge looks to be 2fc4d52 i think ?

**Additional context**
Zabe found that the fix is just a simple addition to add_to_clan in cats.py! it checks for child.exiled and child.outside, but not if theyre dead.

**Screenshots**
![fadedkit_crash](https://github.com/ClanGenOfficial/clangen/assets/144476842/46b5bde4-f645-404b-8e4c-4b1b94a1e261)

ZtheCorgi:
to be clear the parent cat would likely need to become lost while pregnant, or them and their kits would need to become lost in the same event (which is only possible with mass extinction turned on at the moment)

ReenieKitty:
> to be clear the parent cat would likely need to become lost while pregnant, or them and their kits would need to become lost in the same event (which is only possible with mass extinction turned on at the moment)

I'm the one who reported this, I created this account for any bugs in the future and just happened to see this right afterwards.

For the sake of clarity, the returning cat, ID 311, was male. He joined the clan with two newborn kits at a young age, then was lost in the mass extinction where people come and take a bunch of cats. As far as I know, neither of his kits were lost at that time. The kit in question, ID 313, died at 5 moons old, and faded. 311 was lost for over 200 moons before dying outside the clan at 240 moons old. 

[...]

Issue #2340, opened 2024-05-14
[BUG] Wrong cat 'lost one of their nine lives' giving birth.


JkHowling:
**Type**: Code?

**Describe the bug**
Sunstar's mate, Stormyfox (the deputy), died while giving birth, however the event incorrectly stated she lost 'one of her nine lives' instead. 

**Grade**: non-vital

**Commit # or Game Version Number**: 72741e69

**Additional context**

**Screenshots**
![image](https://github.com/ClanGenOfficial/clangen/assets/26156840/ffd87622-1305-49ca-b41b-c557176539d9)
![image](https://github.com/ClanGenOfficial/clangen/assets/26156840/ba762844-81a5-4319-afaf-bf2c84a83c24)

archanyhm:
Should be fixed with #2348

Issue #2326, opened 2024-05-08
[BUG] Cat speaks with his dead mates to start dating a new cat


KiraDragoness:
**Type**: Code

**Describe the bug**
A cat has currently 2 dead mates (from who he hasn't moved on yet) and one living mate. When he wants to start dating a new cat, instead of talking only with his current living mate, he talks to the 2 dead ones as well.

**Grade**: Non-vital

**Reproduce**
My guess is this only happens when a cat has both one or more dead mates and a living mate, and they want to start dating a new one. So to reproduce it you would need to:
1. Kill a cat's mate.
2. Get that cat to date other cat.
3. Get that cat to want to date a third cat, before they move on from the dead one.

**Commit # or Game Version Number**: f586b2462

**Screenshots**
![image](https://github.com/ClanGenOfficial/clangen/assets/121829433/537a41a1-e2d1-4ad4-9abd-f49f5528f8d2)
![image](https://github.com/ClanGenOfficial/clangen/assets/121829433/a3469dc0-7af9-438e-83f1-38dba98609ee)

archanyhm:
Should be fixed with #2348

Issue #2325, opened 2024-05-08
[BUG] The wrong cat gets injured on a training patrol


untitledids:
**Type**
Code

**Describe the bug**
A random cat (r_c) gets stuck on a training patrol in the beach biome. In one outcome after proceeding, triggered by a different cat having a certain stat trait (bloodthirsty, cold, fierce, shameless, strict, troublesome, or vengeful), the stat cat (s_c) gets a blunt force injury and cheek scar. It also says that the stat cat made it out and blames the random cat for pushing them. The random cat was the one stuck in the first place, so it should be them blaming the stat cat for pushing them, and the random cat should also be the one with the injuries.

The patrol's code:
Trailing behind the patrol, the dune {PRONOUN/r_c/subject}{VERB/r_c/'re/'s} crossing shifts beneath r_c's paws. {PRONOUN/r_c/subject/CAP} {VERB/r_c/look/looks} around wildly for safe footing - just as the sand caves in beneath {PRONOUN/r_c/object} and the world goes dark.
Even though s_c made it out, {PRONOUN/s_c/subject} {VERB/s_c/swear/swears} someone pushed {PRONOUN/s_c/object}, and {VERB/s_c/blame/blames} r_c. r_c denies it and says they should get s_c back to camp to get {PRONOUN/s_c/poss} injury looked at.

Plain text example:
Trailing behind the patrol, the dune they're crossing shifts beneath Boulderpaw's paws. They look around wildly for safe footing - just as the sand caves in beneath them and the world goes dark.
Even though Snakepaw made it out, they swear someone pushed them, and blames Boulderpaw. Boulderpaw denies it and says they should get Snakepaw back to camp to get their injury looked at.

**Grade**
Important

**Reproduce**
1. Send a group of 3 - 6 cats on a training patrol in the beach biome.
2. Bring at least one cat that has one of the following traits: bloodthirsty, cold, fierce, shameless, strict, troublesome, or vengeful.
3. Get this specific training patrol.
4. Proceed on the patrol and hope the specific outcome triggers.

**Patrol ID + Outcome Print (if applicable)**
PATROL ID: bch_train_stuck1 | SUCCESS: False

**Commit # or Game Version Number**
v0.11.2

**Additional context**
The other biomes' code for this specific outcome on their stuck training patrols are correct.

archanyhm:
Should be fixed with #2348

PR #2348, opened 2024-05-16
Bugfixes: various

### LINKED ISSUE #2345: [BUG] Cat with both deaf and partial hearing loss permanent conditions
**Type**:
- Code - general bugs related to how the game runs and responds to you. 

I genuinely don't know but this seemed like the closest option given an oversight in the code is a possible cause.

**Describe the bug**
Hopefully the screenshot is pretty self explanatory, but I don't think a deaf cat should also be able to also have partial hearing loss, same for a blind cat having failing eyesight, although I haven't seen that one (I just assume it's possible considering it's a very similar scenario). Since partial hearing loss can turn into deafness, by that logic, I would think one cat shouldn't be able to be both. 

**Grade**: 

Took me a very long time to even notice so there's really no impact on anything except maybe lore. 

**Reproduce**
1. Acquire a cat with the deaf permanent condition
2. Have that cat also gain partial hearing loss (example cat had a torn ear from an enemy warrior despite always being a medicine cat and never a warrior apprentice, it was a random moonskip injury). 

This was found in Lifegen but I was told it was likely a Clangen problem. Most recent version of Lifegen, not sure what version of clangen is running underneath. 

**Additional context**
I brought this up in the Lifegen server a few days ago as I wasn't sure if Lifegen had changed permanent conditions and was told this was likely a vanilla Clangen thing, but now that I've created an account here I figured I would make a report since I can't find one if one was opened for me. 

I did some testing by setting the probability of the partial hearing loss progressing to a second instance of deafness from 80 to 2 (after inquiring is a lower number means higher or lower chances and was told 80 meant 1/80, 2 would be 1/2), and nothing happened after skipping 20 moons. I then changed it to 1, just to see what would happen. After that failed, I edited her conditions file to replace the partial hearing loss with deaf, and it wiped her permanent conditions in game. File was intact, but no permanent conditions displayed in the game itself. That was easily fixed by replacing the second deaf with partial hearing loss again. I don't think it's actually possible to get a double deaf cat, but I still don't think the overlap should be possible. 

**Screenshots**
Screenshot (467)

sedgestripe:
Fixes #2345, #2344, #2340, #2326, #2325, #2324, #2321

### DIFF
--- a/scripts/events_module/relationship/romantic_events.py
+++ b/scripts/events_module/relationship/romantic_events.py
@@ -672,7 +672,7 @@ class Romantic_Events():
         """Prepares the relationship event string for display"""
         # replace mates with their names
         if "[m_c_mates]" in mate_string:
-            mate_names = [str(cat_from.fetch_cat(mate_id).name) for mate_id in cat_from.mate if not cat_from.fetch_cat(mate_id).dead and not cat_from.fetch_cat(mate_id).outside]
+            mate_names = [str(cat.name) for mate_id in cat_from.mate if (cat := cat_from.fetch_cat(mate_id)) is not None and not cat.dead and not cat.outside]
             mate_name_string = mate_names[0]
             if len(mate_names) == 2:
                 mate_name_string = mate_names[0] + " and " + mate_names[1]
@@ -681,7 +681,7 @@ class Romantic_Events():
             mate_string = mate_string.replace("[m_c_mates]", mate_name_string)
 
         if "[r_c_mates]" in mate_string:
-            mate_names = [str(cat_to.fetch_cat(mate_id).name) for mate_id in cat_to.mate if not cat_to.fetch_cat(mate_id).dead and not cat_to.fetch_cat(mate_id).outside]
+            mate_names = [str(cat.name) for mate_id in cat_from.mate if (cat := cat_from.fetch_cat(mate_id)) is not None and not cat.dead and not cat.outside]
             mate_name_string = mate_names[0]
             if len(mate_names) == 2:
                 mate_name_string = mate_names[0] + " and " + mate_names[1]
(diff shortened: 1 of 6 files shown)


scripts/events_module/relationship/romantic_events.py
```diff
@@ -672,7 +672,7 @@ def prepare_relationship_string(mate_string, cat_from, cat_to):
         """Prepares the relationship event string for display"""
         # replace mates with their names
         if "[m_c_mates]" in mate_string:
-            mate_names = [str(cat_from.fetch_cat(mate_id).name) for mate_id in cat_from.mate]
+            mate_names = [str(cat_from.fetch_cat(mate_id).name) for mate_id in cat_from.mate if not cat_from.fetch_cat(mate_id).dead and not cat_from.fetch_cat(mate_id).outside]
```
Lixxis: In general ok, but the thing is, that .fetch_cat might return an None if the id is not found, which would then lead to an error because None doesn't have variables like .name or .dead or .outside.
Not sure if you can change it in this PR to prevent such errors.

scripts/events_module/relationship/romantic_events.py
```diff
@@ -681,7 +681,7 @@ def prepare_relationship_string(mate_string, cat_from, cat_to):
             mate_string = mate_string.replace("[m_c_mates]", mate_name_string)
 
         if "[r_c_mates]" in mate_string:
-            mate_names = [str(cat_to.fetch_cat(mate_id).name) for mate_id in cat_to.mate]
+            mate_names = [str(cat_to.fetch_cat(mate_id).name) for mate_id in cat_to.mate if not cat_to.fetch_cat(mate_id).dead and not cat_to.fetch_cat(mate_id).outside]
```
Lixxis: In general ok, but the thing is, that .fetch_cat might return an None if the id is not found, which would then lead to an error because None doesn't have variables like .name or .dead or .outside.
Not sure if you can change it in this PR to prevent such errors.

Lixxis (changes requested):
Just a small thing to discuss and either fix or leave it like it is now

Lixxis (approved):

Issue #2294, opened 2024-05-01
[CODE] Modify new_cat code to specify romance between outsiders and clan cats


Tiritirimatangi:
Currently the new_cat code (documented here: https://github.com/ClanGenOfficial/clangen/wiki/%5BWriting%5D-%E2%80%90-Patrols#new_cat-listliststr) has the ability to specify mates - so for example you can generate a pair of mated new_cats who join the clan together, or you can have a new_cat join as the mate of a cat on patrol

I would like to modify this code and expand it to include checking if the cat is applicable to have romance with the new_cat. This would be "romance:{index},{index}" and "age:romance". This prevents future bugs when specifying relationship romantic gains with new cats, such as having a 80 moon old new cat flirt with a 13 moon old warrior.

sedgestripe:
![image](https://github.com/ClanGenOfficial/clangen/assets/110428732/e4c743ef-e5b5-4d18-9798-cb0f95610040)
![image](https://github.com/ClanGenOfficial/clangen/assets/110428732/15a7163d-e8a6-45f1-a36c-4712c0798701)
![image](https://github.com/ClanGenOfficial/clangen/assets/110428732/c0d76110-e5b4-4cbb-9487-ceb395812b51)
![image](https://github.com/ClanGenOfficial/clangen/assets/110428732/810280e5-9e87-41d6-90aa-1d00fe78abca)

I have added the "romance:{index},{index}" and "age:romance" tags to patrol. I believe this should be the exact same logic used with mate:{index},{index} and age:mate, except it should not set the two cats as mates? Let me know if this is what you wanted.

Tiritirimatangi:
[...]

PR #2349, opened 2024-05-16
Enhancement: create romance tags

sedgestripe:
Addresses #2294

image
image
image
image

I have added the "romance:{index},{index}" and "age:romance" tags to patrol. I believe this should be the exact same logic used with mate:{index},{index} and age:mate, except it should not set the two cats as mates? Let me know if this is what you wanted.

### DIFF
--- a/scripts/patrol/patrol_outcome.py
+++ b/scripts/patrol/patrol_outcome.py
@@ -877,6 +877,33 @@ class PatrolOutcome():
                 
                 give_mates.extend(patrol.new_cats[index])
         
+        give_romance = []
+        for tag in attribute_list:
+            match = re.match(r"romance:([_,0-9a-zA-Z]+)", tag)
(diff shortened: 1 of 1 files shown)


Thlumyn:
Does this actually give them some relationship/romance values towards each other? Or is it just checking that the other cat is romance-able?

sedgestripe:
This checks the other cat is romanceable since relationship values are changed in a different section of the patrol.

> Does this actually give them some relationship/romance values towards each other? Or is it just checking that the other cat is romance-able?

Thlumyn (approved):

Tiritirimatangi (approved):
Thankyouthankyouthankyouthankyouthankyouthankyouthankyouthankyouthankyou I'm going to get up to SO much nonsense with this, let the drama begin!

PR #2378, opened 2024-05-26
Feature: Leader Den

scribblecrumb:
## About The Pull Request

- Adds Clan symbols.  There is a goal of creating a symbol to correspond with each normal_prefix currently in the game.  As of time of writing, about 260 of 600 are finished.  Clan symbols are assigned first by prefix, then, if no corresponding symbol is available for the prefix, randomly. 
- Adds ability to access the Leader's Den through Camp page and through Den dropdown list
- Leader's Den screen allows for the following:
  - Viewing information (relation, temperament, name, symbol) of Other Clans
  - Choosing an interaction to have with one Other Clan.  One interaction allowed per moon.  Interactions take affect upon timeskip and will influence relations.  Success chance is determined by "distance" between Other Clan temper and Player Clan temper.
  - If the leader isn't able to work, a "helper_cat" steps in to fill their role.  This cat is displayed in the Leader's Den and event text changes appropriately.  If there is no helper cat available, then no interactions can occur with Other Clans. 
  - View information (reputation, name, status, trait, skill) of Outsiders
  - Choose an interaction to have with one Outsider.  One interaction allowed per moon.  Interactions take affect on timeskip and, if successful, either kill, drive away (remove from cat list and possible future interactions with Clan), or invite in/search for.  Success chance is determined by reputation with Outsiders.
- Changes made for Leader's Den:
  - New resource dicts have been added accordingly: `leader_den/{success or fail}/{other_clan or outsider}.json` to hold the possible events for these interactions.  `generate_events.py` has been expanded to handle generating these events.
  - New bool attribute `.driven_out` added to Cat objects to track which Outsiders have been successfully driven out.  These cats should still age and die as normal, they just cannot be seen by the player and cannot interact with the Clan.
  - Some patrols previously did not generate new outsider cats despite referencing an outsider cat.  Some patrols have been edited accordingly to now generate those cats.  This should hopefully allow the Leaders Den to feel as though it has more reason to impact the Outsider population.
  - `handle_lead_den_event()` added to `events.py` to initiate saved lead den events.  Chosen lead den interactions are held within clan_settings until timeskip, then erased after taking effect.
  - `handle_lost_cats_return()` in `events.py` adjusted to be capable of handling a predetermined list of cats.  This allows us to bring back lost cats via the Leader Den and still have them gain the proper ceremonies and statuses.
  - Added `temperament_dict` to Clan class for easier access.  Added `interaction_dict` to OtherClan class for easy access.
- Adds ability to choose a Clan symbol for the Player Clan
  - In saved Clans that have no symbol, a Clan symbol will be chosen upon loading.  
  - Upon clan creation, the player can choose a symbol for their clan.  The possible symbol list can be filtered by the player to only display symbols with certain attributes.  This will hopefully make it easier to sift through the already large list.  A symbol is also recommended to the player if a symbol exists for their chosen Clan prefix.
  - The clan_symbols.json holds information on both number of variants and filter tags for each symbol.
  - Symbol is displayed on Events page

Adds a new den with a meaningful mechanic.  Players previously had no way to track relationships with other Clans and outsiders without viewing raw save files.  Likewise, they could not view their own Clan's temperament or the temperament of other Clans.  The Leader's den allows players to see that information in game as well as further influence those relationships in new ways.  

The addition of Clan symbols is both valuable for increasing the identity and individuality of other Clans as well as providing another way for players to further customize their own Clan's identity (not to mention, opens a whole new way to mod the game).

## Proof of Testing

This PR has already been fairly thoroughly tested by the Beta Testers: [Leader Den Beta Testing Thread](https://discord.com/channels/1003759225522110524/1242623089368371290) 

image

image

image

image

image

## Changelog/Credits

- Added the Leader's Den: Includes ability to view Clan temperaments and relations, influence Clan relations, view Outsider cats and reputation, and influence Outsider population in various ways.
- Added Clan Symbols (223 currently available in game)


scripts/screens/LeaderDenScreen.py
```diff
@@ -0,0 +1,894 @@
...
+            starting_height=3,
+            manager=MANAGER,
+        )
+        y_pos += 70
+        self.focus_button["drive_off"] = UIImageButton(
+            scale(pygame.Rect((0, y_pos), (242, 60))),
+            "",
+            object_id="#outsider_drive",
+            tool_tip_text="This cat will be driven out of the area if found (they will no longer be accessible in "
+                          "game.)",
```
Thlumyn:
```suggestion
                          "game).",
```

Thlumyn (changes requested):
I'm going to add some screenshots, I noticed some things when testing it out

Thlumyn:
dead_cat_helper
Dead cat shows up as a helper? Not sure if this is intended behavior (as the rest of the clan was dead in this situation) or not, either way thought it worth bringing up.

deadleader
Dead leader was showing up in the den after killing off leader, despite having other living cats that could probably do a better job than a ghost.

den_image_placement
With a lot of the den images, the background image cuts real close (or even through, as seen in the cavern background in the previous screenshot) to the text just under it; could we move that text down just a tad? Else, I think we should add some background there so that the text isn't getting interrupted and made hard to read

invited_loner_wrong_backstory
Loner invited into the clan; the backstory doesn't quite make sense here. Maybe might be better to add a specific backstory for cats being invited in like this?

outsider_list_overrun
Cat images in the outsider list overflow to the right

parenthesis_ic_issue
The way this is phrased just seems a bit out of place with the rest of the information in the Events tab; this seems more like debug game information, that breaks the immersion of the game a bit. Maybe something like removing the success or fail part, and just leaving clan relations improved (as long as the player remembers whether they were insulting or complimenting, that should be enough to tell them whether it was success or fail) could still give that information while seeming like an event info that could be in character for the clan.

scribblecrumb:
> dead_cat_helper Dead cat shows up as a helper? Not sure if this is intended behavior (as the rest of the clan was dead in this situation) or not, either way thought it worth bringing up.
> 
> deadleader Dead leader was showing up in the den after killing off leader, despite having other living cats that could probably do a better job than a ghost.
> 
Fixed!  I hadn't even considered what it should do if the whole clan was dead haha, so that's now accounted for.  

I'll have a look at the other things you mentioned later (currently on my lunch break from work lol).  Though with regards to the backstories, I'm honestly a bit unsure how to handle those.  Assigning a special backstory means losing a lot of uniqueness that comes from the variety we currently have (unless we create special variations of all of them, which I'm not keen on doing).  We could *maybe* add something extra to the end about the Clan having sought them out specifically?  But with how backstory history is currently set up, that will be a headache.

scribblecrumb:
> den_image_placement With a lot of the den images, the background image cuts real close (or even through, as seen in the cavern background in the previous screenshot) to the text just under it; could we move that text down just a tad? Else, I think we should add some background there so that the text isn't getting interrupted and made hard to read
> 
> outsider_list_overrun Cat images in the outsider list overflow to the right
> 
> parenthesis_ic_issue The way this is phrased just seems a bit out of place with the rest of the information in the Events tab; this seems more like debug game information, that breaks the immersion of the game a bit. Maybe something like removing the success or fail part, and just leaving clan relations improved (as long as the player remembers whether they were insulting or complimenting, that should be enough to tell them whether it was success or fail) could still give that information while seeming like an event info that could be in character for the clan.

Fixed!

Thlumyn:
> > dead_cat_helper Dead cat shows up as a helper? Not sure if this is intended behavior (as the rest of the clan was dead in this situation) or not, either way thought it worth bringing up.
> > deadleader Dead leader was showing up in the den after killing off leader, despite having other living cats that could probably do a better job than a ghost.
> 
> Fixed! I hadn't even considered what it should do if the whole clan was dead haha, so that's now accounted for.
> 
> I'll have a look at the other things you mentioned later (currently on my lunch break from work lol). Though with regards to the backstories, I'm honestly a bit unsure how to handle those. Assigning a special backstory means losing a lot of uniqueness that comes from the variety we currently have (unless we create special variations of all of them, which I'm not keen on doing). We could _maybe_ add something extra to the end about the Clan having sought them out specifically? But with how backstory history is currently set up, that will be a headache.

Just reading through the current backstories - I think the only ones that seem like a major disconnect to me are the guided ones; could we do new backstory_categories for the invited in cats, to filter out ones like those, instead of writing whole new backstories for them?

scribblecrumb:
> Just reading through the current backstories - I think the only ones that seem like a major disconnect to me are the guided ones; could we do new backstory_categories for the invited in cats, to filter out ones like those, instead of writing whole new backstories for them?

Alright, I actually ended up just adding a check for the guided backstories.  Now, if a cat being invited in has one they'll just be switched to the "outsider1" backstory (a very basic "cat is from outside the clan").  I decided to do it this way because just randomly choosing a new backstory from the entire range of backstories could lead to some very weird discrepancies.  Namely, a backstory not matching with the type of outsider the cat used to be (loner, kittypet, ect.)

Thlumyn:
> > Just reading through the current backstories - I think the only ones that seem like a major disconnect to me are the guided ones; could we do new backstory_categories for the invited in cats, to filter out ones like those, instead of writing whole new backstories for them?
> 
> Alright, I actually ended up just adding a check for the guided backstories. Now, if a cat being invited in has one they'll just be switched to the "outsider1" backstory (a very basic "cat is from outside the clan"). I decided to do it this way because just randomly choosing a new backstory from the entire range of backstories could lead to some very weird discrepancies. Namely, a backstory not matching with the type of outsider the cat used to be (loner, kittypet, ect.)

That sounds good 

One other bug I found, hopefully the last; if you exile a leader they still show up in the leader den

scribblecrumb:
> That sounds good
> 
> One other bug I found, hopefully the last; if you exile a leader they still show up in the leader den

Ah easy to fix at least, got that one taken care of!

Thlumyn (approved):

Issue #1881, opened 2023-11-08
Allow users to set “default” Clan settings


larkgz:
It can become somewhat cumbersome to have to set up preferred Clan settings every time a new Clan is started. It would be convenient if there was a way to save default settings that would apply whenever new Clans are created.

Rustykitty:
Should be pretty straightforward to implement.

PR #2399, opened 2024-06-04
Random Clan Quick Start

kainosaur:
## About The Pull Request

This pr is for a setting in clan creation to have your clan quickly generated for you. After picking your gamemode, the option for the player to generate a quick and random clan will show up.

Gets you into the game quicker if you so choose. 
## Linked Issues

This was inspired by issue #1881 , with the idea of trying to get into the game faster
## Proof of Testing

https://github.com/ClanGenOfficial/clangen/assets/152326772/aadd0deb-969d-47c6-a5fa-3d74816b4574
## Changelog/Credits

add random clans for startup


scribblecrumb:
I like the look of this!  On the UI side though, I'd prefer if we weren't interrupting the "normal" clan creation experience with the question/confirmation screen.  I think it'd be nicer to add a checkbox toggle to the game mode selection screen instead, perhaps right above the prev/next buttons.  You could label it as "Quick Start", have it toggled off automatically, and include a tooltip with a longer explanation (clan will be entirely randomized, season will be newleaf).  If you'd like an additional confirmation from the user, then you could have a popup appear upon clicking next (*if* the Quick Start is toggled on) asking if the user is certain of their choice.  But, I don't think that's particularly necessary in this case, since it's easy enough to just back out of the Clan and make a new one if you do accidentally have Quick Start toggled.

kainosaur:
Ok I'll try to make that! I'm not the best at ui though so it might not look very pretty

kainosaur:
Checkbox completed. Text is higher than it should be though.

https://github.com/ClanGenOfficial/clangen/assets/152326772/69f86328-fa42-44e3-b115-75cec3d39a4c

scribblecrumb (approved):

Issue #2264, opened 2024-04-27
[PROJECT] Moon Event Reformatting


scribblecrumb:
**Description**
This project intends to reformat the base moon events (death, injury, misc, new_cat) to bring them more in line with standards set by patrol format as well as increase consistency in terminology and tags used across events.

I've already reformatted the existing events via a script, but the reformatted events need to be reviewed by hand to ensure that the original intentions of each event is preserved.  Additionally, some old events were mis-tagged or incorrectly formatted, and thus will not have been reformatted correctly.  Moon events also need to be assigned unique event ids, as this was a completely new addition to the format.

The intention of this issue is to find more contributors willing to review the reformatted events.  

**Scope of Project**
Medium-sized Project - hopefully doesn't require an obscene amount of recoding on my end, and the number of moon events to review is relatively small compared to other event pools in the game (looking at you, patrols).

**What has already been completed?**

- [x] Plan new format
- [x] Convert moon events (death, injury, misc, new_cat) to new format
- [x] Move moon events from their `type/biome/rank.json` directory into a more condensed `type/biome.json` directory
- [x] Fix short event generation to work with the new directory
- [x] Review all converted moon events to ensure preservation of original event intentions and assign unique event IDs [**Need writer help here**]
       - [x] misc
       - [x] new_cat
       - [x] death
       - [x] injury
       - [x] freshkill (need to move these over to death and injury categories)
       - [x] herb destruction (these need to be completely rewritten into a short event format)
       - [x] mass extinction/mass death (these need to be moved into short event format)
- [x] Adjust filtering code to function with new format (created `filter_relationship_type()` in utility.py for use both in patrols and events)
- [x] Add `get_random_moon_cat()` to utility.py so that we don't have random_cat choosing code scattered in multiple places in events.py
- [x] Add `gather_cat_objects()` and `unpack_rel_block()` to utility.py to handle both patrol and moon event relationship changes
- [x] Add `get_warring_clan()` to utility.py to handle grabbing the enemy clan name
- [x] Add `create_new_cat_block()` to utility.py to handle creating new cats from patrol and moon event blocks
- [x] Add `get_alive_status_cats.()` to utility.py to gather respective cat IDs for filtering purposes
- [x] Add `get_leader_life_notice()` to make creation of notices regarding a leader's remaining lives easier to obtain
- [x] Adjust `event_text_adjust()` in utility.py to be capable of handling all utilized abbreviations within patrols and moon events
[...]

PR #2407, opened 2024-06-07
[CLEANUP] Age to Enum

j-gynn:
## About The Pull Request
I'm back at it again. This time, I'm doing some work in `cats.py`, lifting out some of the frequently-accessed things and making them not be strings. Part of #2436 

### Blockers
- [x] Blocked by #2264 (no code shared, I just wanna get that in first before this haha)
- [ ] Blocked by #2417 (I would very much like to use the `ExtendedStrEnum` class there in this one if possible, it has some helper functions that are extremely useful)

### Changes
- [x] Convert age to enum
- [x] Move moons-to-age calculation to enum & make sure it's all neato
- [x] Rename `age` on the create_new_cat function to `moons`, as it is incorrectly implying it's the age enum atm.

Happy code, happy devs. Happy devs, happy life. Better internal structures make it easier to make better stuff quicker.

## Proof of Testing

This will require HEAVY testing. I am basically smashing the codebase with a hammer. Along the way I have been making tests to ensure things do what they're supposed to, but I will need lots of help from beta testers to make sure things don't slip through the cracks.

Have written some supporting tests, but if there's any feedback and/or missing tests to integrate, let me know.

## Changelog/Credits

Internal cleanup & revisions


scripts/cat/age.py
```diff
@@ -0,0 +1,20 @@
+from enum import IntEnum
+
+
+class Age(IntEnum):
```
archanyhm: Why are you using an IntEnum instead of a StrEnum? Is this intentional?

j-gynn: It was intentional, yes - it has the very tiny win of letting you sort much easier lol. I had originally been using it for the logic stuff (in that you could go "if younger than adult" for example), but I've now put in a bunch of helpers to make the code easier to read for beginners (and also avoid having to import Age everywhere!)

scripts/cat/pelts.py
```diff
@@ -641,12 +642,12 @@ def init_sprite(self):
         self.cat_sprites['senior adult'] = self.cat_sprites['adult']
 
     def init_scars(self, age):
-        if age == "newborn":
+        if age.is_newborn():
             return
 
-        if age in ['kitten', 'adolescent']:
+        if age.is_underage():  # kitten or adolescent now as newborn already excluded
```
archanyhm: If the comment is required to aid in understanding the code, I'd prefer it if you renamed the method to `is_kitten_or_adolescent()` or simply `is_kitten() or is_adolescent()`

j-gynn: No, it's actually because the check contains newborn, kitten and adolescent. Since newborn is already excluded, this function is OK to use and still means what you think it means.

archanyhm: Ah, I skimmed through the PR and didn't see the previous check. With the given context, I don't think the comment is necessary as the code is pretty clear, no?

j-gynn: Good point, I'll remove it.

PR #2417, opened 2024-06-08
[CLEANUP] Status to Enum

j-gynn:
## About The Pull Request
I'm back at it again. This time, I'm doing some work in `cats.py`, lifting out some of the frequently-accessed things and making them not be strings. This flavour is `status`! Part of #2436 

### Blockers
- [x] Blocked by #2264 (no code shared, I just wanna get that in first before this haha)

### Changes
- [x] Convert status to enum
- [x] **(?)** Consider whether to set as property of Cat & migrate `status_change` to `@status.setter`

Happy code, happy devs. Happy devs, happy life. Better internal structures make it easier to make better stuff quicker.

## Proof of Testing

This will require HEAVY testing. I am basically smashing the codebase with a hammer. Along the way I have been making tests to ensure things do what they're supposed to, but I will need lots of help from beta testers to make sure things don't slip through the cracks.

Have written some supporting tests, but if there's any feedback and/or missing tests to integrate, let me know.

[Link to testing thread on Discord](https://discord.com/channels/1003759225522110524/1249433287773130874)

## Changelog/Credits

Internal cleanup & revisions


j-gynn:
Okaaaay... can someone take a look and help me figure out why freshkill_test is failing here? It passes on my local machine (no uncommited/unpushed changes either)

j-gynn:
Ready for review! At this point, I think I have to ignore freshkill_test, because it straight-up chooses when to work. I can rerun it until it passes if you want proof, but...

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+    ROGUE = "rogue"
+
+    def is_working_any(self):
+        """True if cat is a working role (warrior + app, medcat + app, mediator + app, deputy and leader)"""
+        return not (self.is_kit_any()
+                    or self.is_elder()) \
+            and self.is_clan_status()
+
+    def is_patrol_any(self):
+        """True if able to go on patrol (warrior + app, medcat + app, deputy and leader)"""
```
archanyhm: I think `can_patrol_any` would be preferrable here.

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+        return (self.is_warrior_any()
+                or self.is_medcat_any()
+                or self.is_deputy_or_leader())
+
+    def is_patrol_app(self):
+        """True if apprentice who can go on patrol (warrior app and medcat app)"""
```
archanyhm: See above

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def is_warrior_any(self):
+        """True if cat is warrior or apprentice."""
+        return self in [StatusEnum.WARRIOR, StatusEnum.WARRIORAPP]
+
+    def is_deputy_or_leader(self):
+        """True if cat is deputy or leader."""
+        return self in [StatusEnum.DEPUTY, StatusEnum.LEADER]
+
+    def is_warrior_medcat_or_mediator(self):  # TODO grr.
```
archanyhm: This method is quite wordy. My bigger issue is the unhelpful TODO comment. Please fix it with a more concise, descriptive TODO or remove it.

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+        """True if cat is warrior, deputy or leader"""
+        return self.is_deputy_or_leader() or self.is_warrior()
+
+    def is_outside_clan(self):
+        """True if cat is former Clancat, exiled, kittypet, loner or rogue"""
+        return self in [
+            StatusEnum.EXCLAN, StatusEnum.EXILED, StatusEnum.KITTYPET,
+            StatusEnum.LONER, StatusEnum.ROGUE]
+
+    def is_clan_status(self):
```
archanyhm: Just a suggestion, but mayhaps `is_inside_clan`? Or `is_in_clan`?

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
+from scripts.game_structure.extended_strenum import ExtendedStrEnum
+
+
+class StatusEnum(ExtendedStrEnum):
```
archanyhm: I know you specified a reason somewhere, but I personally find it a bit redundant to name variables, type definitions […] with their respective type. I know we're in Python and this is still a class rather than an actual enum, but calling a status enum "StatusEnum" is inconsistent with what we (should) usually do. I mean, think of the Cat class being called "CatClass"-

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Roles that a cat may hold. Check ExtendedStrEnum for more things!
```
archanyhm, line 6: Extremely nitpicky here, but is the "Check […]" part necessary?

scripts/cat/cats.py
```diff
@@ -792,8 +796,9 @@ def gone(self):
         mentors and apprentices."""
         self.outside = True
 
-        if self.status in ["leader", "warrior"]:
-            self.status_change("warrior")
+        if self.status.is_leader() or self.status.is_warrior():
+            # this seems like it should be is_deputy, but I'm doing a like-for-like replacement
```
archanyhm, line 820: Elaborate, please.

scripts/cat/cats.py
```diff
@@ -819,79 +824,67 @@ def add_to_clan(self) -> list:
...
-            if old_status == "leader":
+        elif new_status.is_warrior() or new_status.is_elder():
+            if old_status.is_leader():
                 if game.clan.leader:
                     if game.clan.leader.ID == self.ID:
                         game.clan.leader = None
                         game.clan.leader_predecessors += 1
 
-            if game.clan.deputy:
+            if game.clan and game.clan.deputy:  # don't remove the check for game.clan, this is needed for tests
```
archanyhm: I'd reword this to state _why_ it is needed.

scripts/cat/cats.py
```diff
@@ -819,79 +824,67 @@ def add_to_clan(self) -> list:
...
+                    and child.moons < 12
             ):
                 child.add_to_clan()
                 History.add_beginning(child)
                 ids.append(child_id)
 
         return ids
 
-    def status_change(self, new_status, resort=False):
+    def status_change(self, new_status: StatusEnum | str, resort=False):
```
archanyhm: We should not offer/encourage passing strings instead of their string enum counterpart.

scripts/cat/cats.py
```diff
@@ -819,79 +824,67 @@ def add_to_clan(self) -> list:
...
         """Changes the status of a cat. Additional functions are needed if you want to make a cat a leader or deputy.
-        new_status = The new status of a cat. Can be 'apprentice', 'medicine cat apprentice', 'warrior'
-                    'medicine cat', 'elder'.
+        new_status = The new status of a cat.
         resort = If sorting type is 'rank', and resort is True, it will resort the cat list. This should
                 only be true for non-timeskip status changes."""
+
+        # auto-converting string
+        if isinstance(new_status, str):
+            new_status = StatusEnum(new_status)
```
archanyhm: See above

scripts/cat/cats.py
```diff
@@ -819,79 +824,67 @@ def add_to_clan(self) -> list:
...
-                    game.clan.deputy_predecessors += 1
-
-        elif self.status == "medicine cat":
+        elif new_status.is_medcat():
             if game.clan is not None:
                 game.clan.new_medicine_cat(self)
 
-        elif self.status == "elder":
-            if old_status == "leader":
+        elif new_status.is_warrior() or new_status.is_elder():
```
archanyhm: Again very nitpicky, but why are we checking `new_status` instead of `self.status`? They might be equivalent in value, but since `self.status` is assigned the value of `new_status`, I'd much rather check that (like the old code did).
If you disagree, feel free to dispute this. I'm not hard sold on that.

scripts/cat/cats.py
```diff
@@ -902,7 +895,7 @@ def status_change(self, new_status, resort=False):
     def rank_change_traits_skill(self, mentor):
         """Updates trait and skill upon ceremony"""
 
-        if self.status in ["warrior", "medicine cat", "mediator"]:
+        if self.status.is_working_any():
```
archanyhm: Technically not "like for like" since the `is_working_any` check includes apprentices.

scripts/cat/cats.py
```diff
@@ -819,79 +824,67 @@ def add_to_clan(self) -> list:
...
+        # auto-converting string
+        if isinstance(new_status, str):
+            new_status = StatusEnum(new_status)
+
         old_status = self.status
         self.status = new_status
         self.name.status = new_status
 
+        if new_status.is_outside_clan():
+            return
```
archanyhm: Why return early? Doesn't this skip the updating of the `all_cats` dictionary?

scripts/cat/skills.py
```diff
@@ -366,17 +366,17 @@ def generate_new_catskills(status, moons, hidden_skill: HiddenSkillEnum = None):
...
             new_skill.primary = Skill.get_random_skill(points=0, interest_only=True)
-        elif status == "apprentice":
+        elif status.is_warrior_app():
             new_skill.primary = Skill.get_random_skill(point_tier=1, interest_only=True)
             if random.randint(1, 3) == 1:
                 new_skill.secondary = Skill.get_random_skill(
                     point_tier=1, interest_only=True, exclude=new_skill.primary.path
                 )
-        elif moons < 50:
+        elif moons < 50:  # Should these thresholds be in a config file?
```
archanyhm: Open a new issue or ask any of us Senior Developers in the Discord; comments like this sadly tend to remain in the code forever untouched.

archanyhm: (Or turn it into a TODO comment with a clear description of the task like you did below)

scripts/cat/skills.py
```diff
@@ -552,7 +552,7 @@ def progress_skill(self, the_cat):
                     elif self.primary:
                         self.primary.points += amount_effect
 
-            elif the_cat.moons > 120:
+            elif the_cat.moons > 120: #TODO: Check if this should be tied to a config option (start of elder, e.g.?)
```
archanyhm: Niiitpick time, space after the `#`

scripts/game_structure/extended_strenum.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ExtendedStrEnum(StrEnum):
+    @classmethod
+    def list(cls):
+        return list(map(lambda c: c.value, cls))
+
+    @classmethod
+    def index(cls, val: str, **kwargs):
+        return cls.list().index(val)
```
archanyhm: Documentation comments, please!

scripts/cat/enums/status.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def is_deputy_or_leader(self):
+        """True if cat is deputy or leader."""
+        return self in [StatusEnum.DEPUTY, StatusEnum.LEADER]
+
+    def is_warrior_medcat_or_mediator(self):  # TODO grr.
+        """True if cat is warrior, medcat or mediator"""
+        return self in [StatusEnum.WARRIOR, StatusEnum.MEDCAT, StatusEnum.MEDIATOR]
+
+    def is_normal_adult(self):
```
archanyhm: What does `normal` mean? I don't think that's the most fitting word.

scripts/screens/ClanScreen.py
```diff
@@ -343,36 +345,33 @@ def choose_cat_positions(self):
...
+            if Cat.all_cats[x].status.is_newborn() or game.config['fun']['all_cats_are_newborn']:
                 if game.config['fun']['all_cats_are_newborn'] or game.config['fun']['newborns_can_roam']:
                     # Free them
                     Cat.all_cats[x].placement = self.choose_nonoverlapping_positions(first_choices, all_dens,
                                                                                      [1, 100, 1, 1, 1, 100, 50])
                 else:
                     continue
 
-            if Cat.all_cats[x].status in ['apprentice', 'mediator apprentice']:
+            if Cat.all_cats[x].status.is_app_any():
```
archanyhm: `is_app_any` includes medicine cat apprentice. Not equivalent.

scripts/screens/WarriorDenScreen.py
```diff
@@ -71,7 +71,7 @@ def handle_event(self, event):
                         if "mediator" in description and self.save_button.is_enabled:
                             # only create the mediator list if needed to check
                             mediator_list = list(filter(
-                                lambda x: x.status == "mediator" and not x.dead and not x.outside, Cat.all_cats_list
+                                lambda x: x.status.is_mediator_any() and not x.dead and not x.outside, Cat.all_cats_list
```
archanyhm: why `is_mediator_any` instead of `is_mediator`?

archanyhm (changes requested):

scripts/cat/enums/status.py
j-gynn: I renamed them to this because I think I'd like to reorganise the Cat class to have an organisation object that holds status and a few other things. Enums can't hold other objects aside from Enums, so I was sort of future-proofing here to leave the `Status` namespace available. I can of course revert this and reconsider how I'd like to look at that organisation haha

archanyhm: Wouldn't this be remedied if you imported the enums module and referred explicitly to the Status enum via `enums.Status`?

scripts/cat/enums/status.py
j-gynn: Will do.

scripts/cat/enums/status.py
j-gynn: Removed TODO

scripts/cat/enums/status.py
j-gynn: My biggest concern with renaming the `is_patrol_blahblah` functions is that autofill will no longer prompt you as it's the only ones that won't now start with `is`. If that's a non-issue, I'll update it. Was just trying to ensure they wouldn't get missed as possibilities

scripts/cat/enums/status.py
j-gynn: Honestly no, but it's there as a prompt to newer developers that there are other things you can do with this class that aren't right there. I don't know if that's actually necessary or not in honesty haha, I just like how friendly the codebase is for beginners. I don't want my refactors to inhibit that nature, even if they're inherently making things more complicated.

archanyhm: Good enough for me. Keep it in then!

scripts/cat/cats.py
j-gynn: The docustring says that this is for handling removing cats from the clan (as in lost). It says it "handles removing them from special positions", and this section of the code is the only part that changes a status. Deputy never gets changed... and I think that changing one's status from warrior to warrior is a little superfluous. Hence my thought that this is supposed to be "is leader or deputy", as in the two special warrior roles.

scripts/cat/cats.py
j-gynn: It's hard to explain it succinctly. Would something like this be clear enough? `game.clan check is needed to let tests work as tests don't have an instance of game`

scripts/cat/enums/status.py
archanyhm: Yeah, I don't think that would be a big issue. ~Hopefully.~

scripts/cat/cats.py
j-gynn: I wholeheartedly agree - except that it means that a lot of classes need StatusEnum as a dependency now in order to assign one. If small class dependencies isn't an issue, I'll go through and change it (I think 99% of things have it there anyway)

scripts/cat/cats.py
archanyhm: Sounds good to me!

scripts/cat/cats.py
j-gynn: Honestly, I didn't notice this mistake. I _do_ think this is clearer for the reader, but it does introduce a small risk that somehow `self.status != new_status`.

scripts/cat/cats.py
archanyhm: I think the benefits of having clear, refactoring safe enums in place of magic strings everywhere possible reduces our headaches overall, so feel free to go ahead.

scripts/cat/enums/status.py
j-gynn: I suppose it would, yeah. More than happy to go and do that if preferred.

scripts/game_structure/extended_strenum.py
j-gynn: Whoops! I'm usually the one asking people to do these, must have slipped through the cracks. Will do that now.

scripts/cat/enums/status.py
j-gynn: I was told to use this phrasing because it matches the patrol phrasing. I'm not the biggest fan of it either - `is_deputy_leader_or_warrior` is "worse" but also clearer, so will swap to that

scripts/screens/ClanScreen.py
j-gynn: Good catch, thank you. This was before we had `can_patrol_app` so I'll replace it with that.

scripts/screens/WarriorDenScreen.py
j-gynn: Ah. The original seems to be an oversight: in `medicine cat` below, both medicine cats and their apprentices are counted in the enable/disable toggle, but here only mediators are. I think this might've just been a mistake, as in most cases we do want both (and annoyingly it autofills with "any" versus staying blank) but looking at it I'm not sure whether I should actually change it back. Thoughts?

scripts/cat/cats.py
j-gynn: Good spot, will fix.

scripts/screens/WarriorDenScreen.py
archanyhm: Optimally we want to _only_ refactor and fix things afterwards. Maybe revert this back to the "original logic" and fix in a separate PR with dedicated testing?

scripts/cat/cats.py
j-gynn: Good point, I have no idea why this is here, evicting it immediately.

scripts/cat/cats.py
archanyhm: If there are no early returns in this method, you can assign `new_status` to `self.status` at the end instead, giving us both the certainty that the two don't diverge + the clearer intent.

scripts/screens/WarriorDenScreen.py
j-gynn: Makes sense to me! I'll revert and add a note to self.

scripts/cat/skills.py
j-gynn: Fixed

scripts/cat/cats.py
j-gynn: Oh that's a good idea; will do

scripts/cat/skills.py
j-gynn: TODO-ified

j-gynn:
A few bugs have been found so temporarily putting this back to draft

scripts/cat/enums/status.py
j-gynn: Change made :)

Ryos00 (approved):

scripts/screens/ChooseMentorScreen.py
```diff
@@ -673,7 +671,7 @@ def get_valid_mentors(self):
 
             return valid_medcat_mentors
 
-        elif self.the_cat.status == "mediator apprentice":
+        elif self.the_cat.status.is_mediator_app()":
```
j-gynn:
```suggestion
        elif self.the_cat.status.is_mediator_app():
```

PR #2435, opened 2024-06-11
[CLEANUP] Freshkill Refactor

j-gynn:
## About The Pull Request

- [x] Overhaul FreshkillPile code for simplicity & robustness
- [x] Move Nutrition to a new file
- [x] Convert all properties of Nutrition to @property properties, max and current being managed, percentage and nutrition_text just being calculated
- [x] Refactor freshkill_test.py to never fail when things should be working
  - [x] GH does seem to throw errors on it now and then which I can't reproduce locally. I don't know why but I suspect it's to do with the order the cats enter the freshkill system in. Needs further investigation.

Happy devs, happy life. This is an especially crunchy part of the code that likes to break when unrelated features are modified, so it's good to try and stabilise it a bit more.

Note: I would like, in the not-too-distant future, to rework both Herbs and Freshkill to draw from a base Resources class.

## Linked Issues
Part of #2436 

## Proof of Testing

[Testing thread on Discord](https://discord.com/channels/1003759225522110524/1249885586647683123)

## Changelog/Credits

Freshkill refactored


scripts/clan_resources/freshkill.py
```diff
@@ -107,45 +58,29 @@ def remove_freshkill(self, amount, take_random: bool = False) -> None:
...
+        # handle kits and queens
+        clan_cats = self.handle_kits_and_queens(living_cats)
 
         # all normal status cats calculation
         needed_prey = sum(
             [
                 PREY_REQUIREMENT[cat.status]
                 for cat in living_cats
-                if cat.status not in ["newborn", "kitten", "exiled"] and not cat.outside
+                if cat.status not in ["kittypet", "rogue", "loner", "exiled", "former Clancat"] and not cat.outside
```
Ryos00: I feel like `not cat.outside` should cover all of these statuses? Ideally, of course

j-gynn: Unfortunately, "ideally" really is the operative word. This line is the number-one (or at least in the top 3) cause of test failures specifically because of that assumption. Because "outside" is a transient variable that technically depends on other variables to determine its state - but doesn't actually compute its state from them - it causes a lot of problems. The proper fix will be to tie `cat.outside` to its determining variables correctly (or just delete it as I'm really hoping to do). That said, that's outside the scope of this particular refactor.

j-gynn: Though there _is_ a bug in this line, so I'll go back in and fix that.

Ryos00: I see, I see. I was hoping that wasn't the case, but it's not surprising that it is, which is why I said "ideally" xD
Don't worry about that then.

j-gynn:
Closing as it's very stale and frankly, I kind of hate it UwU

PR #2465, opened 2024-06-15
Create tests for HandleShortEvents.

susieoneil:
- Add test structure
- Test reset()
- Test handle_accessories()
- Some error checks in handle_short_events.py

## About The Pull Request
- Created the file **test_handle_short_events.py** to test the functions of the HandleShortEvents class.  I made an empty TestCase for each of the functions, as a base to be filled in later.
- Created 1 test for reset() function. A little long, but tests all attributes changing.
- Created 7 tests for handle_accessories() function, 100% coverage.
- Made some modifications to **handle_short_events.py**, to:
  - Use getatt() to check some attributes exist. 
  - Simplify "tail_accessories in acc_list" check to remove print exceptions

- Note:  I mocked the event, but left in the Cat and Pelt imports for now. The other functions all use the imported **game**.  I did get a handle_herb_supply() test working with the game imported, but it's not finished.  Ideally tests would only test the class, without needing imports, but this could take a while.

- Increasing test coverage helps with finding potential bugs in the codebase.
- I asked on on Discord for something to test. This one was suggested by Scribble here: https://discord.com/channels/1003759225522110524/1005586496142704800/1250465137937092763

## Proof of Testing
image

### DIFF
--- a/scripts/events_module/handle_short_events.py
+++ b/scripts/events_module/handle_short_events.py
@@ -298,7 +298,7 @@ class HandleShortEvents:
         if extra_text and extra_text not in self.chosen_event.text:
             self.chosen_event.text = self.chosen_event.text + " " + extra_text
 
-    def handle_accessories(self):
+    def handle_accessories(self, pelts=Pelt):
         """
         handles giving accessories to the main_cat
         """
@@ -307,19 +307,19 @@ class HandleShortEvents:
         acc_list = []
         possible_accs = getattr(self.chosen_event, 'new_accessory', [])
         if "WILD" in possible_accs:
-            acc_list.extend(Pelt.wild_accessories)
+            acc_list.extend(pelts.wild_accessories)
         if "PLANT" in possible_accs:
-            acc_list.extend(Pelt.plant_accessories)
+            acc_list.extend(pelts.plant_accessories)
         if "COLLAR" in possible_accs:
-            acc_list.extend(Pelt.collars)
+            acc_list.extend(pelts.collars)
 
         for acc in possible_accs:
             if acc not in ["WILD", "PLANT", "COLLAR"]:
                 acc_list.append(acc)
 
-        if getattr(getattr(self.main_cat, "pelt", None), "scars", None):
+        if hasattr(self.main_cat.pelt, "scars"):
             if "NOTAIL" in self.main_cat.pelt.scars or "HALFTAIL" in self.main_cat.pelt.scars:
-                for acc in Pelt.tail_accessories:
+                for acc in pelts.tail_accessories:
                     if acc in acc_list:
                         acc_list.remove(acc)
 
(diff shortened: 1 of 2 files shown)


tests/test_handle_short_events.py
```diff
@@ -0,0 +1,164 @@
...
+        self.assertIn(self.test.main_cat.pelt.accessory, Pelt.plant_accessories)
+
+    def test_cat_gets_random_collar_accessory(self):
+        setattr(self.test.chosen_event, "new_accessory", ["COLLAR"])
+
+        self.test.handle_accessories()
+        self.assertIn(self.test.main_cat.pelt.accessory, Pelt.collars)
+
+    def test_notail_cats_do_not_get_tail_accessories(self):
+        setattr(self.test.chosen_event, "new_accessory", [random.choice(Pelt.tail_accessories)])
```
j-gynn: Not a big fan of using randomisation in unittests - the point of them is to be very static. Perhaps pick one of the tail accessories, or - if it's really important that every possible variation is tested - do a `for accessory in Pelt.tail_accessories` loop.

tests/test_handle_short_events.py
```diff
@@ -0,0 +1,164 @@
...
+        self.test.main_cat.pelt.scars = "NOTAIL"
+
+        self.test.handle_accessories()
+        self.assertNotIn(self.test.main_cat.pelt.accessory, Pelt.tail_accessories)
+
+    def test_halftail_cats_do_not_get_tail_accessories(self):
+        setattr(self.test.chosen_event, "new_accessory", [random.choice(Pelt.tail_accessories)])
```
j-gynn: As above - try not to use randomisation in unit tests.

scripts/events_module/handle_short_events.py
```diff
@@ -317,14 +317,14 @@ def handle_accessories(self):
             if acc not in ["WILD", "PLANT", "COLLAR"]:
                 acc_list.append(acc)
 
-        if "NOTAIL" in self.main_cat.pelt.scars or "HALFTAIL" in self.main_cat.pelt.scars:
-            for acc in Pelt.tail_accessories:
-                try:
-                    acc_list.remove(acc)
-                except ValueError:
-                    print(f'attempted to remove {acc} from possible acc list, but it was not in the list!')
+        if getattr(getattr(self.main_cat, "pelt", None), "scars", None):
```
j-gynn: May want to investigate `hasattr` instead of `getattr` here, at least for this outer one. Makes it a bit clearer what you're trying to do, and I can't find any documentation on the truthiness of `getattr`

susieoneil: Thanks.  `hasattr` is definitely clearer and will always be True or False.  Whereas `getattr` was only able to be False because of my setting the default to None.  I may remove the inner clause for now, as it works in it's current form. If I can mock the cat out of the test, I'll check for it then.

tests/test_handle_short_events.py
susieoneil: Yes, I agree.  I think I have figured out a good mock for it now.  I needed to abstract the Pelt call in the function out to a parameter.

tests/test_handle_short_events.py
susieoneil: As above - Fixed

scribblecrumb (approved):

PR #2476, opened 2024-06-16
Relationship/kin thought weights | Option to save dead relationships

Snowstar38:
## About The Pull Request

When generating cat thoughts, there will now be a chance for the other_cat to be a cat that they already have a relationship with, or one they are related to (determined by relationship_thought_chance and kin_thought_chance).
- Default settings are 20% relationship/20% kin, so 60% of other_cats for thoughts will still be pure RNG.
- The chance is for setting other_cat, not whether or not a thought will actually MENTION another cat, so the actual % of thoughts the player sees that contain a related cat's name will be lower.

In addition, an option (dead_relations) was added in the clansettings to enable retaining dead cats' relationships. This is false by default. The reason these two changes are bundled is because when dead_relations is enabled, dead cats can now have relationship thoughts rolled, so they will have a chance to think about cats they were friends with before dying.

This also fixes a bug I discovered during testing in which some cousins would not properly be displayed in the family tree (now cousins aren't checked until the full list of aunts and uncles is complete).

Thoughts are one of the main ways to get insight into the day-to-day lives of the cats in your Clan. Oftentimes, cats in Clangen can feel a bit disconnected with one another, especially in larger Clans; while there is a relationship and kin system, it has no impact on who they're thinking of. This change will hopefully make the Clan feel more "alive" by allowing cats' connections to affect thoughts.

It also makes dead cat thoughts more interesting, as now a dead cat is more likely to be thinking of their kin (or friends, if dead_relations is enabled) left behind in the living world.

Allowing dead cats to retain their relationships is a commonly requested feature; it is by default off and includes a tooltip warning mentioning possible performance impacts on large Clans (though it's not clear if this impact would be all that significant). Keeping that information is valuable for those who like to look back through their cats in StarClan and the Dark Forest and remember who they were friends or rivals with.

## Proof of Testing

image

## Changelog/Credits

Big changes:

- Added relationship-based and kin-based thought weights to make cats more likely to think about cats they already know or are related to, with the weights being adjustable in the config file
- Added a Clan setting, dead_relations (default: false), that saves and displays dead cats' relationships when enabled in the Clan settings page

Technical changes:

- Added a helper function called get_kin_groups to utility.py
- Added get_uncle_aunts, get_cousins, and get_grandkits, which are extremely similar to other get_relative functions
- Added get_distant_kin and recursive_get_relatives so we can gather cats from further away on the family tree, generally excluding "in-laws" and cousins of cousins
- Added two game_config settings, relationship_thought_chance and kin_thought_chance, to choose the % of thoughts that will be of that type (default: 20%)

Bugfix:

- Fixed an issue with init_cousins where cats were being checked as cousins before all aunts and uncles had been added to the parents_siblings list, resulting in some cousins not appearing in the family tree under certain circumstances


Ryos00 (approved):

scripts/cat/cats.py
```diff
@@ -1794,6 +2003,92 @@ def get_relatives(self, cousin_allowed=True) -> list:
...
+            new_relatives = set()
+            
+            for relative_id in relatives:
+                relative_cat = Cat.all_cats.get(relative_id)
+                if relative_cat:
+                    # don't add in-laws or cousins' cousins as they cause possible distant_kin
+                    # to balloon out of control with older, larger Clans
+                    sibling_mates = [
+                        mate_id for sibling_id in self.get_siblings()
+                        for mate_id in Cat.all_cats.get(sibling_id).mate
```
scribblecrumb, line 2075: I'm getting a crash when I try to run this PR, seems to be caused by this line
image

scripts/game_structure/game_essentials.py
```diff
@@ -429,7 +429,12 @@ def save_cats(self):
                 inter_cat.save_history(directory + "/history")
                 # after saving, dump the history info
                 inter_cat.history = None
-            if not inter_cat.dead:
+            if not game.clan.clan_settings["dead_relations"]:
+                if not inter_cat.dead:
+                    inter_cat.save_relationship_of_cat(directory + "/relationships")
+                else:
```
Lixxis: The code part "else: pass" can be removed, it does not change the way the code is handled.

scripts/utility.py
```diff
@@ -1526,6 +1526,44 @@ def change_relationship_values(
                     if log_text not in rel.log:
                         rel.log.append(log_text)
 
+def get_kin_groups(self, dead_chance=None, use_dead_chance=False):
```
Lixxis: Why is there a "self" in a utility function? I guess from the code it should be a cat. Can you rename the parameter?

scripts/utility.py
```diff
@@ -1526,6 +1526,44 @@ def change_relationship_values(
...
+    """
+    Retrieves the kin groups (close kin, kin, and distant kin) for the cat.
+    :param dead_chance: The chance of considering dead cats (optional).
```
Lixxis: Why is it called death_"chance"? If I understood it correct, it does not influence how likely a cat is added to the list. 

Can you explain why the two parameters "use_dead_chance" and "dead_chance" are separated? 
When you set "use_dead_chance" to True, it will always add the dead cats, only the chance is set to 1 they are not (intuitively a chance of 0 would also make more sense, that no dead cat will be added)

Lixxis (changes requested):
Some minor things I found, the files cat.py and load_cat.py I still have to look at (when I have more brain power)

scripts/cat/cats.py
```diff
@@ -1559,39 +1568,220 @@ def thoughts(self):
...
+                if cats_with_relationship:
+                    chosen_cat = choice(cats_with_relationship)
+                    other_cat = chosen_cat.ID
+                    # print("relationship thought: from: ", self.name, " to: ", chosen_cat.name)
+                else:
+                    # if no cats available, roll again for a chance at a kin thought at a slightly higher chance
+                    # since we rolled a special thought already and couldn't get one
+                    thought_choices_reroll = ['kin', 'normal']
+                    weights_reroll = [
+                        (1.25 * game.config["relationship"]["kin_thought_chance"]),
```
Lixxis, line 1588: Why is there still a multiplication when there is a setting for the chance? The reason for the settings is to reduce such values as much as possible.

(same as in other lines such as 1733)

scripts/utility.py
```diff
@@ -1665,6 +1665,43 @@ def change_relationship_values(
...
+    close_kin = {
+        "gen_parents": [cat.all_cats.get(cat_id) for cat_id in cat.get_parents()],
+        "gen_siblings": [cat.all_cats.get(cat_id) for cat_id in cat.get_siblings()],
+        "gen_children": [cat.all_cats.get(cat_id) for cat_id in cat.get_children()],
+        "gen_mates": [cat.all_cats.get(cat_id) for cat_id in cat.mate],
+        "gen_former_mates": [cat.all_cats.get(cat_id) for cat_id in cat.previous_mates]
+    }
+
+    kin = {
+        "gen_grandparents": [cat.all_cats.get(cat_id) for cat_id in cat.all_cats if cat.all_cats.get(cat_id).is_grandparent(cat)],
```
Lixxis, line 1686: The "get_kin_groups" function is not very performant. To improve it at least a little add the new "kin" to the inheritance class. It seems it was done but not used?

scripts/utility.py
```diff
@@ -1665,6 +1665,43 @@ def change_relationship_values(
                     if log_text not in rel.log:
                         rel.log.append(log_text)
 
+def get_kin_groups(cat, include_dead=True):
```
Lixxis, line 1668: Is it important at this point to already have the cat class? It would be more performant to just use the id's as long as possible and then fetch the cat, when the information from them is needed.

scripts/cat/cats.py
```diff
@@ -1559,39 +1568,220 @@ def thoughts(self):
...
+                
+            # Roll kin-based thought
+            if thought_type == 'kin':
+                kin_group_choice = randint(0, 100)
+                # If dead_chance == 1, include dead cats, otherwise exclude them
+                include_dead = (dead_chance == 1)
+                close_kin, kin, distant_kin = get_kin_groups(self, include_dead=include_dead)
+                
+                # 50% chance of choosing close kin
+                if kin_group_choice >= 50:
```
Lixxis, line 1601: The chances of each type of kin group can and should also be added in the settings.

scripts/cat/cats.py
```diff
@@ -1559,39 +1568,220 @@ def thoughts(self):
...
+                        pass
+                
+                # 15% chance of choosing distant kin
+                elif kin_group_choice >= 0:
+                    distant_kin_cats = [cat for sublist in distant_kin.values() for cat in sublist]
+                    if distant_kin_cats:
+                        chosen_cat = choice(distant_kin_cats)
+                        other_cat = chosen_cat.ID
+                        #print("distant kin thought: from: ", self.name, " to: ", chosen_cat.name)
+                    else:
```
Lixxis, line 1627: 'else: pass' 
sections doesn't change anything and should be removed.

scripts/cat/cats.py
```diff
@@ -1559,39 +1568,220 @@ def thoughts(self):
...
-                or (other_cat not in self.relationships)
-            ):
-                other_cat = choice(list(all_cats.keys()))
-                i += 1
-                if i > 100:
-                    other_cat = None
-                    break
-        # for dead cats
+            # Roll relation and kin thought chances
+            if thought_type == 'relationship':
```
Lixxis, line 1572: The thought type should be handled before the "if where_kitty == " statement. In that way there won't be that many code duplications in the class.

Lixxis (changes requested):

Lixxis:
@Snowstar38 Sorry that it (to be specific I) took so long (also there was the feature freeze in between)
I have some comments to address + there is as well a merge conflict which should be resolved.

scribblecrumb:
Closing as stale for now. Please make a new PR when you get around to applying feedback!

PR #2477, opened 2024-06-16
"FAWN" white patch

Snowstar38:
## About The Pull Request

Adds a new white patch, FAWN, inspired by both deer and by cats like Blossomfall in the main series.

Clangen doesn't actually have a single white patch that adds smaller white spots similar to how Blossomfall is described. Currently, the only way for a cat to get something like it is if they are a tortie, get the FRECKLED pattern, and happen to have their secondary color be white. Now there will be a bit more variety.

## Proof of Testing

image
image

## Changelog/Credits

- add FAWN white patch


anjubatus (changes requested):
I like the idea of this patch, but you should use a pixel brush instead of a regular, softer brush for this. We're trying to phase out unneeded/inconsistent transparent pixels that appear when using soft brushes like you can see here. Using transparency is still okay, but it should be intentional and consistent!
image

Snowstar38:
> I like the idea of this patch, but you should use a pixel brush instead of a regular, softer brush for this. We're trying to phase out unneeded/inconsistent transparent pixels that appear when using soft brushes like you can see here. Using transparency is still okay, but it should be intentional and consistent!)

Is this any better? I hope I was able to capture the style - I made areas slightly transparent where spots are meant to be smaller. Let me know what you think!~

fawn

anjubatus (commented):
Hi, sorry for leaving this for so long! Tbh I think it's wiser to close this PR, this white patch needs a lot of work still and may need to even be re-done to make the spots more even and natural, and I think it's easier to work on that if the patch is first posted into the Cat Sprites thread on discord to receive feedback and edits there.

PR #2393, opened 2024-06-03
Bug fix. Fixes #2275

susieoneil:
## About The Pull Request

- Modified reveal_murder in History class to record the current moon. Added a new revelation_moon key to both murderer and victim history dictionaries. 
- Modified get_murder_text and get_text_for_murder_event in ProfileScreen to add revelation_moon to both murderer and victim histories.
- Minor adjustments to variable names, line lengths, comments, etc.

## Linked Issues

Fixes: #2275 

## Proof of Testing

Followed the same steps from the issue:
1. skip moons till a cat kills another
2. skip moons till its found out
3. turn on "show moons" in the killer's history.

Murderer history
murder-history

Victim history
victim-history

Issue #2275, opened 2024-04-28
[BUG] history does not show moon a murderer was discovered


selkirks:
**Type**: (only select one)
...ui..?

**Describe the bug**
despite having "show moon when an event occurred" on in a cat's history, the moon the cat was found out as a murderer is not listed

**Grade**: (only select one)
- Non-vital - This bug can easily be ignored and does not have a large impact on play

**Reproduce**
1. skip moons till a cat kills another
2. skip moons till its found out
3. turn on "show moons" in the killer's history

**Patrol ID + Outcome Print (if applicable)**:

**Commit # or Game Version Number**: 0e554eee9ae64863e6ffb1aeef0c77ce0c5b2405

**Additional context**

**Screenshots**
![image](https://github.com/ClanGenOfficial/clangen/assets/126827015/833300c9-2922-4b3a-8636-84847dc18813)

scribblecrumb:
Looks like `reveal_murder` in History class doesn't actually record the current moon, just an oversight.  Need to add a new `revelation_moon` key to both murderer and victim history dictionaries.  Also adjust `get_murder_text` in profile screen accordingly.

PR #2479, opened 2024-06-16
Bugfix: Victim text generation.

susieoneil:
## About The Pull Request

Fixed an error in my previous bug-fix #2393.  Error making victim a murderer.

## Linked Issues
#2275 #2393
Exposed in conversation stating here:
https://discord.com/channels/1003759225522110524/1005586496142704800/1251871993104437310

### DIFF
--- a/scripts/cat/history.py
+++ b/scripts/cat/history.py
@@ -101,10 +101,10 @@ class History:
                     {
                     "victim": ID,
                     "revealed": bool,
-                    "moon": moon
+                    "moon": moon the murder occurred
                     "revealed_by": ID of the discoverer
                     "revelation_moon": moon the murder was revealed
-                    "revelation_text": revealed death history
+                    "revelation_text": revealed murder history
                     },
                 ]
             "is_victim": [
--- a/scripts/cat/history.py
+++ b/scripts/cat/history.py
@@ -645,7 +645,6 @@ class History:
         return cat.history.murder
 
     @staticmethod
-
     def reveal_murder(cat, other_cat, cat_class, victim, murder_index):
         """ Reveals the murder properly in all associated history text.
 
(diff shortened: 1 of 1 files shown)


scripts/cat/history.py
```diff
@@ -677,12 +676,13 @@ def reveal_murder(cat, other_cat, cat_class, victim, murder_index):
...
                 victim_history["revelation_moon"] = game.clan.age
                 if not other_cat:
-                    murder_history["revelation_text"] = \
-                        "The truth of {PRONOUN/m_c/poss} murder is known to the Clan."
+                    victim_history["revelation_text"] = \
+                        "The truth of their murder is known to the Clan."
                 else:
                     victim_history["revelation_text"] = \
-                        "The truth of {PRONOUN/m_c/poss} murder was discovered by [discoverer]."
+                        "The truth of their murder was discovered by [discoverer]."
```
scribblecrumb: I'm not sure I understand why we're removing the pronoun tag?  Can you explain the reasoning? /gen

susieoneil: Check my screenshots on #2393.  This text appears appended to the initial murder text.  I did change it to [victim] in my first commit, but realised the name didn't really need to be there.  
Note: The {PRONOUN/m_c/poss} part is the murderer, so that was wrong anyway.

scribblecrumb: Hmm, this is still a pronoun that needs to be capable of being dynamic, though, isn't it?  It refers to the victim, so it needs to be able to change to the victim's pronouns, thus the pronoun tag is necessary.  Does "m_c" here refer to the murderer?  If so, we should change it to "r_c" or whatever it needs to be to refer to the victim cat.

susieoneil: Seems it was correct the first time.  I think I get it now.  m_c is the main cat in the murder history, but the victim in the victim history.  When I was testing it the first time, I didn't realise both cats used them/them pronouns.  I changed the murderer to He and the victim to She and tried again.  I then found their history entries in the save files to see how they had been saved. 
badgerpelt
Shellblur
victimhistory

scribblecrumb (approved):

Issue #2436, opened 2024-06-11
[PROJECT] `Cat` Class Rework


j-gynn:
**Description**
This involves a rework of `Cat`. At present, related data is entirely disparate, `str` is pretty much the de-facto variable type, there are so, so many interconnected dependencies that it kind of hurts my soul, testing the cat itself is really difficult...

**Scope of Project**
Please choose from following list, feel free to give further details:
- Large Project (substantial new feature, like a new page for a new purpose)

This is **huge**. I will be doing this in separate PRs, because I'm not a madman. Ideally, I'd like to have `__init__` have a MAXIMUM of 5-7 inputs, but I think that's a pipe dream. The current proposal would have 11, for perspective. I don't need help with this project, I just need to put it somewhere so that everyone can keep track of the proposed changes, what's already implemented and the future of `Cat`.

**What this is NOT:**
- Not a JSON rework. I'm not touching that with a ten foot rusty spoon if I can avoid it.

Checkmark: merged into `development`.
## Major Changes Summarised
- Move `Personality` to a different file to deal with some other time #2466 
- Decouple code from unrelated code by appropriate use of nesting, grouping variables by purpose in new classes
- Reduce the number of ""helper"" functions in `Cat` by moving them to the appropriate/related classes
- STOP USING CAT TO STORE THE ENTIRE CLAN. Please, we have a `Clan` class, why is that not what we use for who belongs to the clan?!
- 

### DELETIONS
- [ ] `exiled`
- [ ] `outside` (convert to property)
- [ ] `driven_out`
- [ ] `also_got`

### Type changes
- [ ] `age` to Enum #2407 #3173
- [ ] `status` to Enum #2417 
- [ ] `gender` to a Class with helper functions **(discuss)**
- [ ] `ID` to Int **(DANGER ZONE)**

### Name changes
- [ ] `gender` -> `sex` (to remove confusion around what this is)
- [ ] `ID` -> `cat_id` (should be lowercase and `id` is taken by a Python builtin)
- [ ] `favourite` -> `favorite` (Murica) #2437 
- [ ] `mate` -> `current_mates` (in line with `current_apprentices` and the poly update)
- [ ] `previous_mates` -> `former_mates` (in line with `former_apprentices` and `former_mentor`)

## Classes (grouping similar information together for clarity - and getting it out of `cats.py`!)
- [x] Age
  - `_age`
  - `moons`
  - `is_baby` (currently a function, make it a property)
- [ ] Identity
[...]

PR #2482, opened 2024-06-16
[CLEANUP] Clan subclassing

j-gynn:
## About The Pull Request

- Clan is now three separate classes: BaseClan, PlayerClan and OtherClan, each in their own file in a new directory, `scripts/clan`.
- PlayerClan and OtherClan inherit from BaseClan.
- StarClan has been deleted as it doesn't actually do anything.
- `seasons` variable renamed to `SEASONS` as it is a "constant".
- Removed the `CAT_TYPES` variable as it is dead and has no references.

Say it with me, gang. Happy devs... happy life. Separating the domains like this means that less dependencies are ensnared amongst each other, making it easier to not have things like the `clan_class` variable, which I **will** eradicate. Just not in this PR. I wanted to try doing a "smaller" change to make it easier to review & test.

## Linked Issues
Tangentially related to #2436 but mostly as a result of trying to write unit tests for murders and attempting to scaffold the clan that way.

## Proof of Testing

https://github.com/ClanGenOfficial/clangen/assets/48025294/549979a9-e5d8-4de7-9b76-4c54c560f578


scripts/clan/starclan.py
```diff
@@ -0,0 +1,40 @@
+import pygame
+
+from scripts.cat.sprites import sprites
+
+
+class StarClan:
```
scribblecrumb: Could we rename this to DeadClan?  Since this class handles all the dead cats, not just StarClan (or, at least, i presume it does, since all dead cats fade regardless of group), it seems like that would make more sense.

j-gynn: I was debating doing that but then I hesitated 😂 I'll make that change.

scripts/clan/baseclan.py
```diff
@@ -0,0 +1,42 @@
...
+        "Leaf-fall",
+        "Leaf-fall",
+        "Leaf-bare",
+        "Leaf-bare",
+        "Leaf-bare",
+    ]
+
+    current_season = "Newleaf"
+    starting_season = "Newleaf"
+    biome = "Plains"
```
Ryos00: more a request than an issue, but can the default biome be forest? it was the First and Only biome for a while, so it makes more sense in my brain :3

Ryos00 (approved):

scripts/clan/baseclan.py
j-gynn:
```suggestion
    biome = "Forest"
```

j-gynn: Fixed :) That does make more sense to me too haha

PR #2481, opened 2024-06-16
Bugfix: Cat List

scribblecrumb:
## About The Pull Request
- Fixes page arrows not disabling correctly
- Fixes #2480 

## Linked Issues
Fixes #2480 
Fixed #2483

## Proof of Testing
image
Arrows disable as expected, all cats show in the cat list.

## Changelog/Credits
- Fixes page arrows not disabling correctly
- Fixes the bug preventing cats from displaying on the cat list

PR #2488, opened 2024-06-18
Fix for cat list breaking bug

### LINKED ISSUE #2480: [BUG] Cat list page completely broken
**Type**: UI

**Describe the bug**
my cat list page just randomly decided to break completely. i have 2 pages of cats, yet i only see two or three cats on each. names are even cut off. every cat appears just fine on the camp screen, but the list specifically is busted. aside from the cats in the screenshot, i can't click on any others, like they're just completely missing. sorting does nothing to fix it.

**Grade**: Dire, I'd say?

**Reproduce**
i'm.. honestly not sure. this was a random thing that happened after a moonskip. i had 2 pages of cats earlier and everything was fine. i'm on moon 44, with 22 living cats, 2 starclan cats (guide included) and a pregnancy was announced the moon before.

**Patrol ID + Outcome Print (if applicable)**: n/a

**Commit # or Game Version Number**: version 50c25beb

**Additional context**
i don't think anything specific was done to trigger this? i can always try to remember if i did anything out of the ordinary, but i was just playing casually until i moonskipped and this happened.

**Screenshots**
below are both of my cat pages, plus a screen recording to showcase the bug in more detail.
https://github.com/ClanGenOfficial/clangen/assets/172712347/20d0d948-d87f-4144-a1ec-a7cb6822bd7c (video)
image
image

Snowstar38:
## About The Pull Request

Fixes a game-breaking bug with the Clan list that could permanently disable a large chunk of the cat list buttons until the game is restarted. When using "back_button" to return to a list view with 15 or less cats, the cat_display container would be created with dimensions appropriate for the *smaller* number of cat buttons, meaning that if COTC had one row and your Clan had 4, all but the first row of your Clan will be cut off. Switching group views would not actually fix the cat_display dimensions because it would be modifying the preexisting cat_display. Now if it's less than 20 cats, the cat_display is deleted and a new, properly sized, one is generated. This might be inefficient but it's the only solution I could find that works reliably for now as trying to directly set things like self.cat_display.relative_rect didn't work.

## Linked Issues

Fixes: #2480

The issue would happen like this:

1. Open Clan list

2. Switch the view to another group (StarClan, COTC, etc) that has at least 1, but no more than 15, cat(s)

3. Click on one of these cats' sprites to open the profile view

4. Click the back button

Now all lists of cats that contain more cats than the smaller group will be broken. This will happen 100% of the time you follow these steps given that the conditions (more rows/cats in the first page of your living Clan than the first page of the other view) are met.

## Proof of Testing

Before:
https://github.com/ClanGenOfficial/clangen/assets/104292791/7ba3ba37-22fc-4c52-8622-73ee54b5fd41

After:
https://github.com/ClanGenOfficial/clangen/assets/104292791/e9244063-4928-448d-9ca8-fd1f3100ff10

## Changelog/Credits

- Fixes a bug that would disable and hide cat sprites and buttons in the list view until game restart

### DIFF
--- a/scripts/screens/ListScreen.py
+++ b/scripts/screens/ListScreen.py
@@ -555,11 +555,33 @@ class ListScreen(Screens):
                 manager=MANAGER
             )
         else:
-            self.cat_display.text_theme = (
-                get_text_box_theme("#text_box_30_horizcenter")
-                if self.death_status == "living"
-                else "#text_box_30_horizcenter_light")
-            self.cat_display.update_display(current_page=self.current_page, cat_list=self.current_listed_cats)
+            if self.cat_display.relative_rect != "Rect(0, 0, 710, 480)":
+                self.cat_display.kill()
+                self.cat_display = UINamedCatListDisplay(
+                    scale(pygame.Rect((0, 0), (0, 0))),
+                    container=self.cat_display_container,
+                    object_id="#cat_list_display",
+                    starting_height=1,
+                    cat_list=self.current_listed_cats,
+                    cats_displayed=20,
+                    x_px_between=240,
+                    y_px_between=200,
+                    columns=5,
+                    prev_button=self.display_container_elements["previous_page_button"],
+                    next_button=self.display_container_elements["next_page_button"],
+                    first_button=self.display_container_elements["first_page_button"],
+                    last_button=self.display_container_elements["last_page_button"],
+                    current_page=self.current_page,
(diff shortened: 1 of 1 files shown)


scribblecrumb:
Ah, Snow, thanks for putting in the work to find a fix for this, but I'm sorry to say I already fixed it in #2481 ^^;

Snowstar38:
Oops! I somehow didn't see that 🤣 my bad! Oh well - at least I learned something about the UI, haha.

PR #2491, opened 2024-06-18
[CLEANUP] SPaG Review #1

j-gynn:
## About The Pull Request
Cleaning up the spelling, punctuation and grammar in ClanGen by manually reading every... single... string. Insane? Probably, yeah.

This covers `ceremonies`, `conditions` and `thoughts`. I need a break between all of this, and then I'll slowly start to tackle patrols.

Regarding thoughts: where I've noticed, I've pulled out things with constraints. That isn't the goal of this review, so I will 100% have missed many and I don't care <3 /lh

When words word good, is good.

## Proof of Testing

Pronouns test passes? Idk what else to say here, this is... it's strings?

## Changelog/Credits
SPaG cleanup!


resources/dicts/thoughts/alive/apprentice.json
```diff
@@ -19,18 +19,6 @@
...
-            "Is dreading {PRONOUN/m_c/poss} apprentice duties",
-            "Fell into the nearby creek yesterday and is still feeling damp",
-            "Is having a hard time keeping up with {PRONOUN/m_c/poss} training",
-            "Was tasked with lining nests with fresh moss today",
-            "Is dreaming of someday making {PRONOUN/m_c/poss} Clan proud",
-            "Is thinking about the time {PRONOUN/m_c/subject} caught a huge rabbit",
-            "Practices some battle moves",
-            "Is helping to reinforce the camp wall with brambles",
-            "Is daydreaming about having a mate and kits someday",
-            "Had quite the adventure today",
```
scribblecrumb, line 33: Was the removal of these intentional?

resources/dicts/thoughts/alive/deputy.json
```diff
@@ -306,11 +305,8 @@
         ]
     },
     {
-        "id": "story_deputy_apprentice",
+      "id": "story_deputy_to_apprentice",
         "thoughts": [
-            "Is recalling an old story that might inspire the Clan in tough times",
-            "Feels inspired to craft a new legend based on recent events",
-            "Uses the stories of old to help guide {PRONOUN/m_c/self} decisions",
```
scribblecrumb, line 313: Same question here

resources/dicts/thoughts/alive/apprentice.json
j-gynn: Yes - they're duplicates of content found in general with a wider scope. Confusingly because of the way these are laid out, if you want to make a general apprentice-wide thought (i.e. for warrior, medcat or mediator apprentices), it has to live in general.

scribblecrumb: I see!  I figured it was something like that, just wanted to check

resources/dicts/thoughts/alive/deputy.json
j-gynn: Yep - intentional. These ones lived in the "story deputy to apprentice" tag, but they also lived in basic story deputy. Since they don't explicitly identify apprentices, nor particularly require them, I removed them from the apprentice variant.

resources/dicts/thoughts/alive/apprentice.json
j-gynn: I think that a case could be made that "practicing battle moves" should be warrior apprentice only, but I don't think it's unreasonable for all apprentices to be learning battle moves anyway just in case.

scribblecrumb (approved):

PR #2518, opened 2024-06-22
[FEATURE] Classic Illnesses and Conditions

larkgz:
## About The Pull Request
* Enables illnesses, injuries, and conditions in classic mode, including grief. They can't work if they're sick like in expanded
* A single medicine cat is enough for the whole Clan ~~assuming the Clan has less than 1,000,000 cats~~
* Specific herb amounts aren't shown on the med den screen
* All herbs are displayed on med den screen graphic 
* A herb will be to improve the condition regardless of actual herb amounts (though this is kind of invisible because the log is hidden in classic mode? I'm a little worried the whole menu would be confusing for classic players, though)
* Random herb amount every moon to possibly trigger herb events and to change herb amount display text 

See previous discussions on Discord.

## Linked Issues
See `classic mode overhaul` in development chat on Discord. 

## Proof of Testing
image

image

image

image

image

image

## Changelog/Credits
Enables illnesses, injuries, and conditions in classic mode.


scripts/cat/thoughts.py
```diff
@@ -201,47 +201,43 @@ def cats_fulfill_thought_constraints(main_cat, random_cat, thought, game_mode, b
...
+                    if main_cat.permanent_condition:
+                        if not [i for i in main_cat.permanent_condition if
+                                i in thought["perm_conditions"]["m_c"]] and \
+                                "any" not in thought['perm_conditions']["m_c"]:
                             return False
+                    else:
+                        return False
 
-        if game_mode != "classic" and "perm_conditions" in thought:
+                if "r_c" in thought["perm_conditions"] and random_cat:
```
Lixxis, line 231: The old line 244 contains a condition and was removed, but this line indent didn't change, is this intended?

scripts/events.py
```diff
@@ -582,15 +581,16 @@ def herb_gather(self):
...
-                    game.clan.herbs.pop(herb)
-            if not int(random.random() * 5):
-                new_herb = random.choice(HERBS)
-                game.clan.herbs.update({new_herb: 1})
+            # in classic, you have a random amount of herbs.
+            # the actual herb doesn't matter; it's just the count.
+            herb_owned = random.choice(HERBS)
+            # values that will change the text that is displayed to say how many herbs you have
+            required_herbs = get_living_clan_cat_count(Cat) * 4
+            adjustment_factor = random.choices([0.25, 0.5, 1, 2, 3], [1, 2, 3, 2, 1], k=1)[0]
```
Lixxis:
```suggestion
            adjustment_factor = random.choices([0.25, 0.5, 1, 2, 3], weights=[1, 2, 3, 2, 1], k=1)[0]
```
Just using the parameter name would communicate more clear why this approach is used at a quick glance.

scripts/events_module/condition_events.py
```diff
@@ -235,9 +235,9 @@ def handle_illnesses(cat, season=None):
         This function handles overall the illnesses in 'expanded' (or 'cruel season') game mode.
```
Lixxis:
```suggestion
        This function handles overall the illnesses.
```

Lixxis (commented):
Still have to run it locally but looking good for now. 
One thing I have in mind, is that the clan focuses can now be also expand since illnesses and injuries are possible in the classic mode.

scripts/events_module/condition_events.py
j-gynn: If we are changing this, it should be "This function handles illnesses overall", but it's not a particularly helpful description for what this function does. The name is "handle_illness", the definition shouldn't be "it handles illness". Would it be possible to get a more useful description added? I know that's not _technically_ what this PR is for, but... helpful improvements to documentation are always a win!

scripts/events.py
larkgz: Sure. Although, there are a lot of random.choices calls in the same file that don't use the parameter name either...

scripts/cat/thoughts.py
larkgz: Those lines don't correspond to each other. The old 244 corresponds to the the new line 240.

Looking at it again, there seems to be some duplicate logic. But maybe there's some kind of edge case it's dealing with? If it is a duplicate, I'd rather it be fixed in another PR.

Ryos00 (approved):

scribblecrumb:
Looks good!  Gonna approve, just one last questions: Would it be possible to still have the med cat graphic change instead of being statically full?  It feels weird going to the med den to see "all the herb stores are bare", but also see a completely stocked graphic.

larkgz:
@scribblecrumb I set it so it will draw the same number of herb pictures (??) as “herbs” the Clan has.  It’s possible for there to be some mismatch if the required herb amount is drastically different from the amount of pictures, but I think it will at least be closer — and you definitely won’t get a picture with many herbs if you have no herbs.

scribblecrumb (approved):

PR #2537, opened 2024-06-26
Bugfix : minor litter chances fix

Chinch-Bug:
## About The Pull Request

Noticed that the line meant to increase litter chances when single parentage disabled is doing the opposite


scribblecrumb (approved):

scribblecrumb:
Thanks for the bugfix!  Are you part of the clangen discord and, if so, would you like access to the developer side?

Chinch-Bug:
I'm @ chinchbug on the server! But nah, probably not, just happened to notice this while doing mod-related stuff

PR #2541, opened 2024-06-26
Add constraint priority system

### LINKED ISSUE #2539: [CODE] New Weight System
**Description**
This issue is for a new weight system to be in place so that patrols or ShortEvents that have more constraints are more likely to be picked if the constraints are met.

This system needs to be in place before more possible constraints can be added to the patrols or short events.

This helps to give more varied patrols, and let's harder to access patrols show more often then more common patrols.

Currently @j-gynn is working on it to my knowledge.

j-gynn:
## About The Pull Request

Adds a kind of priority system for constraints. The way this works is the following:

1. The number of types of constraint (and in some cases, the rarity of that particular constraint) are added up for each patrol.
2. For each tier of rarity, the system runs a flat 1/x chance to only have that pool available.
3. This check runs in order of rarity, rarest first - so the rarest tier gets a full 1/x, second rarest gets (x-1/x * 1/x) etc.
4. If none of the rolls pass, then the list returns to the original behaviour of a flat chance across the board
5. The subgroup is then rolled according to the weights on the JSON.

Allowing rarer patrols to show up is always a good thing!

## Linked Issues

Closes #2539 

## Proof of Testing

For now, I need to get some feedback about the weighting as well as finding any bugs, so I'm setting it as a draft PR.

[Link to the Discord testing channel](https://discord.com/channels/1003759225522110524/1255576559725707324)

## Changelog/Credits
- Enhanced patrol choice system to slightly prioritise more heavily constrained patrols when requirements are met.


j-gynn:
Closing this PR as I'm not particularly happy with the implementation. I'm going to go back to the drawing board on this and that'd be out of the scope of the feature-freeze for this version.

PR #2572, opened 2024-07-03
Button bugfixes

### LINKED ISSUE #2557: [BUG] View Dead/Living button not behaving appropriately
**Type**: UI 

**Describe the bug**
View Living/Dead button doesn't switch with the clan screen. Clicking View Living while on the regular catlist does nothing but change it to View Dead.

**Grade**: Non-vital

**Reproduce**
1. On the cat list page, click 'View Dead'
2. From the Starclan screen, click 'Patrol', 'Camp', or 'Events'
3. Click back to Cat List from Patrol. The button will be unchanged, but the list will be living cats.

**Patrol ID + Outcome Print (if applicable)**: null

**Commit # or Game Version Number**: dd218d29

**Additional context**
I imagine it's related to the old Sort living cats by death bug since it's triggered the exact same way. Appears on both classic and extended.

**Screenshots**
Screenshot 2024-06-28 231639

Snowstar38:
## About The Pull Request

- Fixes an issue where the "show living" and "show dead" buttons would display improperly when the player has a dead cat list open and switches to another menu, then switches back to the cat list with the main menu "cat list" button (Fixes: #2557)

- Fixes a bug with cat lists where if you entered a cat's profile before navigating away to another screen such as camp or patrol, the current group (StarClan, DF, etc) you had before clicking on the most recent cat sprite would load again when returning to the cat list [example: go to StarClan, open a SC cat's profile, go back to camp, go back to cat list, StarClan is still displayed. This is unintended behavior as it only happens when a cat's profile is opened and the last_list_forProfile flag was never cleared]

- Fixes a bug with cat lists where if you entered a StarClan/DF/UR cat's profile and opened a screen from there (such as the "choose mate" or "see relationships" screen), then went back to the cat list from their profile, the displayed group would be switched back to the living Clan

~~- Fixes "Dens" button displaying in camp view~~

~~- Changed the Dens menu so that it will automatically shift when the moons and seasons widget is enabled~~


Ryos00 (approved):

scribblecrumb:
I'm not sure how I feel about the changes to how the den button displays with the m+s box.  Previously, if m+s was activated then the dens button simply displayed on the event page only (as the moon and season info is already on that page in a different manner).  Your new method leaves things feeling a bit crowded, though I understand the QoL it provides.  What if we shift it over to the other corner of the screen when m+s is active instead?  So that it's under the Settings button.

Snowstar38:
@scribblecrumb Got it, does this work? :)

scribblecrumb:
When moving between Cat List and Patrol screens, the Den button does a little blink.  Not sure if this displays for others as well.  It's probably due to killing the den button every single time the screen is changed.  Even though it doesn't need to be killed between Cat List and Patrol (due to being in correct spot), it's still killed and rebuilt, leading to a blink.  I'd suggest adding a check to `mns_ui_offset` to see if we're moving between those two screens.  If we are, don't kill and rebuild.

scribblecrumb (changes requested):
Just in case you didn't notice before, I'd made a comment about some changes that need to be made ^^

Snowstar38:
I removed the changes to the m+s widget and dens bar!

scribblecrumb:
Looks good @Snowstar38!  There's some conflicts to resolve first.  I was gonna do it for you but it's a few more conflicts than I thought, so I'll let you do it instead.  I don't wanna mess it up accidentally haha

scribblecrumb (approved):

PR #2596, opened 2024-07-29
Switch Run Scripts To Use python3

Sumwatt0:
## About The Pull Request
Switches run scripts from using python to using python3. 

python points to python2 on many systems, and even where it doesn't, using python3 is more explicit, clear and safe.

## Proof of Testing
Screenshot 2024-07-29 at 1 38 16 PM

It opens clangen like normal.

## Changelog/Credits
[ENHANCEMENT]: Run scripts now use python3

Credit: (Sumwatt/Sumwatt0)

### DIFF
--- a/run.bat
+++ b/run.bat
@@ -1,2 +1,2 @@
 %AppData%\Python\Scripts\poetry install --no-root
-%AppData%\Python\Scripts\poetry run python main.py
+%AppData%\Python\Scripts\poetry run python3 main.py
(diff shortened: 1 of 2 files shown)


archanyhm:
While the reasoning is technically correct in regular python contexts, the run scripts utilise poetry which sets both `python` and `python3` — if applicable — to the version the configured environment uses. Therefore, there's no real benefit to explicitly call `python3`.

Poetry also omits the `3` in their basic usage documentation found here: https://python-poetry.org/docs/basic-usage/#using-poetry-run

Sumwatt0:
That makes sense, I guess I'm just hardwired into using python3 every time lol.

PR #2628, opened 2024-08-12
Fixes: #2463 (cats dying in "taken by Twolegs" mass event)

### LINKED ISSUE #2463: [BUG] "Taken by Twolegs" extinction event kills some cats
**Type**: Code

**Describe the bug**
For the mass extinction event where several cats are taken by Twolegs, some cats actually go to the COTC tab, but some of them (those who are not listed in the event) are dead instead, without a death event appearing for them.

**Grade**: Important

**Reproduce**
1. Activate mass extinction events in settings
2. Skip moons until the "taken by Twolegs" event happens
3. Check which cats were listed in the event and if there are any cats that have mysteriously died.

**Commit # or Game Version Number**: 6c5f99b55

**Screenshots**
In this example, Pearecho, Flamepaw and Oliveclaw are dead, but there's no dead event for them; you can see Oliveclaw appears in the buttons under the taken by Twolegs event but died instead of being taken away.
image
image
image

KiraDragoness:
## About The Pull Request
It seems that iterating over dead_cats and removing cats from it at the same time was causing some issues. I tested this and the bug didn't happen after triggering the event. I added the implicated cats into a new variable and then used it to remove them from dead_cats.

## Linked Issues
Fixes: #2463

## Proof of Testing
All cats listed in the event appear in the buttons below:
image
image
image

All the implicated cats appear correctly as lost cats:
image

No cats have misteriously died after the event (showing the last dead cat, the previous moon, after sorting by death):
image
image

## Changelog/Credits
Changelog: Fixed cats wrongly dying after the taken by Twolegs mass event.
Credits: Kira

### DIFF
--- a/scripts/events_module/handle_short_events.py
+++ b/scripts/events_module/handle_short_events.py
@@ -406,10 +406,13 @@ class HandleShortEvents:
             if self.main_cat not in self.dead_cats:
                 self.dead_cats.append(self.main_cat)  # got to include the cat that rolled for death in the first place
 
+            taken_cats = []
             for kitty in self.dead_cats:
                 if "lost" in self.chosen_event.tags:
                     kitty.gone()
-                    self.dead_cats.remove(kitty)
+                    taken_cats.append(kitty)
+            for kitty in taken_cats:
+                self.dead_cats.remove(kitty)
                 self.multi_cat.append(kitty)
                 if kitty.ID not in self.involved_cats:
                     self.involved_cats.append(kitty.ID)
--- a/scripts/events_module/handle_short_events.py
+++ b/scripts/events_module/handle_short_events.py
@@ -411,11 +411,12 @@ class HandleShortEvents:
                 if "lost" in self.chosen_event.tags:
                     kitty.gone()
                     taken_cats.append(kitty)
-            for kitty in taken_cats:
-                self.dead_cats.remove(kitty)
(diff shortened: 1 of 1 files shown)


scripts/events_module/handle_short_events.py
```diff
@@ -406,10 +406,13 @@ def handle_mass_death(self):
...
 
+            taken_cats = []
             for kitty in self.dead_cats:
                 if "lost" in self.chosen_event.tags:
                     kitty.gone()
-                    self.dead_cats.remove(kitty)
+                    taken_cats.append(kitty)
+            for kitty in taken_cats:
+                self.dead_cats.remove(kitty)
                 self.multi_cat.append(kitty)
```
scribblecrumb, line 414: Nesting this and the next line within the `taken_cats` for loop means it will never happen for the cats who are just killed.  Both of those lines need to happen for all cats, regardless of lost or dead.

I'd suggest the removal loop should just be:
```
            for kitty in taken_cats:
                self.dead_cats.remove(kitty)
```
while the multi_cat appending and ID appending should happen in the `self.dead_cats` loop as it did before

KiraDragoness: I updated it! I hope I understood correctly what you meant!

scribblecrumb (approved):

PR #2641, opened 2024-08-19
seasonal thoughts 

Yzzuwyzzuf:
## About The Pull Request

I was directed here when a senior dev, Scribble saw the mod I was making and suggested I put in a PR. this request is about thoughts cats can have in each season. 

I believe this would benefit clangen by adding more immersion and personality. not every cat just goes through leaf-bare and doesn't think about how cold it is or go through the heat of Greenleaf without even thinking about hiding in the shade or taking a dip in a nearby river. And elders and kits love to complain of course. I think clangen needs something like this to add a bit more life into the day by day and I hope you agree too :)

## Proof of Testing

my proof of testing is the mod ive worked on, "fuzzy's expanded text mod" on discord. I've also added a file with all my general seasonal thoughts code and medicine cat seasonal thoughts code. its not set to work directly from the file but it'll work just fine if each part is copy pasted easy
[seasonalthoughts.txt](https://github.com/user-attachments/files/16666895/seasonalthoughts.txt)

## Changelog/Credits

I believe the changlog would mention seasonal thoughts, I'm not the best with wording lol and my credit is Yzzuwyzzuf on discord as that's the platform im most active on

### DIFF
--- a/resources/dicts/names/names.json
+++ b/resources/dicts/names/names.json
@@ -298,7 +298,7 @@
     "Acorn", "Almond", "Antler", "Aspen", "Badger", "Bark", "Bat", "Beech", "Birch", "Boar", "Branch", "Branch", "Buck", "Canopy",
     "Cedar", "Civet", "Deer", "Doe", "Elk", "Fern", "Fern", "Fir", "Forest", "Frond", "Grouse", "Grove", "Honey", "Log", "Lynx", "Magnolia",
     "Maple", "Moose", "Moss", "Moss", "Mushroom", "Oak", "Oak", "Pine", "Robin", "Robin", "Sap", "Skunk", "Sparrow", "Sprig",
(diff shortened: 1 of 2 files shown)


resources/dicts/names/names.json
scribblecrumb, line 1: Please remove the name changes from this PR!  If you'd like to make name changes, please do so in a separate PR.  Though fair warning, we're extremely picky about the warrior name options.

Yzzuwyzzuf: Sorry, I didn't realize my names were added to the pr, I thought only my thoughts dile was included. I'll remove those as soon as I can figure out how to add them to github without them being apart of a PR

Yzzuwyzzuf:
closed because the things in this pr were put in a more suitable pr placing these thoughts in the file they belong

PR #2647, opened 2024-08-21
More war events

j-gynn:
## About The Pull Request

Adds ~20 new war ShortEvents. These primarily cover positive changes to Clan relationship as we've been sorely lacking in those - but it does also add a few new negative ones too! Notably, kit-napping is now a thing, with the chance for kits to be adopted from another Clan or just straight-up abducted with no parent chosen. This also adds a new backstory for these stolen kits since nothing else really fit.

Should reduce the frequency of the error message about not being able to find war events.

## Proof of Testing
image

## Changelog/Credits
Pawsitive war events (sorry lmao)


resources/dicts/events/death/general.json
```diff
@@ -5634,385 +5923,414 @@
...
+        "event_id": "gen_death_war_predatorsaved",
+        "location": ["any"],
+        "season": ["any"],
+        "sub_type": ["war"],
+        "tags": ["no_body", "all_lives"],
+        "weight": 20,
+        "event_text": "m_c acts without thinking, saving the life of a o_c_n warrior by distracting a predator - which instead turns on m_c.",
+        "m_c": {
+            "status": ["warrior", "deputy", "leader", "medicine cat", "medicine cat apprentice", "apprentice"],
+            "age": ["adolescent", "young adult", "adult", "senior adult"],
```
scribblecrumb: in this case you could omit the age parameter.  the status parameter already constrains the cats to those statuses, which consequently will also constrain it to the ages those statuses are allowed to be.  notably, with the current set up, senior medicine cats and leaders would be incorrectly excluded.  you could just add senior to the age parameter to fix this, but i find it better to simply omit the age parameter to entirely avoid forgetting some edge case age.

please carry this suggestion on to the other events you've added in this PR

resources/dicts/events/death/general.json
```diff
@@ -5634,385 +5923,414 @@
...
+            {
+                "cats": [
+                    "m_c"
+                ],
+                "reg_death": "m_c was eaten by a predator.",
+                "lead_death": "eaten by a predator"
+            }
+        ],
+        "other_clan": {
+            "changed": 1
```
scribblecrumb: i think dying to save your enemy should have a larger impact, at least 3

resources/dicts/events/injury/general.json
```diff
@@ -7214,5 +7214,187 @@
...
+            "changed": 1
+        }
+    },
+    {
+        "event_id": "gen_misc_war_medcat_assist2",
+        "location": ["any"],
+        "season": ["any"],
+        "sub_type": ["war"],
+        "weight": 20,
+        "event_text": "Having heard that o_c_n was running low on vital emergency herbs, m_c gathers up some excess stock and makes the trek. However, with a mouth full of leaves, m_c takes a claw to the face before {PRONOUN/m_c/subject} can explain {PRONOUN/m_c/poss} intentions.",
```
scribblecrumb: while not totally necessary, it might be good to have this actually be based upon the clan's current herb supply.  i could see a potential scenario where the clan actually has no (or very low) herb and this becomes a contradictory event.  you could set it up to trigger on excess of a random herb, then "destroy" a small amount of that herb.  you could even then create some other similar events that are triggered by other herb supply states, such as one where the clan's supply actually *is* low, but they *still* give the herbs to their enemy and thus gain a larger positive relation increase.

resources/dicts/events/injury/mountainous.json
```diff
@@ -117,5 +117,40 @@
...
         ]
-    }    
+    },
+    {
+        "event_id": "mtn_injury_war_cliffsaved",
+        "location": ["mountainous"],
+        "season": ["any"],
+        "sub_type": ["war"],
+        "weight": 20,
+        "event_text": "m_c acts without thinking, yanking the o_c_n warrior {PRONOUN/m_c/subject}'d been fighting back from the edge of the cliff and saving their life.",
```
scribblecrumb: could do with a version of this event in the beach biome, featuring an ocean cliff and roiling waters below.

resources/dicts/events/injury/mountainous.json
```diff
@@ -117,5 +117,40 @@
...
+            "status": ["warrior", "deputy", "leader", "apprentice"],
+            "age": ["adolescent", "young adult", "adult", "senior adult"]
+        },
+        "injury": [
+            {
+                "cats": [
+                    "m_c"
+                ],
+                "injuries": [
+                    "battle_injury"
```
scribblecrumb: not sure this is the best injury to use here, as the event text doesn't actually imply direct injury to the cat.  i think it might actually be nice to give a lower stakes injury, as opposed to lots of the more serious ones we're giving, maybe the cat just wrenches a muscle in the struggle to pull the enemy cat back onto solid ground?  giving the "sore" condition?

resources/dicts/events/misc/general.json
```diff
@@ -1,3722 +1,3943 @@
...
+    },
+    {
+        "event_id": "gen_misc_war_medcat_woundedenemy_neg",
+        "location": ["any"],
+        "season": ["any"],
+        "sub_type": ["war"],
+        "weight": 20,
+        "event_text": "Whilst out collecting herbs, m_c stumbles upon a o_c_n warrior attempting to pack a wound with leaves. Their movements are slow and uncoordinated - without help, they'll soon be hunting with their ancestors. It would be so easy to just... walk past. But {PRONOUN/m_c/subject} can't. A few berries placed into their jaws with some soothing words... and that's one less o_c_n cat to deal with.",
+        "m_c": {
+            "status": ["medicine cat", "medicine cat apprentice"]
```
scribblecrumb: i think some trait constraints would be good both here and in the positive version of this event.  i.e. bloodthirsty, vengeful, righteous could kinda go both ways, sneaky, ect.

resources/dicts/events/misc/general.json
```diff
@@ -1,3722 +1,3943 @@
...
+        "event_id": "gen_misc_war_secret_friendship_apps_outspoken",
         "location": ["any"],
         "season": ["any"],
-        "sub_type": ["accessory"],
-        "weight": 10,
-        "event_text": "No matter how hard r_c tries, there's always something stuck in m_c's fur.",
-        "new_accessory": ["HERBS", "PETALS", "DRY HERBS", "DRY CATMINT", "DRY NETTLES", "DRY LAURELS"],
+        "sub_type": ["war"],
+        "weight": 20,
+        "event_text": "m_c is discovered gambolling along the border with an apprentice from o_c_n! Despite the reminder that the two Clans are far from allies, m_c loudly declares that {PRONOUN/m_c/subject}{VERB/m_c/'re/'s} not giving up {PRONOUN/m_c/poss} friend over some dumb war.",
```
scribblecrumb: a some_clan negative relationship change might be nice here, I imagine not everyone in the clan would appreciate this app's stance

resources/dicts/events/misc/general.json
```diff
@@ -1,3722 +1,3943 @@
...
         "season": ["any"],
-        "weight": 15,
-        "event_text": "m_c is scolded after sneaking out of camp.",
+        "sub_type": ["war"],
+        "weight": 20,
+        "event_text": "m_c is plagued by strange dreams where the leaders of o_c_n and c_n come together peacefully to trade fresh-kill.",
         "m_c": {
-            "age": ["kitten", "adolescent"]
+            "status": ["any"],
+            "age": ["kitten", "adolescent", "young adult", "adult", "senior adult", "senior"]
```
scribblecrumb: a skill constraint here for the dream and prophet skills would be nice, could even make it the lowest skill threshold so that even kittens can hit it

resources/dicts/events/misc/general.json
```diff
@@ -1,3722 +1,3943 @@
...
-    "event_id": "gen_misc_any_medicine_accessory_bug",
-    "location": ["any"],
-    "season": ["leaf-bare"],
-    "sub_type": ["accessory"],
-    "weight": 10,
-    "event_text": "m_c finds a stray acc_singular in its last moments of life leading them to a new herb gathering spot before {PRONOUN/m_c/subject} {VERB/m_c/have/has} never found before. It gives a final flutter before it gently decends to the ground. Is it a sign from star clan, or a coincidence? m_c doesnt know but {PRONOUN/m_c/subject} {VERB/m_c/decide/decides} to keep it with {PRONOUN/m_c/object} in its honor.",
-    "new_accessory": ["BLACK CICADA", "CICADA WINGS", "ROSY MOTH WINGS", "MORPHO BUTTERFLY", "MONARCH BUTTERFLY"],
-    "m_c": {
-      "age": ["adolescent", "young adult", "adult", "senior adult", "senior"],
-      "status": ["medicine cat"]
```
scribblecrumb: i know it's probably just formatter weirdness, but just confirm for me that there hasn't been a bunch of acc events deleted somehow

resources/dicts/events/new_cat/general.json
```diff
@@ -693,5 +693,116 @@
...
+                "mediator",
+                "mediator apprentice"
+            ]
+        },
+        "new_cat": [
+            [
+                "can_birth",
+                "age:has_kits",
+                "clancat",
+                "meeting",
```
scribblecrumb: hmm, i'm not sure how well this will actually work.. will it actually generate a clancat as an outsider and then handle it correctly? i'm not sure we have the functionality for that

scribblecrumb (changes requested):

resources/dicts/events/injury/forest.json
```diff
@@ -284,40 +284,5 @@
                 "injuries": [ "headache" ]
             }
         ]
-    },
-    {
```
scribblecrumb: was this intentionally removed? i don't think i saw it added to a different file

resources/dicts/events/new_cat/general.json
DeadGirlRisen: An idea to work around it, make it a loner/rogue with former clancat backstory, one of the ones of leaving the clan and not willing to talk about.

Then players can be like 'oh the cat left their clan to try and reclaim their kits' or some other story of why the parent left the clan in order to get back their kits. /idea

resources/dicts/events/injury/forest.json
j-gynn: Yeah - it got moved to a general injury :) No point in only being edible in a forest lol

scribblecrumb (approved):

PR #2657, opened 2024-08-25
Updated Time Skip Animation

PoppyBlossom:
## About The Pull Request
I updated the time skip animations to make it smoother and less aliased.

The animation looks smoother which is better I guess.

### Before:
ezgif com-animated-gif-maker (2)
### After:
Timeskip
May not look this smooth in real life applications.


j-gynn:
I think the aliasing/blockiness is intended. It's supposed to look like phases of the moon, and I feel like it loses that when it's smoothed out like this. It also seems to lose the weight and "bounce" the original had... Think Puss in Boots: The Last Wish and how they use animating on twos and threes for that oomph. idk, just my two cents.

PoppyBlossom:
> I think the aliasing/blockiness is intended. It's supposed to look like phases of the moon, and I feel like it loses that when it's smoothed out like this. It also seems to lose the weight and "bounce" the original had... Think Puss in Boots: The Last Wish and how they use animating on twos and threes for that oomph. idk, just my two cents.

It's a great point, I'll close this one for now.

PR #2711, opened 2024-09-20
[BUGFIX] Clicking and "cancelling" a timeskip plays the sound effect #2704

### LINKED ISSUE #2704: [BUG] Clicking and "cancelling" a timeskip plays the sound effect
**Type**: (only select one)
- Audio

**Describe the bug**
If you click and hold, drag away & release the click (as one would to cancel it), the sound effect still plays. may also happen with other buttons, not sure, just most noticeable on this

**Grade**: (only select one)

**Reproduce**
1. Click the Timeskip button
2. Hold down the mouse
3. Hear sound effect
4. Move mouse away from button & release click so that you "cancel" the click

PoppyBlossom:
## About The Pull Request
This PR addresses the issue where the time skip button plays the time skip sound effect when holding and moving the mouse away, even without time skipping. That problem is fixed by ~not letting the player hold that button at, all since almost all interactive buttons are not holdable, so might as well convert this one.~ making time skip sound effect play using UI_BUTTON_PRESSED, instead of UI_BUTTON_START_PRESS.

~I tried making the time skip sound effect play after the button had been actually pressed, but it was very confusing and I don't think it's worth it anyway.~

## Linked Issues
fixes: #2704

## Proof of Testing

https://github.com/user-attachments/assets/79073c83-48b3-43f4-97d3-41e13c4c66f2

### DIFF
--- a/scripts/game_structure/audio.py
+++ b/scripts/game_structure/audio.py
@@ -249,12 +249,17 @@ class _SoundManager:
         UIImageButtons have a sound_id parameter for assigning unique sounds to individual buttons
         :param event: the event that is taking place
         """
+        # This think make sounds play using UI_BUTTON_PRESSED, instead of UI_BUTTON_START_PRESS
+        try:
+            if event.ui_element.sound_id in ["timeskip"]:
+                if event.type == pygame_gui.UI_BUTTON_PRESSED:
+                    self.play("button_press", event.ui_element)
+                else:
+                    return
+        except:
+            pass
 
-        if (
-            event.type == pygame_gui.UI_BUTTON_START_PRESS
-            and event.ui_element.sound_id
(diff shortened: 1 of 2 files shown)


j-gynn:
Hmm, I think being able to cancel a timeskip is useful - it's a common feature in most software to allow a user to compensate for a misclick. Especially on a button that has no "undo" and no confirmation (not that either feature should necessarily be added for the timeskip, for the record, but worth pointing out)

PoppyBlossom:
I tried to make the time skip sound play correctly with the canceling feature, but I thought the finished code looked too ugly... Will add it back though.

PoppyBlossom:
Why are there 19 commits? Git is so weird.

PoppyBlossom:
Scroll bars are broken now.

scribblecrumb (approved):

PR #2723, opened 2024-09-23
next and previous cats now determined by class

### LINKED ISSUE #2671: [BUG] Next/Previous Cat doesn't go in proper order when viewing their sprites
Sorry if this has been posted already, couldn't find mention of it (even though I think it's been an issue for a long time)

**Type**: (only select one)
- UI 

**Describe the bug**
Clicking "Next Cat" or "Previous Cat" when viewing the sprites of the cats (the little magnifying glass button) seems to put them in a strange order and doesn't follow what you set the sort by order.

**Grade**: (only select one)
- Regular - A typical bug that should be seen too soon, but doesn’t necessarily break the game or inhibit the player

**Reproduce**
1. Click on a cat, then the magnifying glass button to view their sprite.
2. Click Next or Previous Cat

- d20d6765

**Additional context**
I can't find any sort of pattern that matches the order it goes in. The closest it seems to match to is Rank, which it follows for the first 9 cats, but at some point it just starts skipping around weirdly.

**Screenshots**
image
image

j-gynn:
## About The Pull Request

Fixes #2671 by bringing its outdated ordering code to parity with the other ordering code. Stores the function within Cat to centralise it. Also optimises the code itself since it was very slow for what it was doing.

## Linked Issues
Fixes #2671 

## Proof of Testing

https://github.com/user-attachments/assets/7ead79ce-4d6a-43f8-9d74-4bdc514d653c

## Changelog/Credits

Bugfix: cat order on sprite screen is consistent

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -593,11 +593,7 @@ class Cat:
         self.update_mentor()
 
         # if game.clan and game.clan.game_mode != 'classic' and not (self.outside or self.exiled) and body is not None:
-        if (
-            game.clan
-            and not self.outside
-            and not self.exiled
-        ):
+        if game.clan and not self.outside and not self.exiled:
             self.grief(body)
 
         if not self.outside:
@@ -722,9 +718,7 @@ class Cat:
                 text += " " + choice(MINOR_MAJOR_REACTION["major"])
                 text = event_text_adjust(Cat, text=text, main_cat=self, random_cat=cat)
 
-                cat.get_ill(
-                    "grief stricken", event_triggered=True, severity="major"
-                )
+                cat.get_ill("grief stricken", event_triggered=True, severity="major")
 
             # If major grief fails, but there are still very_high or high values,
             # it can fail to to minor grief. If they have a family relation, bypass the roll.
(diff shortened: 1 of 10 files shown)


scribblecrumb:
Seems like the chosen sorting order (rank, age, EXP, ect.) isn't preserved when viewing cat pages other than profile (such as relationships, family tree, choosing pages, ect.), instead it defaults to the rank sorting order.  Can't remember if it *was* preserved before, but either way it would be nice to see that fixed here.

j-gynn:
It always uses ID for this... I think it might cause problems to keep track of the other but I haven't looked at the codebase to confirm

j-gynn:
> It always uses ID for this... I think it might cause problems to keep track of the other but I haven't looked at the codebase to confirm


Update: this isn't true, the code only defaults to ID if the list screen hasn't been opened this session. Otherwise, it uses the same list that list screen uses, in the same order

scribblecrumb:
Ah, I see, the screens I was looking at were ones that you hadn't implemented the new system on to.  Could you implement this on the other screens with next/prev cat buttons?

j-gynn:
Which ones are missing? I replaced all the ones I could find but maybe I've been dumb and not done it right.

scribblecrumb:
Relationships screen, family tree, Choose mate/mentor/adoption

scribblecrumb (approved):

PR #2794, opened 2024-09-29
[CONTENT] Dad Jokes

tobes03:
## About The Pull Request

I've added some outcomes to common general/border and general/hunting patrols that are stat locked to childish/playful cats (and a few that are for strange cats as well). I'd love to add some more of these silly outcomes for other patrols.

Because dad jokes improve everything.

## Proof of Testing

Example of an outcome (this one's not a dad joke, it's for cats that are strange, playful, or childish)
image

Also not a dad joke (I admit the title was misleading): 
image

Here's one:
image

Here's another one (I was just excited about using the word "chortles"):
image

Here's several:
image

### DIFF
--- a/resources/game_config.json
+++ b/resources/game_config.json
@@ -250,7 +250,7 @@
 		"best_stat_modifier": 10,
 		"fail_stat_cat_modifier": -15,
 		"chance_of_romance_patrol": 16,
-		"debug_ensure_patrol_id": null,
+		"debug_ensure_patrol_id": "gen_hunt_fish2",
 		"debug_override_patrol_stat_requirements": false,
 		"debug_ensure_patrol_outcome": null,
 		"comment": [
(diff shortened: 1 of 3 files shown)


resources/game_config.json
PoppyBlossom, line 1: I don't think this is supposed to be here...

resources/game_config.json
```diff
@@ -250,7 +250,7 @@
 		"best_stat_modifier": 10,
 		"fail_stat_cat_modifier": -15,
 		"chance_of_romance_patrol": 16,
-		"debug_ensure_patrol_id": null,
+		"debug_ensure_patrol_id": "gen_hunt_fish2",
```
PoppyBlossom: I don't think this is supposed to be here...

scribblecrumb (approved):

PR #2900, opened 2024-10-15
[CONTENT] As You Like It-inspired Patrol

tobes03:
## About The Pull Request

Another patrol where two cats on a border patrol find two kittypets in an argument.

fun!

## Proof of Testing
image
image
image
image


resources/dicts/patrols/general/border.json
```diff
@@ -3001,5 +3001,522 @@
...
+                ["meeting", "kittypet", "age:young adult"],
+                ["meeting", "kittypet", "age:young adult"]
+                ],
+                "art": "gen_bord_twocat_hostile_loner_loner"
+        }
+	],
+
+	"antag_fail_outcomes": [
+		{
+            "text": "p_l doesn't care about kittypet drama. {PRONOUN/p_l/subject/CAP} {VERB/p_l/interrupt/interrupts} and {VERB/p_l/tell/tells} the kittypets to get lost. One kittypet doesn't appreciate p_l's tone and takes a swipe at {PRONOUN/p_l/object}. r_c jumps into help a humiliated p_l drive off the kittypets.",
```
scribblecrumb:
```suggestion
            "text": "p_l doesn't care about kittypet drama. {PRONOUN/p_l/subject/CAP} {VERB/p_l/interrupt/interrupts} and {VERB/p_l/tell/tells} the kittypets to get lost. One kittypet doesn't appreciate p_l's tone and takes a swipe at {PRONOUN/p_l/object}. r_c jumps in to help a humiliated p_l drive off the kittypets.",

```

scribblecrumb (changes requested):
loving this!  I think it might also be fun to add the "exists" tag to the new cats.  I like the idea of linking two cats that the player might already be aware of.

tobes03:
ooh fun!! I wasn't really sure how the "exists" tag worked but if it'd be okay here that would be great. I'll add it :)

scribblecrumb:
It should work here!  It'll only trigger if there's existing cats of the correct status and age

scribblecrumb (approved):

PR #2903, opened 2024-10-15
BUGFIX: Patrol debug_ensure not working properly

scribblecrumb:
## About The Pull Request
The debug_ensure was incorrectly pulling the ensured patrol from the unfiltered list of patrols instead of the filter list, leading to confusing results.

## Proof of Testing
image
Print message showing that an ensured patrol is now being properly constrained.  This is the patrol I was trying to debug when I discovered the debug_ensure wasn't working correctly.  Prior to my fix, all cats seemed to be able to access this deputy-specific patrol, even though the filters were working correctly.

### DIFF
--- a/scripts/patrol/patrol.py
+++ b/scripts/patrol/patrol.py
@@ -379,7 +379,7 @@ class Patrol:
         # This is a debug option. If the patrol_id set isn "debug_ensure_patrol" is possible,
         # make it the *only* possible patrol
         if isinstance(game.config["patrol_generation"]["debug_ensure_patrol_id"], str):
-            for _pat in possible_patrols:
+            for _pat in final_patrols:
                 if (
                     _pat.patrol_id
                     == game.config["patrol_generation"]["debug_ensure_patrol_id"]


Lixxis:
The thing is once I wanted to test a patrol for the outcome and was confused why the patrol couldn't get find because I assumed this debug patrol id works on all possible patrols without filter. Some patrol have tricky constraints which you don't easily can fulfill to check on one patrol I see that at some points you want to have the filtering when checking for an certain patrol and sometimes not.

I would suggest to make two types of id filtering: "debug_ensure" and "debug_ensured_filtered", so nobody else gets confused again.

scribblecrumb:
We actually already have a way to make the debug_ensure ignore filtering.  I think this was added while you were on hiatus?  Game config now has `debug_override_patrol_stat_requirements`, if you set it to `true` then filtering does not occur and the patrol will be pulled from all possible.  The fact that debug_ensure was doing this by default is now redundant and confusing if you're trying to test that a patrol *is* being constrained correctly.

Lixxis (approved):
Uh nice, just checked the whole file and saw the boolean you mentioned.

PR #2927, opened 2024-10-21
[BUGFIX] Inappropriate Romantic Like

FangWasHere:
## About The Pull Request

I found the issue that's been plaguing me the past week or so. A romantic event was in the platonic increase events file which was resulting in the occasional romantic interest between older cats and much much younger cats, so I moved it to romantic increase where it should be.

No inappropriate relationships!

## Proof of Testing

Event
Faded
Red
romancebar

## Changelog/Credits

FangoJett


scribblecrumb (approved):
Thanks for catching this!

PR #2946, opened 2024-10-23
[BUGFIX] Various Patrols Format Fixing

larkgz:
## About The Pull Request

Several patrols have slightly wrong formatting, causing them to not work as expected. This corrects them to have the correct formatting.  

More specifically:
* Relationship outcomes use "cats_to" and "cats_from", which are arrays
* Relationship outcomes (somewhat confusingly) use "jealous" instead of "jealousy"
* Relationship outcomes values are a list of strings, not a single comma separated string 
* Hunting patrols are "hunting" patrols, not "hunt" patrols
* Injury outcomes are arrays of objects that specify the injury
* To use existing cat for new_cat, should use "exists", not "existing"
* new_cat is specified by an array of arrays
* Very small prey is specified as "very_small", not "very small"
* Patrol art is a string, not a list of strings
* "can_have_stat" and "stat_trait" are arrays of strings. "stat_trait" refers to traits.
* Removes some `null` values since it's the same as if the property was not specified
* Fixes some improper JSON formatting that caused a patrol to be skipped over, or parse outcomes strangely

### DIFF
--- a/resources/dicts/patrols/mountainous/border/greenleaf.json
+++ b/resources/dicts/patrols/mountainous/border/greenleaf.json
@@ -399,20 +399,6 @@
                 ]
             },
             {
-                "text": "The strange scent leads to a stinky dogwood tree, newly in bloom... app1 huffs with embarrassment, suddenly grateful that {PRONOUN/app1/poss} patrol is so small. Fewer eyes watching that embarrassing moment. {PRONOUN/app1/subject/CAP} {VERB/app1/don't/doesn't} want to look dumb in front of other cats.",
-                "exp": 0,
-                "weight": 20,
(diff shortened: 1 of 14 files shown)


resources/dicts/patrols/mountainous/border/greenleaf.json
Lixxis, line 1: Why are the outcomes are removed? Maybe I miss something.

larkgz: They were removed because they're duplicated outcomes. The same outcome was in the same patrol multiple times.

Lixxis (approved):

PR #2956, opened 2024-10-25
Fixes torn pelt bug

### LINKED ISSUE #2938: [BUG] "torn pelt" related event crashes game on moonskip
**Type**: Code

**Describe the bug**
I moonskipped and the game just... crashed. Looking at the crash log (and with some help from @selkirks, thank you immensely) I think what's going on is that either a) the game is trying to generate an event that will give the "torn pelt" injury to a cat that already has it, or b) it's trying to give a cat a permanent condition off the torn pelt, and because such a condition doesn't exist, the game crashes. Either way, "torn pelt" is the problem because that's what the crash log is pointing to as the source.

**Grade**: Important

**Reproduce**
Honestly, no clue. It was a random crash - Sel and I just moonskipped until the game crashed. I'm guessing you need to have a cat with the "torn pelt" injury for it to happen though.

**Commit # or Game Version Number**: e65b5a53

**Screenshots**
Attaching the stderr;
[stderr_20241022_195132.log](https://github.com/user-attachments/files/17483980/stderr_20241022_195132.log)

j-gynn:
## About The Pull Request
Though technically this doesn't *fix* the problem (that's a writer's task!), it adds a catch to stop the game crashing and displays a potentially clunky event text message with a note to report it if seen. Boom.

## Linked Issues

Fixes #2938

### DIFF
--- a/scripts/events_module/condition_events.py
+++ b/scripts/events_module/condition_events.py
@@ -5,14 +5,14 @@ import ujson
 
 from scripts.cat.cats import Cat
 from scripts.cat.history import History
-from scripts.conditions import (
-    medical_cats_condition_fulfilled,
-    get_amount_cat_for_one_medic,
-)
 from scripts.clan_resources.freshkill import (
     FRESHKILL_ACTIVE,
     MAL_PERCENTAGE,
     STARV_PERCENTAGE,
+)
+from scripts.conditions import (
+    medical_cats_condition_fulfilled,
+    get_amount_cat_for_one_medic,
 )
 from scripts.event_class import Single_Event
 from scripts.events_module.handle_short_events import handle_short_events
@@ -22,7 +22,6 @@ from scripts.utility import (
     event_text_adjust,
     get_alive_status_cats,
     get_leader_life_notice,
-    get_random_moon_cat,
 )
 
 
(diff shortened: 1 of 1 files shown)


scripts/events_module/condition_events.py
```diff
@@ -689,16 +721,28 @@ def handle_already_injured(cat):
...
+                            ]
+                        )
+                    except KeyError:
+                        print(
+                            f"WARNING: No entry in gain_permanent_condition_strings for injury '{injury}' causing "
+                            f"condition '{condition_got}'. Using default."
+                        )
+                        possible_string_list = (
+                            f"After m_c's {injury} healed, {{PRONOUN/m_c/subject}} now {{VERB/m_c/have/has}} "
+                            f"{condition_got}. [Please report this if you see it!]",
```
scribblecrumb: `possible_string_list` needs to be a list, this looks like just a f string, unless I'm misunderstanding something?

scribblecrumb (changes requested):

scripts/events_module/condition_events.py
j-gynn: Ughhhh the IDE changed my bracketssss

scribblecrumb (approved):

PR #2971, opened 2024-11-01
New thoughts and thought constraints

Snowstar38:
## About The Pull Request

Added three new thought constraints for more thought variety:

- **clan_missing_status_constraint**: for detecting if the Clan is missing a leader/deputy/medcat

This is because the random_cat method for thoughts can only check if the random cat chosen has a given status. It can't actually detect if a status is absent from the Clan altogether.

- **herb_max_constraint**: For thoughts related to herb stores. I do plan to find a way to calculate the needed herb amounts for the Clan's population in the future.

- **leader_life_constraint**: For thoughts related to how many lives the leader has. I made it so it can take multiple values as a way to avoid needing both a max and a min constraint; let me know if you think it should be done differently!

It's always nice to have more thoughts directly related to the current state of the Clan. These will make the cats feel more connected to what's happening in the game.

## Proof of Testing

image
image
image

## Changelog/Credits

Added new thoughts:

- Thoughts for when the Clan has no leader, deputy, and/or medicine cat
- Thoughts about the herb stores
- Thoughts about the leader's remaining lives

### DIFF
--- a/resources/dicts/thoughts/alive/general.json
+++ b/resources/dicts/thoughts/alive/general.json
@@ -1038,12 +1038,20 @@
         ]
     },
     {
-        "id": "compassionate_warrior_to_lone_app_med_cat",
+        "id": "compassionate_to_lone_app_med_cat",
         "thoughts": [
             "Is helping r_c with {PRONOUN/r_c/poss} duties, hoping {PRONOUN/r_c/subject}{VERB/r_c/'re/'s} doing alright without a mentor"
         ],
         "random_status_constraint": [
             "medicine cat apprentice"
+        ],
+        "main_status_constraint": [
+            "leader",
+            "deputy",
+            "warrior",
+            "mediator",
+            "apprentice",
+            "mediator apprentice"
         ],
         "main_trait_constraint": [
             "compassionate"
--- a/resources/dicts/thoughts/alive/general.json
+++ b/resources/dicts/thoughts/alive/general.json
@@ -4630,7 +4630,7 @@
     {
         "id": "skilled_no_medcat_gen",
           "thoughts": [
-              "Wonders if they should become a medicine cat"
+              "Wonders if {PRONOUN/m_c/subject} should become a medicine cat"
           ],
           "main_skill_constraint": [
               "HEALER,1",
--- a/resources/dicts/thoughts/alive/general.json
+++ b/resources/dicts/thoughts/alive/general.json
@@ -1038,6 +1038,21 @@
         ]
     },
     {
+        "id": "compassionate_warrior_to_lone_app_med_cat",
+        "thoughts": [
+            "Is helping r_c with {PRONOUN/r_c/poss} duties, hoping {PRONOUN/r_c/subject}{VERB/r_c/'re/'s} doing alright without a mentor"
+        ],
+        "random_status_constraint": [
+            "medicine cat apprentice"
+        ],
+        "main_trait_constraint": [
+            "compassionate"
+        ],
+        "clan_missing_status_constraint": [
+            "medicine cat"
+        ]
+    },
+    {
         "id": "competitive_gen",
         "thoughts": [
             "Is challenging cats around camp to participate in some friendly competition",
@@ -4615,7 +4630,8 @@
               "Isn't sure the Clan can survive long without a medicine cat"
           ],
           "clan_missing_status_constraint": [
-              "medicine cat"
+              "medicine cat",
+              "medicine cat apprentice"
           ],
           "main_status_constraint": [
             "apprentice",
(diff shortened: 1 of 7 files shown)


scripts/cat/thoughts.py
```diff
@@ -94,7 +95,36 @@ def cats_fulfill_thought_constraints(main_cat, random_cat, thought, game_mode, b
                 return False
         elif 'random_status_constraint' in thought and not random_cat:
             pass
+        
+        # Constraints for empty positions of power in the Clan
+        # Only works for leader, deputy, and medicine cat
+        if 'clan_missing_status_constraint' in thought:
```
scribblecrumb, line 104: I'd like to see this be applicable to all ranks and I believe it would be relatively simple.  The PatrolEvent and ShortEvent formats already utilize a similar constraint system, however theirs is done via tags and then is filtered via the `get_alive_status_cats()` function in `utility.py`.  Plugging the list from `clan_missing_status_constraint` into `get_alive_status_cats()` would work well, since the function already takes a list of statuses as a kwarg.

It might additionally be wise to create a `clan_present_status_constraint` to handle the opposite situations: where you want to check that a certain status *is* present in the clan.

scripts/cat/thoughts.py
```diff
@@ -94,7 +95,36 @@ def cats_fulfill_thought_constraints(main_cat, random_cat, thought, game_mode, b
...
+                        print("FAIL", main_cat.name, missing_status, getattr(game.clan, missing_status))
+                        return False
+                    else:
+                        print("PASS", main_cat.name, missing_status, getattr(game.clan, missing_status))
 
+        # Constraints for size of herb store
+        if 'herb_max_constraint' in thought:
+            if game.clan:
+                herb_max = (thought['herb_max_constraint'])[0]
+                if herb_max < sum(game.clan.herbs.values()):
```
scribblecrumb: Regarding calculating how many herbs a clan needs, I'd direct you toward the calculations used in ShortEvent filtering.  ShortEvents allow constraining to set supply amounts of both certain herbs and overall herb supply: low, adequate, full, excess, as well as an "always" to trigger no matter how many herbs are in the supply.  I'll admit the filtering is a bit of a headache to look at, so if you want some help understanding it then let me know.

This is admittedly something I'm still trying to balance sufficiently.  If we're going to have it determined in multiple places, then there should be consideration towards either a function or some game_config settings so that balancing can remain consistent and easy to alter in the future.

resources/dicts/thoughts/alive/warrior.json
```diff
@@ -34,12 +34,36 @@
...
+            "Wants to be chosen as the new deputy"
+        ],
+        "clan_missing_status_constraint": [
+            "deputy"
+        ]
+    },
+    {
+        "id": "warrior_gen_no_deputy_ambitious",
+        "thoughts": [
+            "Desperately hopes r_c will choose them to be the new deputy"
```
scribblecrumb:
```suggestion
            "Desperately hopes r_c will choose {PRONOUN/m_c/object} to be the new deputy"
```

resources/dicts/thoughts/alive/leader.json
```diff
@@ -296,5 +296,38 @@
...
+            7,
+            8,
+            9
+        ]
+    },
+    {
+        "id": "leader_no_deputy",
+        "thoughts": [
+            "Is carefully mulling over who would be a good deputy",
+            "Organizes the patrols by {PRONOUN/m_c/object}"
```
scribblecrumb:
```suggestion
            "Organizes the patrols by {PRONOUN/m_c/self}"
```

resources/dicts/thoughts/alive/general.json
```diff
@@ -4607,5 +4607,89 @@
...
+            "mediator",
+            "deputy",
+            "leader",
+            "elder"
+        ]
+    },
+    {
+        "id": "skilled_no_medcat_gen",
+          "thoughts": [
+              "Wonders if they should become a medicine cat"
```
scribblecrumb:
```suggestion
              "Wonders if {PRONOUN/m_c/subject} should become a medicine cat"
```

scribblecrumb (changes requested):
Mostly feedback looking to make your additional constraints more similar to their event counterparts, for the sake of consistency.

scripts/cat/thoughts.py
Snowstar38: Wow, thank you! I had no idea that utility function existed. Now it works much more smoothly!

scripts/cat/thoughts.py
Snowstar38: I tried adding some utility functions based off of the values in the medicine cat concern string (where it shows what they think of their current stores), is that more like what you had in mind?

scribblecrumb: Yeah, very nice!  Could you also allow for checking the supply of a singular herb?

j-gynn:
Closing as stale.

PR #2978, opened 2024-11-03
[BUGFIX] Exiling Causes Name Change

### LINKED ISSUE #2936: [BUG] Apprentice in DF still Obtains Full Name
**Type**: (only select one)
- Code

**Describe the bug**
I had a med apprentice who murdered a warrior via death berries. They were exiled and eventually killed via the 'hunt down' button in the leader's den. The apprentice was 9 moons old when they died and went to the DF, but after reaching 12 moons (I'm assuming) they obtained their full name. This does not happen in SC.

**Grade**: (only select one)
- Non-vital 

**Reproduce**
1. Exile an apprentice (not sure if this is med-app specific)
2. Murder apprentice via 'hunt down'
3. Exile to DF
4. Watch for name change

c6e1d639fd551fe54ee9c7f232fda5a0afd29958

**Screenshots**

PoppyBlossom:
## About The Pull Request
This PR fixes the issue that if you exile a cat that doesn't have its full name yet, its name will change to its full name after exiling it and reloading the game.

## Linked Issues
Fixes: #2936

### Before:

https://github.com/user-attachments/assets/1e5977a4-e019-42c4-98d5-3de220e38157

### After:

https://github.com/user-attachments/assets/acc6e1f4-2f24-4f55-b0ea-7faab0304819


archanyhm:
Please split this up into two PRs; one with the bugfix and one with any unrelated refactorings. Mixing the two in a bugfix PR is kind of counterintuitive.

PoppyBlossom:
Okay.

scribblecrumb (changes requested):
image

Received this error upon loading the game.

scribblecrumb:
An issue I'm seeing is that exiled cats do now change their name as they age, even while still exiled.  The intention, and indeed how it worked when this bug did not occur, was for the cat to keep their name static until such time as they returned to the Clan and had a special ceremony to change their name.  I believe this worked prior because the actual *status* of the cat would not change, merely their outside or exiled bools would change.  And the status was not meant to be updated unless the cat was within the Clan, thus allowing them to keep whatever suffix they had when leaving.  Do you think this intention is possible to preserve?  If not, or if it would require too much work, I do believe this would be a problem solved by my planned changes to how outsider cat statuses function.  In which case, it would be alright for this not to follow the original intention atm.

PoppyBlossom:
> An issue I'm seeing is that exiled cats do now change their name as they age, even while still exiled. The intention, and indeed how it worked when this bug did not occur, was for the cat to keep their name static until such time as they returned to the Clan and had a special ceremony to change their name. I believe this worked prior because the actual _status_ of the cat would not change, merely their outside or exiled bools would change. And the status was not meant to be updated unless the cat was within the Clan, thus allowing them to keep whatever suffix they had when leaving. Do you think this intention is possible to preserve? If not, or if it would require too much work, I do believe this would be a problem solved by my planned changes to how outsider cat statuses function. In which case, it would be alright for this not to follow the original intention atm.

It is definitely possible to preserve the status of the cat, but it does mean that the status can't be overwritten by any other things. Implementing it should be relatively easy, but since a lot of the code has been using status as a general term rather than specific to the cat's "profession," it would take some time to separate them and replace them with flags.

scribblecrumb:
Alr, I'll go ahead and merge this fix for now then.  We can truly "fix" it to bring it back to intended function later lol

scribblecrumb (approved):

PR #3012, opened 2024-11-17
[CLEANUP] minor refactoring

filvas2:
## About The Pull Request

I had some spare time and was curious how the Clangen code worked. I then saw some code that I though might be useful to refactor because it was duplicated or similar

I hope it will help with code readability and therefore maintainability

## Proof of Testing
image
It still seems to run fine

## Changelog/Credits

Minor refactoring of cats.py, thought.py and skills.py


scripts/cat/cats.py
```diff
@@ -3106,13 +3088,12 @@ def mediate_relationship(mediator, cat1, cat2, allow_romantic, sabotage=False):
...
-        if effect > 0 and current_value + effect >= 100:
-            return 100
-
-        return current_value + effect
+        new_value = current_value + effect
+        if new_value < 0:
+            new_value = 0
+        elif new_value > 100:
+            new_value = 100
+        return new_value
```
archanyhm: We're basically clamping the value here, so you could simply do something like this:

```py
return max(0, min(current_value + effect, 100))
```

scripts/cat/skills.py
```diff
@@ -365,7 +365,7 @@ def generate_new_catskills(status, moons, hidden_skill: HiddenSkillEnum = None):
 
         new_skill.hidden = hidden_skill
 
-        # TODO: Make this nicer
+        # TODO: IS this nicer?
```
archanyhm: Nonsensical comment change. If you believe you've improved this section, I'd rather have you remove the comment altogether.

scripts/cat/thoughts.py
```diff
@@ -283,14 +278,15 @@ def load_thoughts(main_cat, other_cat, game_mode, biome, season, camp):
...
 
-        if not main_cat.dead and main_cat.outside:
+        if main_cat.dead:
+            if main_cat.outside:
+                spec_dir = "/unknownresidence"
+            elif main_cat.df:
+                spec_dir = "/darkforest"
+            else:
+                spec_dir = "/starclan"
+        elif main_cat.outside:
```
archanyhm, line 288: If we touch this code, I'd prefer if we used a path object instead, like `os.path`.

archanyhm (dismissed):
Not a complete review just yet, but a few things I noticed.

scripts/cat/cats.py
filvas2: It can indeed be compressed to 1 line, tenary operation is also an option.

But I think the if-else is more readable for a human, which is why I did it that way. If min-max is prefered I'll change it to that. Please confirm if that is what I should do.

scripts/cat/skills.py
filvas2: I was planning to once I got feedback about the change. It looks better in my opinion, but that doesn't mean it is.

Seemed more logical to add it as a code comment rather than doing a PR and then add it as a PR comment.

I also see that I accidentaly capitalized the S in "Is", which make it sound rude. Sorry for that

scripts/cat/thoughts.py
filvas2: I agree with you on that. But I did not feel it's my place to decide or make such design changes.

This is my first time contributing to open source and I don't want to overstep by doing more drastic changes

archanyhm: All good, I don't blame you.

scripts/cat/skills.py
archanyhm: I get the thought behind that, but I personally am used to comments like this making it into the codebase at some points in time and then being forgotten again.

scripts/cat/cats.py
archanyhm: As an alternative, I could suggest adding a helper function `clamp` in any appropriate location and using that instead; you could describe the purpose of the clamp-function with a docstring making it easy for people to comprehend if they're unsure what "clamp" means in this context and provide us with a neat utility to use down the line.

filvas2: I have now implemented a clamp in scripts->utility.py and call that. Therefore I will resolve this conversation.

scripts/cat/thoughts.py
filvas2: After contemplating implementation of os.path (or equivalent alternative), I have concluded that it would be more logical to do so in a PR dedicated to it. Doing so in only 1 location (or a few) only muddle the consistency of the code.

I will therefore mark this as resolved, as I still think the code's intent is easier to read.

scripts/cat/skills.py
filvas2: That is a fair point, and comment is now removed. I do not know if the code itself is an improvement, so I will let the conversation remain unsolved.

filvas2: As nothing more has been said. I will resolve this conversation

Lixxis (approved):

Lixxis:
@archanyhm is there still something which should be changed?

archanyhm:
> @archanyhm is there still something which should be changed?

Nope, should be fine to merge.

PR #3019, opened 2024-11-20
[CONTENT] biome-specific new cat events

arisaema12:
## About The Pull Request
adds two new_cat events for each biome

forest, mountains and plains new_cat event files were previously completely empty. new events help each biome to feel more unique

## Proof of Testing
new_cat event test

new cats generate correctly
new_cat family tree test

## Changelog/Credits
- adds eight new cat events


scribblecrumb (changes requested):
Love these, but some of them are very wordy.  With moon events, we generally want to keep them pretty short, one sentence is heavily preferred.  Ceremonies, Grief, and Birth are all somewhat exceptions to this rule.

Patrols are where we can get more wordy with things, but we want to avoid moon events feeling like a player is reading a novel/essay.  Think of it in terms of patrols being where you can get hyper specific about actions and feelings, which moon events are more of a vague overview of happenings.  For example, mtn_new_cat_eagle would be better condensed to "m_c found an injured loner while on a walk.  They seem to have been attacked by an eagle, m_c offers them treatment and a place in the Clan."

I would definitely encourage you to adapt the longer versions into patrols thought! Because they are wonderful and I think they would add great variety.

arisaema12:
For now, I’ve condensed the longest events - lmk if you think any of the others are also too long, though. I’ll make a separate pr in the future for patrol-versions of the two mountain events & kittypet lost in the tunnels event

scribblecrumb (approved):

PR #3063, opened 2024-11-29
[CLEANUP] Refactor ProfileScreen column generation

prettyflyfora:
## About The Pull Request

The methods that generate the two cat info columns on the profile screen currently use a string that is constantly replaced with the += operator. This request modifies these methods to collect the cat info in a list, then return it as a string joined with a new line. There are also some minor changes to conditional statements.

Lists are mutable and better for performance; changes to conditional statements reduce redundancy and improve readability.

## Proof of Testing

python_iPaOP45CYD
python_TMyLH4awVK
Untitled

## Changelog/Credits

Minor refactoring of cat profile column generation.


prettyflyfora:
Realized many functions in this file concat to an empty string, so this change would be pointless

Issue #2424, opened 2024-06-10
[CODE] Handle transing events as ShortEvents


scribblecrumb:
**Description**
At the moment, transing moon events are handled separate from ShortEvents.  It would be more consistent (and convenient) to handle them as ShortEvents, as well as provide us the opportunity to add more variations of the single event we currently have.

I would suggest adding transing events as their own misc event subtype ("transition"?) and then calling for them when a cat hits the chance to transition.

Broad Steps:
1. Transfer current transing event from being hardcoded in `events.py` to being formatted as a ShortEvent in the misc event dict json.  The wiki docs have instructions on how the ShortEvent formatting works.  A new subtype ("transition"?) would need to be used to differentiate these events from other misc events.
2. Probably add a new function to HandleShortEvents to handle the actual transitioning.
3. When a cat hits the change to transition, call for a trans events using the `handle_short_events.handle_event` function.

j-gynn:
Closing as resolved

PR #3075, opened 2024-11-30
[ENHANCEMENT] Change transing function to use ShortEvents

bakermor:
## About The Pull Request
Transferred hardcoded transing functionality from events.py to 2 misc ShortEvents with "transition" subtype. Adds 2 new attributes to ShortEvents: gender for m_c, and new_gender. 

Coming out is still rolled for each cat every moon. Chances for transitioning are based on age and are the same as before just moved to game_config.

Event text is the same as before but uses m_c's new pronouns.

It is more consistent to handle these events as ShortEvents and provides the opportunity to add more variations of the event.

## Linked Issues
#2424

## Proof of Testing
mtf:
image
image

ftm:
image
image

mtnb/ftnb:
image
image
image


Vickarooni:
i'm not a dev or anything, but i want to put my own two cents. putting trans chances in game_config might make it more accessible to transphobes since they could just... set the chances to 0. and i'm positive that's not something we _want_,,,

scribblecrumb:
> i'm not a dev or anything, but i want to put my own two cents. putting trans chances in game_config might make it more accessible to transphobes since they could just... set the chances to 0. and i'm positive that's not something we _want_,,,

I'll bring it up to the other senior devs.  However, I will say that it's hardly difficult to just go into the source code and do that anyways without it being in game_config.  Our stance has always been plainly about in-game toggles so I'm not sure how that stance will transfer to game_config settings.  We'll discuss it.

ZtheCorgi:
hopefully transphobes will be upset enough that we added pronouns to stay away regardless /hj

Vickarooni:
> I'll bring it up to the other senior devs. However, I will say that it's hardly difficult to just go into the source code and do that anyways without it being in game_config. Our stance has always been plainly about in-game toggles so I'm not sure how that stance will transfer to game_config settings. We'll discuss it.

that's true, but compared to source code, game_config is WAY more easily accessible, it's in the resources folder and is even available in a stable download .. and transphobes are just kind of Like That, i wouldn't be surprised if they would be willing to go through that tiny bit of extra effort just to disable a feature they don't like that wasn't as easily modifiable before.

scribblecrumb:
> > I'll bring it up to the other senior devs. However, I will say that it's hardly difficult to just go into the source code and do that anyways without it being in game_config. Our stance has always been plainly about in-game toggles so I'm not sure how that stance will transfer to game_config settings. We'll discuss it.
> 
> that's true, but compared to source code, game_config is WAY more easily accessible, it's in the resources folder and is even available in a stable download .. and transphobes are just kind of Like That, i wouldn't be surprised if they would be willing to go through that tiny bit of extra effort just to disable a feature they don't like that wasn't as easily modifiable before.

Senior devs have discussed and decided that we're just gonna add a catch that prevents transing chance from ever being zero (unless someone actively undoes the code we add).  That way it can be in game_config for those who want to modify the chance in good faith.

Vickarooni:
> Senior devs have discussed and decided that we're just gonna add a catch that prevents transing chance from ever being zero (unless someone actively undoes the code we add). That way it can be in game_config for those who want to modify the chance in good faith.

that's awesome! thank you so much for hearing me out ^^ tbh that was actually an idea that i had but didn't know how to bring it up nor would i know how to code it in (and if it was even plausible, for that matter) so i'm glad it crossed y'alls minds too. appreciate it, y'all r the best :)

scribblecrumb (approved):

PR #3091, opened 2024-12-02
[BUGFIX] added can_have_stat attribute to patrols that reference both p_l and s_c

### LINKED ISSUE #3090: [BUG] Cat referenced twice in patrol
### Type of bug

### Description

See attached screenshot

### Grade

Unknown

_No response_

0.12.1

### Additional context

_No response_

### Screenshots

Screenshot (913)

bakermor:
## About The Pull Request
Added the can_have_stat attribute to patrols that reference both p_l and s_c in the text. If multiple/all cats in the patrol have the stats (including the p_l), only the p_l is prioritized to be the stat cat. 

For all the patrols I could find that referenced p_l and s_c, and wouldn't make sense if s_c was p_l, the can_have_stat attribute was set to ["r_c"] or ["not_pl_rc"] to prevent the patrol leader from talking to/interacting with themselves.

## Linked Issues
Fixes #3090

### Before:
Since there was no can_have_stat attribute, if the p_l had the stat then it would always become the s_c. Only the p_l is in the possible_stat_cats list at the end of _get_stat_cat.
image

### After:
Patrol leader is no longer in possible_stat_cats
image
image


scribblecrumb:
Could I have you PR this to the v0.12.0 branch instead?  That way we can get the bugfix into the release version, then we'll just merge release into dev to get the bugfix there too.

bakermor:
Yeah! It should be to the right branch now

scribblecrumb (approved):

PR #3103, opened 2024-12-03
Fix error dialogue persisting between screens

### LINKED ISSUE #3040: [BUG] Buttons on main menu are clickable through error dialogue
**Type**: (only select one)
- UI - anything to do with the user interface. maybe a button is not working, or the way some assets are formatted seems to be off. 

**Describe the bug**

when an error occurs while loading the game, the buttons on the main menu are clickable *through* the error dialogue. 

also, the buttons on the error dialogue stop working after clicking into the settings screen and then back, but this is not as important if you can't switch screens in the first place.

expected behavior: you can't press the buttons on the main menuwhen the error dialogue is open

**Grade**: (only select one)

**Reproduce**
1. intentionally cause a json error in save files
2. click on the settings button (but do *not* close out of the error dialogue)
3. return to main menu
4. click on the close button for the error dialogue

**Patrol ID + Outcome Print (if applicable)**: N/A

**Commit # or Game Version Number**: first observed on [stable v0.11.2](https://github.com/ClanGenOfficial/clangen/releases/tag/v0.11.2), reproduced on https://github.com/ClanGenOfficial/clangen/commit/671177d1f5f7f6e2596c0276d227b560a372e82e

**Additional context**

**Screenshots**

https://github.com/user-attachments/assets/e920145f-22ac-4c16-b41f-18ac042267c6

Treefire33:
## About The Pull Request
Fixes the error dialogue from persisting between screens.

## Linked Issues
fixes: #3040 

## Proof of Testing
https://github.com/user-attachments/assets/2fbfd244-e153-4aaa-bb26-fb85d318c1ce

### Other
If this isn't the best UX-wise, I do have another solution in mind.

### DIFF
--- a/scripts/screens/StartScreen.py
+++ b/scripts/screens/StartScreen.py
@@ -55,6 +55,8 @@ class StartScreen(Screens):
 
         self.social_buttons = {}
 
+        self.error_open = False
+
     def handle_event(self, event):
         """This is where events that occur on this page are handled.
         For the pygame_gui rewrite, button presses are also handled here."""
@@ -73,7 +75,7 @@ class StartScreen(Screens):
                 self.new_clan_button: "make clan screen",
                 self.settings_button: "settings screen",
             }
-            if event.ui_element in screens:
+            if event.ui_element in screens and not self.error_open:
                 self.change_screen(screens[event.ui_element])
             elif event.ui_element == self.open_data_directory_button:
                 if platform.system() == "Darwin":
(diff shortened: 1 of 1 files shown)


scribblecrumb:
A useful fix, however it would be better to just disable other buttons on the screen while the window is open.

archanyhm:
Also targets the wrong branch, since the bug seems to affect release as well.

scribblecrumb:
Changes look good, but it seems that switching the branch pulled a bunch of dev changes over too (transing and other clan save data, specifically)

PR #3130, opened 2024-12-06
[CLEANUP] Streamline generate_events

scribblecrumb:
## About The Pull Request
This is dependent on #3119 since it's based on that PR's file directory changes.

This PR greatly streamlines `generate_events.py` by utilizing modular functions for the filtering of ShortEvents (and leader den events).  The new `event_filters.py` script holds said functions.  For now, I have not spread the use of these filters any farther than `generate_events.py`, but they already make that code significantly easier to read.

I made no change to the parameters of the filters except for the herb_supply filter, where I made some adjustments to how many herbs it perceives as "full" as I had changed my mind on what that should be since I first wrote the herb supply filter.  Otherwise, filter parameters were kept as is to keep the scope of this PR lower.  And frankly, herb stuff is just desperate for a herb class ala freshkill pile class.  I've added it to my list of things to work on bc I'm Tired of this being a continued issue lol

My hope is to integrate these filters into other clangen text events, such as thoughts, patrols, and relationship events.  This should help us bring all of these varying event formats closer in parity with each  other.  And by separating each individual filter into functions (`event_for_season`, `event_for_location`, ect.) it becomes very easy to plug in specific filters where they are needed, while leaving off any undesired filters.   

Consistency within how we format different types of events is helpful for our writers and continuing to bring these systems into parity with each other is something we should aspire to.

## Proof of Testing
image
Here is proof of short events and leaders den events still running without a crash.  Beyond just this screenshot, I did some pretty extensive testing via a debugger, so that I could follow along the code as it ran and find any issues.  I caught and fixed quite a few bugs that way and I'm confident that no more persist.

## Changelog/Credits
- Created filter functions to be held within `event_filters.py`
- Utilized those functions to clean up ShortEvent and leader den event filtering within `generate_events.py`
- Slightly reduced requirements for an herb supply to be considered "full"


scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,403 @@
...
+def _check_cat_trait(cat, traits: list, not_traits: list) -> bool:
+    """
+    checks if cat has the correct traits for traits and not_traits lists
+    """
+    if not traits and not not_traits:
+        return True
+
+    cat_trait = cat.personality.trait
+
+    if cat_trait in traits and cat_trait not in not_traits:
```
j-gynn: What happens if `not_traits` contains a value but `traits` is empty?

`if cat_trait in []` will always return false, so I think that would stop anything from generating?

The same concern applies across all the sub-queries that use this format

scripts/events_module/generate_events.py
```diff
@@ -222,16 +230,16 @@ def possible_short_events(event_type=None):
 
         return event_list
 
-    @staticmethod
```
j-gynn, line 225: Why was this changed from a static method?

j-gynn (changes requested):

scripts/events_module/generate_events.py
scribblecrumb: Good question, i have no idea xD I'll put that back

scripts/events_module/event_filters.py
scribblecrumb: Changed how it was set up, let me know if it's acceptable now.  I also don't think any other filters had the same issue? Skill/not_skill was already set up differently.

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+            req_biome = place
+            req_camps = ["any"]
+
+        if req_biome == game.clan.biome.lower():
+            if "any" in req_camps or game.clan.camp_bg in req_camps:
+                return True
+            else:
+                return False
+            
+        return False
```
j-gynn:
```suggestion
        return False
```
Can just shorten it to this, since it doesn't matter where it fails the check, just that it fails it.

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def event_for_season(seasons: list) -> bool:
+    """
+        checks if the clan is within the given seasons
+        """
+    if "any" in seasons:
+        return True
+    elif game.clan.current_season.lower() in seasons:
```
j-gynn:
```suggestion
    if "any" in seasons or game.clan.current_season.lower() in seasons:
```
Python is lazy in its evaluation: if the first part of an `or` is true, it doesn't bother evaluating the second and just proceeds. So you can make this shorter again!

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+                        cat,
+                        ["apprentice", "medicine cat apprentice", "mediator apprentice"]):
+                    return False
+                else:
+                    continue
+
+            if rank in ["leader", "deputy"] and not get_alive_status_cats(cat, [rank]):
+                return False
+
+            elif not len(get_alive_status_cats(cat, [rank])) >= 2:
```
j-gynn:
```suggestion
            if not len(get_alive_status_cats(cat, [rank])) >= 2:
```
No way to get here without `continue` or `return`, so it's okay to just be an `if`

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+        elif "full" in trigger and len(
+                [x for x in herb_supply if needed_amount < herb_supply[x] <= needed_amount * 2]) == num_of_herbs:
+            return True
+        elif "excess" in trigger and len(
+                [x for x in herb_supply if needed_amount * 2 < herb_supply[x]]) == num_of_herbs:
+            return True
+
+        return False
+
+    elif supply_type == "any_herb":
```
j-gynn:
```suggestion
    if supply_type == "any_herb":
```
Same as before

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+            print(f"WARNING: possible typo in supply constraint: {chosen_herb}")
+            return False
+        if "low" in trigger and herb_supply[chosen_herb] < half_amount:
+            return True
+        elif "adequate" in trigger and half_amount < herb_supply[chosen_herb] <= needed_amount:
+            return True
+        elif "full" in trigger and needed_amount < herb_supply[chosen_herb] <= needed_amount * 2:
+            return True
+        elif "excess" in trigger and needed_amount * 2 < herb_supply[chosen_herb]:
+            return True
```
j-gynn, line 247: This whole block doesn't need to be in an `else` statement (I'm not brave enough to remove the indentations in a suggestion so I leave that as an exercise for the reader ;) )

scripts/events_module/event_filters.py
```diff
@@ -0,0 +1,412 @@
...
+            allowed = True
+        else:
+            return False
+    if not_traits:
+        if cat_trait in not_traits:
+            allowed = False
+        else:
+            return True
+
+    return allowed
```
j-gynn:
```suggestion
    if traits and cat_trait not in traits:
        return False
    if not_traits and cat_trait in not_traits:
        return False
    return True
```
I will kill all your interim variables if it kills me /silly

scripts/events_module/short/handle_short_events.py
```diff
@@ -11,7 +11,7 @@
     FRESHKILL_EVENT_TRIGGER_FACTOR,
 )
 from scripts.event_class import Single_Event
-from scripts.events_module.generate_events import GenerateEvents
+from scripts.events_module.generate_events import GenerateEvents, generate_events
```
```suggestion
from scripts.events_module.generate_events import GenerateEvents
```

scripts/events_module/short/handle_short_events.py
```diff
@@ -127,7 +127,7 @@ def handle_event(
             event_type = "injury"
         possible_short_events = GenerateEvents.possible_short_events(event_type)
 
-        final_events = GenerateEvents.filter_possible_short_events(
+        final_events = generate_events.filter_possible_short_events(
```
```suggestion
        final_events = GenerateEvents.filter_possible_short_events(
```

scripts/events_module/generate_events.py
```diff
@@ -949,88 +536,30 @@ def possible_lead_den_events(
...
+                    cat=cat
+            ):
+                continue
 
             possible_events.append(event)
 
         return possible_events
 
 
 generate_events = GenerateEvents()
```
j-gynn, line 565: May be able to delete this line, depends if it's used elsewhere.

j-gynn (commented):
You asked for 'em. Enjoy snacking! There are other things you could technically do but like, at some point you start to impact readability... and in a codebase running *Python*, optimisation is not the priority :wink:

scripts/events_module/generate_events.py
scribblecrumb: I'd have to do more finagling for that bit, since there's some leader den specific logic in here

j-gynn (approved):

scripts/events_module/generate_events.py
j-gynn: Sounds good, maybe make an enhancement post for it? Or just keep it in your mind palace lmao

scribblecrumb: Yeah I'll make a post for it, else I shall def forget (and if i'm lucky, rusty or someone else will do it xD)

PR #3162, opened 2024-12-12
[ENHANCEMENT] Localization Support

j-gynn:
## About The Pull Request

Adds support to localize ClanGen. It's possible, by omitting files, to create a partial translation that uses English as a fallback. This allows for players to localize parts of the UI (or the whole thing) without having to create a whole bunch of events/patrols/thoughts/leader ceremonies/whatevers just to get the game rolling.

It has been explicitly designed so that content in other languages is unrelated to English. In other words, Patrol X in French might not exist in English at all. This is so that content does not have to simply be translations, but can be its own creative ventures in the target languages.

**Pronoun Updates & Changes to Saves**
The way that pronouns are saved in a cat has been changed. Anything that was formerly stored there is now stored in a dictionary key with the "en" value - in other words, as English pronouns. This allows for cats to have different pronouns in different languages.

On that subject, default pronouns have been moved out of `cats.py` and into `resources/lang/en/pronouns.json`. This technically means that players can now modify the default pronouns for cats in English.

**LIMITATIONS**
- If a content file is present but sparse, there is no way to fallback to English. In other words, if there are 2 patrols in Greek for hunting, that's all you're ever going to be getting for hunting. This one is probably the most egregious issue... but I'm not sure how to solve it without the English content overriding the target language too often.
- Scars and death history will be saved in the language they were created in. No way to change that unless we explicitly require every single patrol & event to be translated into every single language before it's allowed to be released which... do not recommend. Also would need a rework of how that stuff is saved so just Nope, No Thanks.
- Thoughts do not regenerate when you swap languages.
- Condition reveals will occur in the language that the cat was generated in. In other words, if a kit is born in French and the game swapped to English, the reveal of their weak paw will be _en Français_.
- There's some fun stuff that happens if you end up using the fallback language (english) but have the target language partially loaded. Especially pronouns. It's not bugs, but it looks... weird.

Finally make the languages buttons on the settings screen do something ;)

## Proof of Testing
Much testing will be required. Ideally, would love to ship with at least 1 additional language in the UI. For now, we run potatotests.

Updated create pronouns screen with grammatical gender dropdown
image

## Changelog/Credits
Localization support


j-gynn:
I've discovered more hardcoded stuff in patrols. back to draft!

archanyhm:
Looks good so far, but please resolve the merge conflicts here.

j-gynn:
@archanyhm conflicts resolved!

archanyhm (approved):

PR #3235, opened 2024-12-23
Logging

j-gynn:
## About The Pull Request

Adds logging support through much of ClanGen, replacing the `print()` statements that would be uncommented to investigate things. Works the same way as logging in most any other library. Default logging level is set to INFO.

Also adds a command `logging` to the debug console, enabling control of logging levels whilst in-game. Honestly, wouldn't suggest using it as you're probably more interested in the overall flow rather than one or two modules (something like patrols is in 10+ modules).

Logging helps us figure out what's going on.

## Proof of Testing
image


j-gynn:
Closing as stale.

PR #3257, opened 2024-12-24
[ENHANCEMENT] add `update_previous_next_cat_buttons` utility function

rustykitty:
## About The Pull Request
I added a utility function `update_previous_next_cat_buttons`. This function's purpose is to enable/disable the "next cat" and "previous cat" buttons on a screen based on returned values from the determine_next_and_previous_cats() function.
I also updated a few screens to use this function.

identical code in multiple places is hard to maintain

## Proof of Testing

https://github.com/user-attachments/assets/df148175-e44b-4184-be0f-bbc41fd6f4b0


j-gynn (changes requested):
Could this function be moved to `Screens` please?

scripts/screens/Screens.py
```diff
@@ -906,6 +906,12 @@ def theme(self) -> str:
                 _settings = ujson.loads(read_file.read())
                 return "dark" if _settings["dark mode"] else "light"
 
+    def update_previous_next_cat_buttons(self) -> None:
+        """Updates disabled status of previous and next cat buttons."""
+        # pylint: disable=no-member
+        self.previous_cat_button.enable() if self.previous_cat else self.previous_cat_button.disable()
+        self.next_cat_button.enable() if self.next_cat else self.next_cat_button.enable()
```
j-gynn:
```suggestion
        self.previous_cat_button.enable() if getattr(self, "previous_cat") else self.previous_cat_button.disable()
        self.next_cat_button.enable() if getattr(self, "next_cat") else self.next_cat_button.disable()
```
Just so it won't throw a crash if it's called on a screen that doesn't have the `previous_cat` and `next_cat` attributes. Also so that we actually do disable the `next_cat` button if it's invalid.

scripts/screens/screens_core/screens_core.py
```diff
@@ -68,7 +68,7 @@ def rebuild_core(*, should_rebuild_bgs=True):
     )
     menu_buttons["catlist_screen"] = UISurfaceImageButton(
         ui_scale(pygame.Rect((0, 60), (88, 30))),
-        "screens.core.cat_list",
+        "screens.core.catlist",
```
j-gynn: Revert, this will display incorrectly.
```suggestion
        "screens.core.cat_list",
```

j-gynn (changes requested):
See comments. Please re-request my review when you've made changes otherwise I won't know!

scripts/screens/Screens.py
rustykitty: Pretty sure you meant hasattr and not getattr, getattr throws an error when the attribute doesn't exist.
Also, I've changed it to `hasattr(self, "previous_cat") and self.previous_cat` since short-circuiting means that it won't throw an error (I believe)

j-gynn (approved):

PR #3263, opened 2024-12-26
[ENHANCEMENT] Switch Clans Without Exiting to Desktop

### LINKED ISSUE #3067: [CODE] Add a way to switch Clans without closing game
From the user perspective, I think that it would be much better to be able to switch Clans *without* closing the game, just as, for example, you can navigate to different websites on a web browser in the same tab. An idea I have for implementing this is to code in a way to unload the current Clan, so then all that is needed to do is to unload the current Clan and load in the new Clan.

I saw a few messages about it from Key in the Discord from about 1.5 years ago, but they don't seem to have gone in any direction.
- https://discord.com/channels/1003759225522110524/1005586496142704800/1084991545427636234
- https://discord.com/channels/1003759225522110524/1108406775000084643/1113136270877671554

Thanks!

Treefire33:
## About The Pull Request
Implements a way for the game to switch clans without having to close the game.
Also adds a way to "hot reload" a clan with F5.

## Linked Issues
closes: #3067 

## Proof of Testing
https://github.com/user-attachments/assets/c056f106-7e80-4658-84c2-63dccb8f5ff2

(Note: proof of testing shows some bugs, but they aren't detrimental to the game running.)


j-gynn:
Was saving always this non-performant for you (looking at the video, this seems like an abnormally long save time). It also seems like the game is being hard-reset every time, instead of only reloading the relevant data (the libpng warning is displayed again, which is probably contributing to the slow load times).

I am personally a little concerned with having a debug tool that could potentially lose players save data accessible from a simple mispress. I don't know whether a kind of "sudo" mode could be implemented for dev tools so that regular players can't accidentally trigger game-wrecking features (perhaps in the same vein as `show_bounds` and co. that must be toggled using the debug console or via changing a section of the code).

I'm super stoked to see this feature added & I know that players are gonna love it. It's a huge QoL that's been on my hit list for months so I'm glad someone's finally taking a stab at it.

Treefire33:
Okay, seems that the two issues listed might have been caused by the inclusion of loading all the spritesheets again.

Treefire33:
I added a command, so here's some testing for it:

https://github.com/user-attachments/assets/790e1d8d-626f-4e87-9fcd-4391f1e3a3ae

j-gynn (approved):

PR #3270, opened 2024-12-26
Fix events to display correctly post-localization

### LINKED ISSUE #3268: [BUG] Update `Single_Event` to support automatic `event_text_adjust` using `cats_involved`
**Description**
As title. `Single_Event` currently takes a list of IDs in `cats_involved`, but it would be good to extend this to be similar to the input for `event_text_adjust` if at all possible. It would mean that events can be automatically pronounified and save me about fifteen headaches with localization.

I'd suggest making an optional argument for `Single_Event` as a dict. If `cats_involved` is not specified, take all the values from this new optional dict and use their IDs.

j-gynn:
## About The Pull Request
Another localization fix, this time modifying events to track who is what in them so they can be rendered correctly. Also fixes a bug with the illness string displaying incorrectly because apparently I can't spell.

## Linked Issues
Fixes #3268

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -48,6 +48,7 @@ from scripts.game_structure.localization import load_lang_resource
 
 import scripts.game_structure.localization as pronouns
 
+
 class Cat:
     """The cat class."""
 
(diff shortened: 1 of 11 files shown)


Tiritirimatangi (approved):
we should really refactor the death texts for lost cats, but that's a mission for another day. Approved lol!

PR #3295, opened 2024-12-30
Updated incorrect translation strings

### LINKED ISSUE #3294: [BUG] Injury text does not display correctly
### Type of bug

### Description

This does not happen with illnesses to my knowledge, but in cat profiles or in the medicine cat den, injuries always have "injuries." at the beginning (localization error?) so for example, a mangled tail looks like "injuries.mangled tail".

### Grade

Regular - A typical bug that should be seen to soon, but doesn't necessarily break the game or inhibit the player

1. Find a cat that is recovering from birth
2. Check their profile
or, alternatively -
3. Check medicine cat den
4. Hover over an injured cat

_No response_

8e1979e8

### Additional context

I believe the injury text is only broken in profiles when it's specifically "recovering from birth" - I did not see it happening with any other injury.

### Screenshots

There are... a LOT more.
image
image
image
image
image
image

j-gynn:
## About The Pull Request

Fixes a few instances pointing to the wrong location for an i18n string. Also fixes getting a runny nose or stomachache string

## Linked Issues
Fixes #3294 

## Proof of Testing
image
image
image
image

### DIFF
--- a/scripts/events_module/short/condition_events.py
+++ b/scripts/events_module/short/condition_events.py
@@ -305,15 +305,16 @@ class Condition_Events:
                 # make em sick
                 cat.get_ill(chosen_illness)
 
+                # try to translate the illness
+                illness = i18n.t(f"conditions.illnesses.{chosen_illness}")
+
                 # create event text
                 if i18n.config.get("locale") == "en" and chosen_illness in [
                     "running nose",
                     "stomachache",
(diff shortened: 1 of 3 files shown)


Tiritirimatangi (approved):
fingers crossed this fixes our strings, but I don't think it'll crash the game regardless

PR #3339, opened 2025-01-04
[ENHANCEMENT] Basic localization with automatic GoogleTranslate translations

filvas2:
## About The Pull Request

Added a script in resources\lang that use a GogleTranslate python module to automatically translate relevant texts after a language folder is made by create_new_lang_files.py.

Also include partially human translated Swedish localization

I do not think GoogleTranslate is a substitute for a proper translation, but ClanGen is a very text heavy game, which make the task of localization is quite daunting. I believe making translation as easy as possible is required if localization is ever going to be feasible. 

The reason I think so is because I wanted to make a Swedish localization and realized that even if I did 2 files every day it would take me half a year. Some of the files are also several hundred lines and the game is still growing.  

By using GoogleTranslate as a base, it cut down the manual translation time significantly, as the translation will not be needed to be done from scratch.

Generating basic translations will also be useful to see if/where the GUI fall short in different languages than english

## Proof of Testing

image
Swedish ClanGen is cursed!

## Changelog/Credits

Ability to make basic automatic GoogleTranlsate localization translations to build proper translations on


archanyhm:
First, scripts like this shouldn’t go in resources. `bin` is probably a more suitable candidate for something like this.

second, I’m not a huge fan of machine made translations. The quality of them depends on the source and target language, but often times complex texts end up sounding unnatural at least and nonsensical at worst. I myself am very much against this approach, especially considering that translations are more of a community effort rather than a mammoth task for a single person. So imo an announcement that we’re looking for translations/translators is more optimal.

filvas2:
@archanyhm 
I will modify and move the script to `bin` if the PR is not denied

I also agree with most of your second point. All translations should be approved be a fluent speaker of a the language. Machine made translation, at its current performance level, has no place in a proper translation and it does a lot of objectively bad takes. For example, it keeps translating "pants" to "byxor", which is the Swedish word for trousers. ClanGen is also too big for a single person to translate and should be done through collaboration. As a swede, I can say the Swedish machine translation is very bad in many places and the Swedish translation included in this PR as a proof of concept should be removed if merge is considered. 

What I do disagree with is not using machine translation as an aid, just because it can be misused. For example the file `resources\lang\en\conditions\healed_and_death_strings\injury_death_strings.json` mostly consist of "cat died from X" sentences. To me there is no value in someone translating straightforward parts such as "cat died from" over and over. By letting machine translation do it, half the work is done before it starts.

For me personally, I also find it faster to turn a mangled translation into a decent one, rather than translating completely from scratch most of the time. If the machine translation is too illegible, I can always fall back to the original text with very little time loss.

I also absolutely do not mind if you or someone else just close the PR if this don't seem suitable for ClanGen, so don't feel any need to keep the PR open to avoid drama with me or out of politeness. This goes for any PRs I do, although it's nice to get told why a PR is not suitable, so I can keep that in mind in the future.

Vickarooni:
gonna drop my two cents here and agree with arch. i too dislike automatically translating the entire game. google translate in specific is also a very bad translator from my experience - especially with my native language, spanish, as there's a lot of words that are spelled the same but mean different things, and the translator often mixes them up. it rarely takes context into account as well. there's also what arch said about the overcomplication of sentences that end up looking weird and may confuse translators more than helping them. 

i've been trying my hand at translating clangen to spanish myself (the basic things since i'm having some issues with gendered words and the naming system) and i've noticed a huge part of translating the game is keeping things brief, something an automatic machine translator obviously wouldn't take into account

anywayz!! just wanted to say what i think :) i very much like the idea of asking for translators in the discord server, i'd love to help with a spanish localization!

(edit: typos, clarifying some things)

filvas2:
As there was no interest in this I close the PR

PR #3358, opened 2025-01-13
[CODE] refactor medical_cats_condition_fulfilled

rustykitty:
## About The Pull Request
Split `conditions.medical_cats_condition_fulfilled` into two functions, `amount_clanmembers_covered` and `medicine_cats_condition_fulfilled`. Updated usages accordingly. 

IMO, it's considered bad practice for a function to be able to return two different things based on arguments it is passed, let alone be capable of returning values of two different types. 
In fact, the following Java example will not even compile, because `boolean` is not implicitly convertible to `int` in Java:

```java
class Test {
    public int test() {
        return true;
    }
}
```

## Linked Issues
N/A

## Proof of Testing
Screenshot 2025-01-12 at 6 41 26 PM

### DIFF
--- a/scripts/conditions.py
+++ b/scripts/conditions.py
@@ -13,23 +13,23 @@ from scripts.game_structure.game_essentials import game
 
 
 
-def amount_clanmembers_covered(all_cats, amount_per_med):
+def amount_clanmembers_covered(all_cats, amount_per_med) -> int:
     """
     number of clan members the meds can treat
     """
 
-    medical_cats = [
+    medicine_cats = [
         i for i in all_cats
         if not i.dead and not i.outside and
            not i.not_working() and i.status in
                     ["medicine cat",
                      "medicine cat apprentice"]
     ]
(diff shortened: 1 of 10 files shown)


scripts/conditions.py
```diff
@@ -47,22 +42,19 @@ def medical_cats_condition_fulfilled(all_cats,
...
             total_med_number += 1
-        
-    
+
     adjust_med_number = total_med_number + total_exp
 
-    can_care_for = int(adjust_med_number * (amount_per_med + 1))
+    return int(adjust_med_number * (amount_per_med + 1)) # number of cats they can care for
 
+def medicine_cats_condition_fulfilled(all_cats, amount_per_med):
```
j-gynn: Since we're here, could this be renamed to something that more clearly explains what "condition" is being fulfilled? Perhaps something like `clan_has_enough_medcats`? This gives the subconscious implication that it's returning a bool and tells you what the "condition" is (improving readability).

j-gynn (approved):

PR #3365, opened 2025-01-17
[HOTFIX] No audio driver? No problem!

### LINKED ISSUE #3389: [BUG] Crashes if no audio driver found
### Type of bug

### Description

"Traceback (most recent call last):
  File "main.py", line 280, in <module>
pygame.error: WASAPI can't find requested audio endpoint: Element not found."

I get this error now anytime I launch my game, I tried to reinstall but it isn't working. Latest update.
I was playing the game earlier, and I never tampered with the game's files. I don't know what happened during that time and now.

### Grade

Game-breaking - a bug that makes it impossible to play the game

Unsure, since all I did was launch the game a little later on after playing it earlier today. Seems like a major file is missing? But it won't repair itself despite the reinstall. Could also be firewall tampering, try that?

_No response_

0.12.2

### Additional context

_No response_

### Screenshots

Image
Image

j-gynn:
## About The Pull Request

Fixes the game so that it can be played without a functional audio driver. Before, the check that was meant to safeguard against this error was after the first use of `pygame.mixer`, so the game would crash before it could do anything.

This was mocked up by disabling all my audio endpoints.
image
During testing, it was discovered that load times were significantly bloated without an audio driver (10+ seconds where it was under a second with one). I think that pygame must attempt init multiple times before it gives up, causing a very long delay for no reason. To fix this, `pygame.init()` and a few helper functions were moved from `Game` into `main.py`. This let the function be called after the audio system had a chance to `init` a dummy, thus skipping attempting initialization & making the game load at a reasonable speed.

## Linked Issues
Resolves an issue reported on Discord.
Fixes #3389 

## Proof of Testing

image

### DIFF
--- a/main.py
+++ b/main.py
@@ -188,7 +188,6 @@ from scripts.utility import (
     quit,
 )  # pylint: disable=redefined-builtin
 from scripts.debug_menu import debugmode
-import pygame
 
 
 # import all screens for initialization (Note - must be done after pygame_gui manager is created)
--- a/main.py
+++ b/main.py
@@ -156,9 +156,22 @@ print("Running on commit " + get_version_info().version_number)
 
 # Load game
 from scripts.game_structure.audio import sound_manager, music_manager
+import pygame
+
+pygame.init()
+pygame.display.set_caption("Clan Generator")
+
+from scripts.game_structure.screen_settings import toggle_fullscreen
+from scripts.game_structure.game_essentials import game
+
+toggle_fullscreen(
+    fullscreen=game.settings["fullscreen"],
+    show_confirm_dialog=False,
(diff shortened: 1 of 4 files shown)


j-gynn:
hm. it seems in the process I have unwittingly disabled *all* audio for everyone.

schwoopsie.

j-gynn:
Fixed! There's a compromise. If you know your system doesn't have audio endpoints, you can enable the game_config option `disable_audio`. This skips the 10-second wait whilst pygame farts around trying to get the audio endpoint, but also guarantees that no sound can ever play even if you do add an endpoint, so... swings and roundabouts.

sedgestripe (approved):

PR #3366, opened 2025-01-18
[ENHANCEMENT] improve run scripts

rustykitty:
## About The Pull Request
On `run.sh`, make sure commands succeed before running next command.
~~TODO: do 'cd dirname' on the run.bat~~ Windows automatically starts it in the script's directory

just making the script better to use

## Linked Issues
N/A

## Proof of Testing
Screenshot 2025-01-17 170734

## Changelog/Credits
N/A

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
-cd "$(dirname "$0")" # script location
-poetry install --no-root
+# cd into script location
+cd "$(dirname "$0")" && \
+poetry install --no-root && \
 poetry run python main.py
+
--- a/run.sh
+++ b/run.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
-# cd into script location
-cd "$(dirname "$0")" && \
-poetry install --no-root && \
-poetry run python main.py
+set -e # exit on failure
 
+# cd into script location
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
-cd "$(dirname "$0")" # script location
-poetry install --no-root
+# cd into script location
+cd "$(dirname "$0")" && \
+poetry install --no-root && \
 poetry run python main.py
+
```
archanyhm: I think you can alternatively just use the `set` builtin to accomplish something like this without making the syntax more verbose than it has to be. Frankly, I don't think this script is growing much further than this, but we can make it nice while we're at it.

See: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html and the example below.

CleanShot 2025-01-18 at 13 50 24@2x

rustykitty: Thanks, I didn't know about that!

rustykitty:
The commit I just pushed introduces changes to the Windows Batch script. This will make resolving issues from running from source easier, especially since the command window will stay open after an error.

For example, if Poetry is not installed:
Screenshot 2025-01-22 at 7 53 06 AM

archanyhm (approved):
lgtm

PR #3392, opened 2025-01-30
[CODE] refactor History class

### LINKED ISSUE #2528: [CODE] Refactor History class to use itself correctly
**Description**
Currently, most functions in History take a `Cat` class item to do anything, and then proceed to refer to `cat.history` to do everything...... which is the History class. Refactor this to correctly use `self` and all calls to use `cat.history.[whatever_we're_calling]`.

rustykitty:
## About The Pull Request
As requested, refactored History class to use self, with the exception of two methods, `add_murders` and `reveal_murder`. I wasn't quite sure which cat's history to make the `self` in those.

The one thing I don't like is how I wrote this line:
https://github.com/ClanGenOfficial/clangen/blob/e7249c72b9a2329a92105b8e72c71808ba9a86d9/scripts/game_structure/game_essentials.py#L424

But I can't think of a better way to do it (I assume we don't want to load_history there).

## Linked Issues
resolves #2528

## Proof of Testing

https://github.com/user-attachments/assets/e4f2e59b-0a2b-4bcf-9bab-aa5ee580bd7c


rustykitty:
Game runs! 
Screenshot 2025-03-10 at 8 12 27 PM

scribblecrumb:
Looks like the json test is failing!  Additionally, just looking at the test screenshot, it seems like text replacement isn't working? r_c should be replaced, it shouldn't be visible to the player.

rustykitty:
> Looks like the json test is failing! Additionally, just looking at the test screenshot, it seems like text replacement isn't working? r_c should be replaced, it shouldn't be visible to the player.

I assumed that `r_c` was from the text replacement bug from before. And the json test is failing due to failing to install dependencies, which is weird.

scripts/events_module/short/condition_events.py
```diff
@@ -587,14 +588,15 @@ def handle_already_ill(cat):
...
                 event = event_text_adjust(Cat, event, main_cat=cat)
 
+
                 if cat.status == "leader":
                     event = event + " " + get_leader_life_notice()
                     history_event = history_event.replace("m_c ", "").replace(".", "")
-                    History.add_death(
-                        cat, condition=illness, death_text=history_event.strip()
+                    cat.history.add_death(
+                        ccondition=illness, death_text=history_event.strip()
```
scribblecrumb:
```suggestion
                        condition=illness, death_text=history_event.strip()
```

scribblecrumb (changes requested):
Oof, sorry it took me so long to circle back around to this. I love how much cleaner it feels! Just one little typo to fix and then the merge conflicts (future events definitely changed some of the murder reveal code, so apologies for that)

scribblecrumb (approved):

PR #3396, opened 2025-02-01
Document that bugfixes should be merged to release

larkgz:
## About The Pull Request

Adds information to PR template about which branch bugfixes should be targeted to.

This seems to be a common point of confusion because it doesn't seem to be documented anywhere.


j-gynn:
I wonder whether we ought to put this in `CONTRIBUTING.md` too if it's not already there - also on the wiki. It's important to also note that you can't just point a `development`-originating branch onto `release`, so this point needs to be made before the PR submission.

larkgz:
@j-gynn I added some text to contributing.md and the docs. Wasn't sure about the page on the docs, but that is the only page about GitHub setup, so I put it there.

And then, maybe in the future we can add docs on how to cherrypick, in case someone *did* base it off the wrong branch?

j-gynn (approved):

PR #3397, opened 2025-02-01
Heterochromia odds as intended

hschawe:
## About The Pull Request

In pelts.py, the odds of heterochromia was not properly increased for cats with mostly/high/full white_patches, because the condition was checking against a nested list. This fixes the check while preserving its original odds as closely as possible.

Effectively this increases the odds of heterochromia for cats with mostly_ and high_white white patches.

we want the code to work as intended, especially since we've been communicating publicly that the odds are increased for cats with white_patches

## Linked Issues

Fixes https://github.com/ClanGenOfficial/clangen/issues/3373

## Proof of Testing

Ran with debugger to verify odds for white_patches, FULLWHITE, WHITE cats (sorry for the massive images):
Screenshot_7
Screenshot_6
image


j-gynn (approved):
LGTM

PR #3398, opened 2025-02-02
[BUGFIX] exclude first starclan cat from sorting

### LINKED ISSUE #3088: [BUG] First StarClan cat is not excluded from sorting
### Type of bug

### Description

Previously, when StarClan cats were sorted, the first StarClan cat was excluded from sorting and was always the first cat in the carousel (for lack of a better term) when clicking Previous/Next. The first cat is still pinned to the first position in the list of cats, but its position in the carousel changes when the list is sorted.

### Grade

1. Open the Cat List
2. Click "view dead" 
3. Sort by a factor where the first StarClan cat would not be first in the list (IE rank, age ascending, EXP)
4. Click on the first StarClan cat and check behaviour of the "Previous Cat" and "Next Cat" buttons

_No response_

v.012.1

### Additional context

_No response_

### Screenshots

https://github.com/user-attachments/assets/289ab695-7cfd-4dde-9982-43888231d616

rustykitty:
## About The Pull Request
I made sure that the `instructor` is always at the front of the list in determining the next/previous cat.

## Linked Issues
resolves #3088

## Proof of Testing
Screenshot 2025-02-02 at 9 07 30 AM
Screenshot 2025-02-02 at 9 07 28 AM


poetry.lock
j-gynn, line 1: Why is Poetry being updated in this PR?

rustykitty: I did `git commmit -am` and I think it was a modified file, but I'll revert

j-gynn (approved):

PR #3402, opened 2025-02-04
[ENHANCEMENT] Allow cats to have multiple accessories

### LINKED ISSUE #2320: [CODE] Allow multiple accessories on a cat
**Description**
Allow a cat to have multiple accessories at once, without having them overlap with each other - for example, no two head accessories at once. I am aware of at least one category of accessories already (tail_accessories) so I suggest making similar categories for the other types of accessories as well (head, neck, back?) if those don't exist already.

Note: This should wait until Scribble's moon event refactor project is completed!
Forum post for the refactor: https://discord.com/channels/1003759225522110524/1232886844644134923

KiraDragoness:
## About The Pull Request

Cats are now able to have up to 3 different accessories.
All accessories have been divided into sections according to the body parts they cover: head, collar, body and tail (so they can't have multiple accessories for the same body part). These categories also determine the rendering order in the sprite generation, to avoid overlapping visuals being weird.
After getting the first accessory, there's a lower chance of getting a second and third one, which can be customized in game_config.
Old save files will convert the accessory attribute of each cat from string to list.

It will add more variety and uniqueness for cats in the Clan, and it's a feature that was already suggested as an enhancement by senior devs.

## Linked Issues
Fixes: #2320

## Proof of Testing
Beta testing thread: https://discord.com/channels/1003759225522110524/1336115112053833738
image

## Changelog/Credits
Enhancement: Allow cats to have multiple accessories

### DIFF
--- a/scripts/events_module/short/handle_short_events.py
+++ b/scripts/events_module/short/handle_short_events.py
@@ -399,8 +399,21 @@ class HandleShortEvents:
                     if acc in acc_list:
                         acc_list.remove(acc)
 
+        acc_categories = ["collars", "head_accessories", "tail_accessories", "body_accessories"]
+
+        if self.main_cat.pelt.accessory:
+            for acc in self.main_cat.pelt.accessory:
+                for category in acc_categories:
+                    if acc in getattr(pelts, category):
+                        for accessory in getattr(pelts, category):
+                            if accessory in acc_list:
+                                acc_list.remove(accessory)
+
         if acc_list:
-            self.main_cat.pelt.accessory = random.choice(acc_list)
+            if self.main_cat.pelt.accessory:
+                self.main_cat.pelt.accessory.append(random.choice(acc_list))
+            else:
+                self.main_cat.pelt.accessory = [random.choice(acc_list)]
 
     def handle_transition(self):
         """
--- a/scripts/events_module/short/handle_short_events.py
+++ b/scripts/events_module/short/handle_short_events.py
@@ -412,7 +412,6 @@ class HandleShortEvents:
                                 acc_list.remove(accessory)
 
         if not acc_list:
-            print("Not accessories for", self.main_cat.name)
             return False
 
         if self.main_cat.pelt.accessory:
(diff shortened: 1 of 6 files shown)


scripts/events_module/short/handle_short_events.py
```diff
@@ -399,8 +401,24 @@ def handle_accessories(self, pelts=Pelt):
...
+        acc_categories = ["collars", "head_accessories", "tail_accessories", "body_accessories"]
+
+        if self.main_cat.pelt.accessory:
+            for acc in self.main_cat.pelt.accessory:
+                for category in acc_categories:
+                    if acc in getattr(pelts, category):
+                        for accessory in getattr(pelts, category):
+                            if accessory in acc_list:
+                                acc_list.remove(accessory)
+
```
j-gynn: Horrifying.
(/silly)

It looks like this is trying to remove accessories that are in an already-occupied slot? Had to sit here and think about it for a hot minute buuut I think this is good? Please verify, I haven't run this code so it might explode!

```suggestion
accessory_groups = [pelts.collars, pelts.head_accessories, pelts.tail_accessories, pelts.body_accessories]

for acc in self.main_cat.pelt.accessory:
    # find which accessory group it belongs to
    for i, lst in enumerate(accessory_groups):
        if acc in lst:
            # remove that group from possible accessories
            acc_list = [a for a in acc_list if a not in accessory_groups[i]]
            break
```

j-gynn (changes requested):
Whilst not the *hugest* problem, I had to triple-take at the section that I *think* is for removing overlapping items from the pool when giving a new accessory. I've provided a potentially cleaner solution, but it's untested, so please, y'know... check it works before using it...

As an aside, very nice implementation overall! Very clean, me likey.

scripts/events_module/short/handle_short_events.py
KiraDragoness: Oh thank goodness you know of a better way of doing that, I knew it was ugly af but I couldn't figure out how to make it better. I shall test that change tomorrow!

KiraDragoness: Tested and it works just fine! I just had to add a previous check to ensure the cat already has accessories

j-gynn (approved):
LGTM!

PR #3403, opened 2025-02-04
Migrate from ujson

larkgz:
## About The Pull Request

Switches out ujson for orjson. There are some slight API differences, so there's a shim to avoid needing to modify lots of code in many different places.

Also, the shim only “supports” the parameters we’ve actually used (there's some slightly different behaviour). Indents will all be 2-spaces if indents are requested because that’s all that’s supported, and by default orjson does no ascii escaping for utf, so that parameter does nothing. 

@j-gynn pointed out that ujson is currently in maintenance-only mode:

https://pypi.org/project/ujson/
> [!WARNING] UltraJSON's architecture is fundamentally ill-suited to making changes without risk of introducing new security vulnerabilities. As a result, this library has been put into a maintenance-only mode. Support for new Python versions will be added and critical bugs and security issues will still be fixed but all other changes will be rejected. Users are encouraged to migrate to orjson which is both much faster and less likely to introduce a surprise buffer overflow vulnerability in the future.

## Notes

We should probably discuss it more before merge (if we decide to at all).


archanyhm:
This is quite dated, so we'll probably have to bring this back up to date. What's the status here since it's still a draft?

larkgz:
The reason I marked it as draft is because I wanted to make sure that we discussed it first before thinking about merging it. But if we're OK with the switch, then after I have it working with uv I can mark it as ready for review.

archanyhm:
> The reason I marked it as draft is because I wanted to make sure that we discussed it first before thinking about merging it. But if we're OK with the switch, then after I have it working with uv I can mark it as ready for review.

I don't mind moving away to a more well-maintained library in this case. The only thing I'm unsure about is whether we want to shim this or whether it wouldn't be easier to just replace all function calls.

larkgz:
Here are my other reasons for making it a shim:
* orjson does not support json functions dump or load
* orjson options are specified pretty differently than ujson / Python’s default json api. Specifically, in orjson, options are given via bitmask
* I’m pretty sure any web ports would have to use a shim for libraries like this anyway
* orjson has slightly different types (i.e. exports `bytes` rather than `str`)

tests/test_json.py
```diff
@@ -12,7 +12,7 @@
 import sys
 import unittest
 
-import ujson
+import json as ujson
```
scribblecrumb: Should this be using the shim?

scribblecrumb (changes requested):

Ryos00:
Can be reopened if still relevant.

Issue #3194, opened 2024-12-17
[BUG] Unknown murderer revealed when pressing button in events log


sketchywyvern:
### Type of bug

Code - general bugs related to how the game runs and responds to you.

### Description

See title

### Grade

Regular - A typical bug that should be seen to soon, but doesn't necessarily break the game or inhibit the player

### Steps to reproduce the behavior

1. Have a cat be murdered by an unknown cat
2. Click on the button in the events log
3. Names shown will be the victim (expected) and the murderer (shouldn't happen)

### Commit # or Game Version Number

0.12.2

### Screenshots

![image](https://github.com/user-attachments/assets/4134584c-ebb5-4553-b135-c72af342c856)

j-gynn:
Unfortunately, this is a limitation of the murder system at present.

sketchywyvern:
Ah, sorry! I never noticed it before so I thought it was a bug.

j-gynn:
Closing as resolved by #3430

PR #3430, opened 2025-02-11
[ENHANCEMENT] "exclude_involved" added to short events

jaysoleson:
## About The Pull Request

hello! this PR adds an "exclude_involved" parameter to short events, specifically to be utilised in "unknown murder" events. currently, the "the culprit is unknown" murder event shows the murderer in the involved cats buttons, which spoils the mystery!

this will solve the issue of unknown murder events being spoiled by the involved cat buttons. additionally, a cat not mentioned in the text being a part of the involved cats is commonly seen as a bug by new players.
this can also provide more writing opportunities for murders with secret culprits.

## Proof of Testing

the unknown murder event's involved cat buttons, with the addition to the JSON and without!

exclude_involved_after
exclude_involved_before

and here's what it looks like as part of the event dict:
exclude_involved_json_ss

## Changelog/Credits

"Unknown murderers will no longer be involved in their murder event." or somesuch :]

### DIFF
--- a/resources/lang/en/events/death/general.json
+++ b/resources/lang/en/events/death/general.json
@@ -27,6 +27,7 @@
             "age": ["adolescent", "young adult", "adult", "senior adult"],
             "status": [ "any" ]
         },
+        "exclude_involved": ["r_c"],
         "history": [{
             "cats": ["m_c"],
             "reg_death": "m_c was secretly murdered by r_c."
--- a/scripts/events_module/generate_events.py
+++ b/scripts/events_module/generate_events.py
@@ -145,6 +145,7 @@ class GenerateEvents:
                         r_c=event["r_c"] if "r_c" in event else {},
                         new_cat=event["new_cat"] if "new_cat" in event else [],
                         injury=event["injury"] if "injury" in event else [],
+                        exclude_involved=event["exclude_involved"] if "exclude_involved" in event else [],
                         history=event["history"] if "history" in event else [],
                         relationships=event["relationships"]
                         if "relationships" in event
--- a/scripts/events_module/short/handle_short_events.py
+++ b/scripts/events_module/short/handle_short_events.py
@@ -190,6 +190,12 @@ class HandleShortEvents:
             self.handle_mass_death()
             if len(self.multi_cat) <= 2:
                 return
+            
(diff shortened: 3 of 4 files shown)


j-gynn:
Hi! Thank you for your interest in this issue. There is a bit of division within the community as to whether or not "unknown murderer" events should show who the murderer is or not. Some players say no: it's an unknown murderer, the point is that we don't know. Some players say yes: they want to write or draw the story and knowing who did the murdering is important to that - even though the Clan doesn't know.

#3194 discusses this issue and it has been labelled `limitation`, but as always the problem is a bit more nuanced than that. I wonder if perhaps a clan setting could be created to enable/disable showing the murderer in unknown murderer events...?

scribblecrumb:
Even if we table the issue of player murder awareness, this is still a good parameter to have. It's some added functionality that I'm certain can be used in other events.

We actually have a less-functional version of it via a new_cat tag ("unknown" i think?) that obv can only be used for a new cat.  I'd love to see this implemented in a way that allows its use in patrols as well, then we could just remove that new_cat tag, as this is a more versatile use of it.

Also a reminder that this should be added to the relevant doc files for event writing.

jaysoleson:
> Hi! Thank you for your interest in this issue. There is a bit of division within the community as to whether or not "unknown murderer" events should show who the murderer is or not. Some players say no: it's an unknown murderer, the point is that we don't know. Some players say yes: they want to write or draw the story and knowing who did the murdering is important to that - even though the Clan doesn't know.
> 
> #3194 discusses this issue and it has been labelled `limitation`, but as always the problem is a bit more nuanced than that. I wonder if perhaps a clan setting could be created to enable/disable showing the murderer in unknown murderer events...?

thats fair!! if you did decide to go the route of a clan setting, i think there might have to be a bit of a makeover done to the current murder events (since a clan setting changing the functionality of literally One short event is a bit dramatic lol). maybe the "enable secret murders" toggle could hide ALL murderers from the player until it's been revealed to the Clan too, while playing with it off would only enable murder events that tell you who did it? i'd personally love to play with the secret toggle on!

this would require more secret murders to be written, but i think it's a fun, spooky writing opportunity!

scribblecrumb (approved):

Issue #3444, opened 2025-02-17
[BUG] Patrol crash when Clan has no deputy


Chinch-Bug:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

Patrol with dep_name in an outcome crashes the game if there is no deputy

### Grade

Regular - Should be handled soon, but doesn't necessarily break the game or inhibit the player

### Steps to reproduce the behavior

1. Make sure Clan has no current deputy
2. Send cats hunting (to trigger it, I set the patrol id in game_config, and set it to always fail)
3. Repeat the 'cat is tempted to eat the prey they just caught' (gen_hunt_sneakprey3) patrol until you get an outcome with dep_name in it - this should crash the game

### Patrol ID + Outcome Print (if applicable)

gen_hunt_sneakprey3
It has a fail outcome with dep_name in it

### Commit # or Game Version Number

ea0a08dc

### Error log

root - CRITICAL - main.py / log_crash / 118 - Uncaught exception
Traceback (most recent call last):
  File "C:\Python312\Lib\runpy.py", line 198, in _run_module_as_main
    return _run_code(code, main_globals, None,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\runpy.py", line 88, in _run_code
    exec(code, run_globals)
  File "c:\Users\smare\.vscode\extensions\ms-python.debugpy-2025.0.1-win32-x64\bundled\libs\debugpy\launcher/../..\debugpy\__main__.py", line 71, in <module>
    cli.main()
  File "c:\Users\smare\.vscode\extensions\ms-python.debugpy-2025.0.1-win32-x64\bundled\libs\debugpy\launcher/../..\debugpy/..\debugpy\server\cli.py", line 501, in main
    run()
  File "c:\Users\smare\.vscode\extensions\ms-python.debugpy-2025.0.1-win32-x64\bundled\libs\debugpy\launcher/../..\debugpy/..\debugpy\server\cli.py", line 351, in run_file
    runpy.run_path(target, run_name="__main__")
  File "c:\Users\smare\.vscode\extensions\ms-python.debugpy-2025.0.1-win32-x64\bundled\libs\debugpy\_vendored\pydevd\_pydevd_bundle\pydevd_runpy.py", line 310, in run_path
[...]

PR #3445, opened 2025-02-17
[BUGFIX] Fix patrol crashing when there's no deputy

Chinch-Bug:
## About The Pull Request

Had a crash brought to my attention and replicated it on basegame - patrols can cause a crash if they include dep_name in outcome, while the Clan has no deputy.
Added a few lines into the outcome filtering that exclude any outcomes with lead_name or dep_name while the needed cats don't exist

## Linked Issues

#3444

## Proof of Testing

Repeated the patrol about 10 times with no further crashes
image
image

### DIFF
--- a/resources/lang/en/patrols/general/hunting.json
+++ b/resources/lang/en/patrols/general/hunting.json
@@ -440,7 +440,7 @@
         "biome": ["any"],
         "season": ["any"],
         "types": ["hunting"],
-        "tags": [],
+        "tags": ["clan:deputy"],
         "patrol_art": "gen_hunt_sneakprey",
         "min_cats": 2,
         "max_cats": 6,
@@ -550,7 +550,7 @@
         "biome": ["any"],
         "season": ["any"],
         "types": ["hunting"],
-        "tags": [],
+        "tags": ["clan:deputy"],
         "patrol_art": "gen_hunt_sneakprey",
         "min_cats": 2,
         "max_cats": 6,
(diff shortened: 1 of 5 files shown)


j-gynn:
Hi there! Thanks for reporting & producing a bugfix.

However, there's an existing solution to this problem in `ShortEvents` through the use of `clan:leader` and `clan:deputy` tags (amongst others like `apps`). To avoid confusion, it'd be nice if both systems could use the same format.

j-gynn (changes requested):
see comment for discussion

Chinch-Bug:
Edited it to apply a "clan:deputy" tag to the patrols found. Patrol now gets filtered out entirely if there is no deputy.
Hope that's better/the expected result

scribblecrumb:
I wonder how viable it would be to utilize the `event_for_tags` function in `event_filters.py`?  It'd be best not to be repeating the `clan:` filtering code in two places.  However the `event_filters.py` code was written with the aim of using it for short events and eventually reworking relevant code to make it viable for patrols as well.  I'm unsure how easily this section of it could be made to fit patrols right now.

If it's not relatively simple matter, then it's probably fine to just leave it how you've got it atm, Chinch.  But it's something to consider.  We *do* eventually want our myriad of event formats to be as close in parity as possible.

Chinch-Bug:
Moved over the tag-based filtering to be done in `event_for_tags`, a few moonskips and random patrols don't show any immediate issues with that. Just put any checks that require a specific cat object under a `if hasattr(cat, "ID")` block

j-gynn (approved):

PR #3457, opened 2025-03-08
fix: temporarily disable the auto-updater

archanyhm:
## About The Pull Request

Now, the auto-updater is not being triggered when pressing "continue" in the update popup; instead it redirects the user to the ClanGen download page on the website (https://clangen.io/download or https://clangen.io/download-development). Reason is that the auto-updater currently isn't functional and breaks people's installs upon execution; a temporary fix to it is thinkable, but a proper rewrite into a more generic project is the sounder way in my opinion, since I'm the one maintaining this.

Removes the broken option for the time being.


j-gynn (approved):
LGTM

PR #3460, opened 2025-03-14
Update clangen font to include Cyrillic

j-gynn:
## About The Pull Request
Adds Cyrillic to the Clangen font. Also updates the info section with attribution for the various bits of our frankenfont - although not strictly necessary by the licenses used, it's both polite and helpful for anyone curious about how exactly it goes together.

There is interest in porting to Russian and adding Cyrillic support enables that.

## Proof of Testing
image
image


larkgz:
@howlagon Can you take a look please?

Thlumyn (approved):

tailhaver:
lgtm for now~! the current clangen font isn't left aligned, is it? if it is, then it should be left aligned, otherwise it's perfectly fine.

PR #3461, opened 2025-03-14
[BUGFIX] Fixed problems related to raiding other Clans

KiraDragoness:
## About The Pull Request
I found a crash caused when trying to raid other Clans (one or more) using the focus feature. Whenever you tried to raid any Clan, the game crashed upon moonskipping with the following crash log:

image

The crash was caused by a list being passed through a count, instead of the length of said list.

Upon fixing this, I realized cats that were getting sick because of the current focus weren't being listed in the event:
image

This was caused by an indentation problem, which was making the code completely skip the loop to correctly add the sick cats' IDs to the event. 

## Proof of Testing
image
image
image

## Changelog/Credits
[BUGFIX] Fixed raid focus causing crash + cats getting sick with focus not being listed in event.

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -775,7 +775,7 @@ class Events:
             for condition_type, value in involved_cats.items():
                 game.cur_events_list.append(
                     Single_Event(
-                        i18n.t(text_snippet, condition=condition_type, count=value),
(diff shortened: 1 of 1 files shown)


j-gynn (approved):
LGTM

PR #3475, opened 2025-03-27
[ENHANCEMENT] format all json files to use 4 spaces

rustykitty:
## About The Pull Request
Ran PyCharm's "Reformat Code" on every `.json` file in the repo.

I apologize for kind of going nuclear, but we had some hard to read JSON files, especially in the patrol dicts. Time lost from trying to comprehend code is time!

## Linked Issues
N/A

## Proof of Testing
image
Patrol *definitely* has to pull from something I touched, IMO.

## Changelog/Credits
N/A


j-gynn:
These changes are not just on JSON files, and not just to convert to using 4 spaces. Additionally, this is a huge PR (which is hypocritical for me to say but- shh). Please be mindful of what you are running and on what - you've touched a great more than your PR title and description implies.

With that in mind, I would prefer to see individual PRs per major directory. One for patrols, one for thoughts, one for events, etc. - that sort of size is even still a bit large, but more reasonable. I know it's barely any work on your end if you're just clicking a button (believe me, I did some of the linting as one of my earliest PRs here), but it's an insane amount of work to verify and ensure that nothing silly happened as a result. Case in point, my own PR doing something similar caused a game-breaking bug of rather epic proportions, simply because the PR was so massive that a single line of me poking around had slipped through the cracks.

In summary: be mindful of what filetypes you're actually running this formatting on; do multiple, smaller PRs targeting specific areas individually; ensure your PR title and description matches what you're actually doing. I'll be closing this PR but I look forward to receiving the smaller ones in future :)

PR #3479, opened 2025-04-02
April Fools hats will appear on April Fools

larkgz:
## About The Pull Request

Replaces lineart with the hat lineart on April Fools.

Technically there's another April Fools thing that's supposed to happen (newborns roaming and patrolling) but we'd have to fix a circular import to get that working with the special date code.

* It's fun
* I forgot to add it when I added automatic special date support

## Proof of Testing
No hats on April 2nd :( 
image

Hats on April 1st :)
image

## Changelog/Credits

Hats will now automatically appear on April Fools.

### DIFF
--- a/scripts/cat/sprites.py
+++ b/scripts/cat/sprites.py
@@ -5,6 +5,7 @@ from copy import copy
 import pygame
 import ujson
 
+from scripts.special_dates import SpecialDate, is_today
 from scripts.game_structure.game_essentials import game
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


larkgz:
I forgot the egg suffixes... But that kind of feels a bit more disruptive than the hats (even if it doesn't change names -- it's still confusing), so I kind of would prefer that doesn't auto-activate, especially if we decide to add extra April Fools Day content in the future.

scribblecrumb (approved):

Issue #3330, opened 2025-01-01
[BUG] 'focus last changed' text is unavailable


selkirks:
### Type of bug

UI - anything to do with the user interface - maybe a button is not working, or the way some assets are formatted seems to be off.

### Description

usually there should be a number of moons next to the 'focus last changed' button. its no longer present

### Grade

Regular - A typical bug that should be seen to soon, but doesn't necessarily break the game or inhibit the player

### Steps to reproduce the behavior

1. open a clan
2.  go to the change focus screen
3. error occurs

### Commit # or Game Version Number

88ae28e97a877297b32203b0e1df4876bf4e979f

### Screenshots

![image](https://github.com/user-attachments/assets/7bb0c400-fd36-4db9-8384-c746c9976d29)

scribblecrumb:
Unfortunately this doesn't provide any helpful info for replicating the bug.  If you can provide any further information, or perhaps a save file for the Clan, then we might be able to fix it.  I can't seem to replicate it on my own, so there must be some triggering factor.  Do you know how old your Clan was? Does this happen every time you open it?  Does it reappear if you timeskip or leave the focus screen and then reenter?

scribblecrumb:
wait, ignore that, i was on the release branch still and didn't notice this was for dev specifically lmao

PR #3482, opened 2025-04-02
BUGFIX : Fix Focus screen not updating with new clans

### LINKED ISSUE #3168: [BUG] warrior's den does not update "last changed" display correctly on first focus change
### Type of bug

UI - anything to do with the user interface - maybe a button is not working, or the way some assets are formatted seems to be off.

### Description

i started a new clan and immediately went to change my focus to "feeding the clan". the only thing i did beforehand was send the medicine cat out on patrol. after changing focus, the warrior's den still says the last change is unknown and the next change is in 0 moons. it did not update even after skipping a moon. this appears to be only visual, as i was not able to change my focus again until moon 3. at this point, the display updated to correctly show what moon the focus was changed on and how long the cooldown is.

### Grade

1. create a new clan
2. go to the warrior's den and change the focus
3. the cooldown and last changed on display will not update correctly

_No response_

45e07794a03f69a35b6a76606ddc90a713e15644

### Additional context

_No response_

### Screenshots

image
image

scribblecrumb:
## About The Pull Request
A clan with an age of 0 was prevented from updating the Focus screen due to 0 being detected as a NoneType, this accounts for that edgecase.  I also removed an unused ["time"] element.

## Linked Issues
Fixes #3168 
I believe will also take care of #3330 once this is also merged to dev?

## Proof of Testing
image

## Changelog/Credits
- Warrior's Den screen now correctly updates for Clans on Moon 0

### DIFF
--- a/scripts/screens/WarriorDenScreen.py
+++ b/scripts/screens/WarriorDenScreen.py
@@ -302,8 +302,6 @@ class WarriorDenScreen(Screens):
         # delete previous created text if possible
         if "current_focus" in self.focus_information:
             self.focus_information["current_focus"].kill()
-        if "time" in self.focus_information:
-            self.focus_information["time"].kill()
         if self.focus_text:
             self.focus_text.kill()
 
@@ -322,7 +320,7 @@ class WarriorDenScreen(Screens):
 
         last_change_text = "unknown"
         next_change = "0 moons"
-        if game.clan.last_focus_change:
+        if game.clan.last_focus_change == 0 or game.clan.last_focus_change:
             last_change_text = "moon " + str(game.clan.last_focus_change)
             moons = (
                 game.clan.last_focus_change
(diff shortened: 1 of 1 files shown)


scripts/screens/WarriorDenScreen.py
```diff
@@ -322,7 +320,7 @@ def create_top_info(self):
 
         last_change_text = "unknown"
         next_change = "0 moons"
-        if game.clan.last_focus_change:
+        if game.clan.last_focus_change == 0 or game.clan.last_focus_change:
```
j-gynn: could we also consider:

```suggestion
        if isinstance(game.clan.last_focus_change, int):
```

IMO, this is much clearer than the previous, since it appears to imply that `game.clan.last_focus_change` can either be `0` or `True`, which is... confusing.

j-gynn (changes requested):
Code legibility suggestion.

j-gynn (approved):

PR #3487, opened 2025-04-03
BUGFIX: Clan symbol displaying as "blank"

### LINKED ISSUE #3318: [BUG] Blank Clan Symbols
### Type of bug

Sprite - an issue with the art in the game

### Description

I am somehow generating clans with blank symbols. I've tried to see if this happens only with quick creation, but it also appears with manual clan building as well

### Grade

Regular - A typical bug that should be seen to soon, but doesn't necessarily break the game or inhibit the player

1. Open the game
2. Get to where you see the clan symbol for the first time
3. Blue screen brain

_No response_

0.12.0 or f1346a01

### Additional context

This is not a one-time thing - every clan I make seems to have this issue. This also only happens for the chosen symbol?? The strange behavior is shown below. I genuinely have no idea if this is github desktop messing with my files 😭 

### Screenshots

Screenshot (1656)
Screenshot (1659)

Muxasii:
## About The Pull Request

Changes line 777 distance=0.2 to distance=0

This caused the symbol to display as a square because "distance" is meant to be distance away from the specified color so it's recoloring everything including the transparency instead of JUST the color specified

unsquares your clangen *beams*

## Linked Issues

Fixes: #3318

## Proof of Testing

image
image
image

## Changelog/Credits

- Clan symbol code no longer recolors the transparent pixels

### DIFF
--- a/scripts/cat/sprites.py
+++ b/scripts/cat/sprites.py
@@ -774,7 +774,7 @@ class Sprites:
             pygame.Color(game.config["theme"]["dark_mode_clan_symbols"])
             if not force_light and game.settings["dark mode"]
             else pygame.Color(game.config["theme"]["light_mode_clan_symbols"]),
-            distance=0.2,
+            distance=0,
         )
         del var
 


j-gynn (changes requested):
The `distance` argument is needed to ensure that minor alterations to the icons that occur due to lossy compression still get their colour changed - we were getting speckles occasionally due to it. Perhaps it was simply set too high and the cutoff should be much lower? I can't track down the issue if one was ever made, but it was mentioned in #3274.

Muxasii:
That doesn't look like the right PR to me? That one is a huge fix on the symbols.png file not something in the code, unless I'm misinterpreting (I had only 3 hours of sleep so apologies)
(NEVERMIND I UNDERSTAND NOW but it's unfortunate that this isn't going to work :,3)

But anyway, it colors transparency regardless of the value of distance set.
image

j-gynn:
Ack, you're right. It looks like the real solution here is to avoid the transparency issue by saving the sprites with an eye-searing background and calling `set_colorkey` later, as [recommended in this StackOverflow post](https://stackoverflow.com/questions/32145145/keep-alpha-in-pygame-pixelarray). I suppose that's out-of-scope of this particular PR though. Well, we can reinvestigate it if the spotty pixels recur.

Change approved, thanks for investigating for me!

j-gynn (approved):
LGTM

PR #3498, opened 2025-04-05
Added Text for Hoarding and Raiding Results

rslashrats:
## About The Pull Request

I added new categories inside hardcoded.en.json to include text that prints if no cats receive conditions while the Clan is currently hoarding or raiding other Clans.

It prevents text from printing every moon skip that claims that a cat got sick and injured during a moon skip, despite it not happening - which could confuse the player. 

## Proof of Testing

image
image
image

## Changelog/Credits

N/A


j-gynn:
I wonder if a better solution might be code-side to just... not put the event on the log if nobody got injured? I don't recall what other focuses do in a "nothing happened" situation.

j-gynn (approved):
LGTM

PR #3510, opened 2025-04-22
[Content/Bugfix] More Thoughts & Thought Edits

tobes03:
## About The Pull Request

Just adding thoughts to the general thoughts file, as well as minor edits on some that are already in there.

More content! more thoughts!

## Proof of Testing
image
image
image


tobes03:
for a future PR: there are a bunch of apprentice thoughts in here that will be moved into the apprentice-specific file.
edit: nvm, looks like they're in gen because they're warrior app & medicine app & mediator app. That being said, there are a few that are specific to only one rank and those do actually need to be moved.

resources/lang/en/thoughts/alive/general.json
```diff
@@ -3800,7 +4025,10 @@
             "Got in trouble for shirking {PRONOUN/m_c/poss} training the other day",
             "Is talking back to {PRONOUN/m_c/poss} mentor",
             "Is mocking another apprentice",
-            "Was told off by {PRONOUN/m_c/poss} mentor for not taking {PRONOUN/m_c/poss} training seriously enough"
+            "Was told off by {PRONOUN/m_c/poss} mentor for not taking {PRONOUN/m_c/poss} training seriously enough",
+            "Spreads rumors about {PRONOUN/m_c/poss} mentor",
+            "Ignores {PRONOUN/m_c/poss} mentor's advice and substitutes {PRONOUN/m_c/poss} own",
+            "Tattletales on a fellow apprentice"
```
scribblecrumb, line 4127: this needs to required random_cat to be an app so that we can guarantee the clan has multiple apps!

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #3534, opened 2025-04-30
[ENHANCEMENT] Newborn Extended Family Relationships

scribblecrumb:
## About The Pull Request
This PR allows extended family to automatically generate relationships toward newborn family members based off their view of the parents.  Essentially, it mirrors their relationship with the parents onto the new kittens. 

The relative's relationship values toward the parents are quartered (divided by 4) and then applied to their relationship with the newborns with a modifier between -10 and 10 to create some variance.  A relationship log is generated to to mark this event as positive, negative, or neutral (aka conflicted, the relative has both negative and positive relationship values with the parents).

This helps establish the newborns within the Clan dynamic much easier and creates links between related cats. After all, a grandparent should have some sort of opinion of their grandkits!  It adds some fun dynamics, like cats immediately having negative view of kits due to their negative view of the parents (just like in canon ;) ).  These relationships do still tend to be lower value, so there's opportunity for them to completely change with time!

## Proof of Testing
image

New kittens! Let's see what the other members of this family think.

image

Looks like Termitepaw, a cousin, has a bit of interest!

image

Mulberry, an older sibling, has much more investment already.

image

For some reason, she likes this new sibling much more lmao

image

And grandma certainly has a least favorite grandchild

image

Gander, an uncle, also seems rather conflicted.

image

Also wanted to note that Sandypath, a grandparent to the new kits who isn't part of the Clan, features no relationship to them! Since she isn't in the Clan, there's no way she would know of their birth so she has no reaction.

## Changelog/Credits
- Extended family of newborns will now generate relationship values reflective of the relative's opinion toward the parents. This means that, for example, grandparents will have relationships toward their grandkits! Aunts and uncles might be doting (or... perhaps resenting...)  Who knows what sort of drama this could unfold...


scripts/events_module/relationship/pregnancy_events.py
```diff
@@ -966,6 +975,99 @@ def get_kits(
             blood_parent.outside = True
             clan.unknown_cats.append(blood_parent.ID)
 
+        # check for more extended family members to create relationships with
+        for kit in all_kitten:
+            for cat_id in kit.get_relatives():
```
j-gynn: I wonder if this order is correct. It's a minor thing, but all kits in the same litter should have the same collection of family members (since they have the same biological parents, and this triggers the moment the potatoes pop). 

It might make more sense to get every related cat's ID and put it in a list so we don't have to iterate through X number of kits (many) relatives multiple times. Does that make sense? Basically make some variable `all_relatives` equal to `all_kitten[0].get_relatives()`, which could then allow for list comprehensions to be used to randomly allocate later? Just reduces the number of nested if statements, since I can see this function MASSIVELY bloating moonskips on big clans (they're already hefty when a new litter is born).

If you can get hold of a very large test clan (there's one shared in the discord iirc that's pretty chonk) maybe you can compare whether this has a significant impact on load times.

scripts/events_module/relationship/pregnancy_events.py
```diff
@@ -966,6 +975,99 @@ def get_kits(
...
+
+                # this finds what the relative's relationship is toward each parent and applies a reflection of that
+                # relationship to the kit. reflection values will be divided by 4 by default and then modified
+                # by the random y value
+                for parent_id in parents:
+                    try:
+                        relation_toward_parent = relative_cat.relationships[parent_id]
+                    except KeyError:
+                        continue
+                    relation_toward_kit = relative_cat.relationships[kit.ID]
```
j-gynn: Thinking out loud here, it might be beneficial to look at whether we can make the relationship class iterable over the attributes.

j-gynn (changes requested):
One kinda major, one minor.

scribblecrumb:
Alright, I've updated this to the new rel (and status, damn it's been a while) systems. I've also done some profiling to see the performance impact! On my 39 cat clan I profiled the `get_kits` func both on this branch and on the dev branch. There was one cat giving birth with 11 relatives that would form relationships to the kits. This was using the exact same save beginning from the exact same state each time.

In this branch, the func took 4.58213 seconds while dev took 4.50778 (keep in mind that the profiler adds some overhead, so using it does make this func just take longer in general.) While this is a pretty small scale test, I think it does demonstrate that the performance cost here is pretty negligible imo

j-gynn (approved):

PR #3546, opened 2025-05-09
Trait Range Edits According To Polls

keplerstelescope:
## About The Pull Request

editing trait_ranges.json according to the results of the polls! some trait renames and removals (most if not all will return as display traits afaik) and one additional trait!
(let me know if i should have the additional trait in a different PR or anything like that!)

https://discord.com/channels/1003759225522110524/1366932294203473970/1368610452636041338


j-gynn (changes requested):
1. If we're removing or renaming traits, we should simultaneously be removing them from any tagged content they're included in. If it's not done now, it won't get done. I appreciate that's not exactly what you signed up for - sorry! It should be a relatively simple find-and-replace task.

2. Please ensure that `resources/lang/en/cat/personality.en.json` is updated with these changes, as this is what actually displays in-game. Otherwise, these might crash the game or just display weirdly!

scribblecrumb:
Looks like there's json errors to fix!

Ryos00:
May be reopened for review once changes are made.

PR #3557, opened 2025-05-30
Migrate to uv

archanyhm:
## About The Pull Request

Back when we switched to poetry for easier management of dependencies, uv didn't exist. Now, obviously it exists, and does a few things notably better:
- Resolving dependencies is significantly faster
- Its CLI is slightly less verbose
- It's a standalone binary, so it's not necessary to have Python installed. uv automatically creates a .venv with the appropriate version defined in the .python-version file.
- It is significantly easier to install
- It has saner defaults (creates .venv by default, unlike poetry)

See above


j-gynn (approved):

j-gynn:
Looks like these changes have caused a test to fail on one of the messy helper scripts...? If you could make that stop screaming, then you can go ahead and merge!

PR #3562, opened 2025-06-06
[CLEANUP] open data directory & social buttons

rustykitty:
## About The Pull Request
Consolidate platform-specific code for Open Data Directory and Socials Buttons, which were previously in multiple places.

Not only is it easier to maintain since its only in one place, the layer of abstraction may help with less experienced coders.

## Linked Issues
N/A


rustykitty:
Linux (Mint Cinnamon 22.1 Xia):

https://github.com/user-attachments/assets/f8903a53-2c4a-47ec-af26-29d866aef5a2

Ignore the console errors, those pop up whenever I launch the file manager and I'm not sure why. Doesn't seem to affect behavior though

scripts/utility.py
j-gynn, line 1: I feel as though `open_data_dir` and `open_url` should live in `datadir.py`?

j-gynn (changes requested):
Move those two functions to the `datadir.py` file and we're good to go!

j-gynn (approved):

PR #3564, opened 2025-06-10
[FEATURE] In-game ShortEvent Editing

scribblecrumb:
## About The Pull Request
This PR adds an in-game ShortEvent editor. The editor is only made accessible if the game config `dev_tools` is set to True.

Using this tool allows you to both edit existing events *and* add entirely new ones. It guides the user through the process and all possible parameters/options. Common errors are prevented by making them simply impossible to do (for example, the editor does not allow you to have both `child/parent` and `parent/child` tags in use at the same time.)  An event text preview is available to show the user how the text will look in game with replaced pronoun tags and names.  The buttons for existing events also displays the event's preview text within it's tool tip.

The vast majority of the editor's options are pulled from other scripts within the game, which means that as we change aspects of the game, the editor should remain accurate with minimal maintenance from us.  There are some attributes that are still unconnected from the game's scripts, but I foresee that being relatively simple to fix with some small adjustments (i.e. game.clan attributes aren't accessible without a clan loaded)

Some tabs lag the game when loading, but as this isn't exactly a player facing screen, I don't think it's unacceptable. This screen is so UI heavy compared to our usual screens, so while I set up in a modular way so that we never have all the UI elements showing/processing at once, it still impacts performance.  I did attempt to set up the moon skip loading screen and subsequently failed to really make it work. If someone with more threading knowledge than me wants a crack at, it would be great to get working!

There's also some fixes made to our ui_element classes for bugs encountered while building this screen.

There's still plenty of improvements to make for QoL, but I believe it's in a good enough state to be merged. More QoL updates can happen later as we get user feedback.

It makes the process of creating ShortEvents significantly easier for writers.  Rather than requiring them to learn how to edit json files directly (and dealing with all the issues that come with it), they can use this editor.  I'm sure modders will also enjoy it!

## Proof of Testing
https://discord.com/channels/1003759225522110524/1368423551392153600
Been using this ^ thread to show progress of the editor, there's videos in there that demonstrate it's usability.

## Changelog/Credits
Added an in-game ShortEvent editor accessible via the dev_tools game config.  This editor allows users to both edit existing events and create entirely new ones.   It guides the user through the process and all possible parameters/options. Common errors are prevented by making them simply impossible to do (for example, the editor does not allow you to have both `child/parent` and `parent/child` tags in use at the same time.)  An event text preview is available to show the user how the text will look in game with replaced pronoun tags and names.  The buttons for existing events also displays the event's preview text within it's tool tip.


scripts/game_structure/ui_elements.py
```diff
@@ -196,11 +199,14 @@ def show(self):
...
     def on_hovered(self):
         if self._is_tab and self.tab_movement["hovered"]:
+            self.find_text_layer_pos()
             self.text_layer.set_position(self.text_layer_active_offset)
         super().on_hovered()
 
     def on_unhovered(self):
         if self._is_tab and self.tab_movement["hovered"]:
+            if self._is_bottom_tab:
+                self.find_text_layer_pos()
```
j-gynn, line 209: i'm just gonna trust that there's a reason this has to be here

scripts/game_structure/windows.py
```diff
@@ -343,6 +344,142 @@ def process_event(self, event):
...
+        event_json = ujson.dumps(event_list, indent=4)
+        event_json = event_json.replace(
+            "\/", "/"
+        )  # ujson tries to escape "/", but doesn't end up doing a good job.
+
+        try:
+            with open(path, "w", encoding="utf-8") as write_file:
+                write_file.write(event_json)
+        except:
+            print(f"Something went wrong with event writing. Is {path} valid?")
```
j-gynn, line 416: can you not use `ujson.dump`? genuine question

scripts/game_structure/windows.py
```diff
@@ -343,6 +344,142 @@ def process_event(self, event):
...
+
+        super().__init__(
+            ui_scale(pygame.Rect((200, 200), (400, 200))),
+            window_display_title="Info Missing",
+            object_id="#editor_missing_info_window",
+            resizable=False,
+            always_on_top=True,
+        )
+
+        # adding a variable for starting_height to make sure that this menu is always on top
```
j-gynn: 😞 is this the only way? where are things displaying above this?

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6067 @@
...
+        },
+        {
+            "subject": "he",
+            "object": "him",
+            "poss": "his",
+            "inposs": "his",
+            "self": "himself",
+            "conju": 2
+        }
+    ]
```
j-gynn: localisation strikes again! it's 2am so i can't think what the right way of doing this is atm (something like pulling the defaults from the language files). poke me if you need help w/ it.

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6067 @@
...
+    """Pronoun dicts to assign to our test cats."""
+
+    all_camps: dict = {
+        "Forest": ["Classic", "Gully", "Grotto", "Lakeside"],
+        "Mountainous": ["Cliff", "Cavern", "Crystal River", "Ruins"],
+        "Plains": ["Grasslands", "Tunnels", "Wastelands"],
+        "Beach": ["Tidepools", "Tidal Cave", "Shipwreck", "Fjord"]
```
j-gynn: mmh not sure i like this being independently hardcoded

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6067 @@
...
+
+    event_types: dict = {
+        "death": ["murder", "old_age", "mass_death", "war"],
+        "injury": ["war"],
+        "misc": ["murder_reveal", "accessory", "ceremony", "war"],
+        "new_cat": ["war"]
+    }
+    """Dict with key as event type and value as allowed subtypes for that type."""
+
+    # TODO: consider moving some of these into a file that better facilitates new additions
```
j-gynn: definitely do this - JSON it! it makes it more accessible for modders too.

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6067 @@
...
+            "tag": "mentor/app",
+            "setting": False,
+            "conflict": ["mates",
+                         "not_mates",
+                         "child/parent",
+                         "app/mentor"]
+        }
+    ]
+    """List of dicts for relationship_values. Each dict holds tag name, conflicts, and setting."""
+    rel_value_types: list = RelationshipScreen.rel_value_names
```
j-gynn: possibly add a comment in `RelationshipScreen` on `rel_value_names` that indicates that changing or removing the variable will break this? tho that runs the same risk of being outdated...

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6067 @@
...
+        """The currenlty viewed editor tab."""
+
+        self.type_tab_buttons = {}
+        self.biome_tab_buttons = {}
+        self.event_buttons = {}
+
+        self.editor_element = {}
+
+        self.current_preview_state: int = self.preview_states[0]
+        """The currently used preview state. This can be 0 (preview off), 1 (plural), or 2 (singular"""
```
j-gynn:
```suggestion
        """The currently used preview state. This can be 0 (preview off), 1 (plural), or 2 (singular)"""
```

scripts/game_structure/ui_elements.py
```diff
@@ -536,6 +542,21 @@ def _get_scroll_bar_width(self) -> int:
...
+        """
+        any_hovered = False
+        for element in self:
+            if any(sub_element.hovered for sub_element in element.get_focus_set()):
+                any_hovered = True
+            elif isinstance(element, IContainerLikeInterface):
+                any_hovered = element.are_contents_hovered()
+            if any_hovered:
+                break
+        return any_hovered
```
j-gynn:
```suggestion
        for element in self:
            if any(sub_element.hovered for sub_element in element.get_focus_set()):
                return True
            elif isinstance(element, IContainerLikeInterface) and element.are_contents_hovered():
                return True
        return False
```
I think my indents are correct (and that this all checks out?)

scripts/game_structure/ui_elements.py
```diff
@@ -622,6 +643,42 @@ def set_visible_percentage(self, percentage: float):
...
+
+    def _check_for_scrolling_element(self, element, should_handle):
+
+        if isinstance(element, UIScrollingButtonList) and element.are_contents_hovered():
+            should_handle = False
+        elif isinstance(element, UIScrollingDropDown) and element.are_contents_hovered():
+            should_handle = False
+        else:
+            should_handle = True
+        return should_handle
```
j-gynn:
```suggestion
        def recursive_check_if_ignore(element):
            """
            If this is TRUE, we should ignore the scroll. This just helps with shortcutting
            :param element: The UIElement to check
            :return: True to ignore, False if we should care
            """
            if (
                isinstance(element, (UIScrollingDropDown, UIScrollingButtonList))
                and element.are_contents_hovered()
            ):
                return True
            elif isinstance(element, IContainerLikeInterface):
                for sub_element in element:
                    if recursive_check_if_ignore(sub_element):
                        return True
            return False

        # inverting the outcome of that
        if any(recursive_check_if_ignore(ele) for ele in self._container_to_scroll):
            return False
        else:
            return (
                self._container_to_scroll
                and self._container_to_scroll.are_contents_hovered()
            ) or self._check_is_focus_set_hovered()
```
I have tested this and I think it works as intended...

resources/lang/en/screens/event_edit.en.json
```diff
@@ -0,0 +1,91 @@
...
+        "empty_event_list": "There's no events for this type and biome.",
+        "event_text_initial": "Write the event's text here!",
+        "empty_event_id": "be sure to add an id!",
+        "valid_id": "valid ID",
+        "invalid_id": "invalid ID",
+        "dupe_id": "ID already in use",
+        "location_info": "<b>location:</b> Click to add or remove a place as a possible location. You only need to include the biome name if any camp is applicable.",
```
j-gynn:
```suggestion
        "location_info": "<b>location:</b> Click to add or remove a place as a possible location. Use the submenu to constrain permitted camps - by default, all camps are enabled.",
```

resources/lang/en/screens/event_edit.en.json
```diff
@@ -0,0 +1,91 @@
...
+        "rel_values_admiration": "admiration toward r_c is > than:",
+        "rel_values_trust": "trust toward r_c is > than:",
+        "litter": "Instead of being a single cat, this will be a litter of newborns/kittens.",
+        "meeting": "This cat/litter will continue living outside of the Clan.",
+        "exists": "If an existing outsider matches these constraints, they will be used instead of generating a new cat.",
+        "new_name": "This cat will take a new Clan-like name. If you wish to refer to their old name in the event text, use n_c_pre in the abbreviation instead of n_c.",
+        "old_name": "This cat will keep their old name. This will not work with litters, kittens, or newborns.",
+        "mate": "Cat will generate with the same age category as their first listed mate.",
+        "has_kits": "Cat will generate with an age within 14-120 moons.",
+        "can_birth": "If the same sex breeding setting is True, the cat will randomly choose between male and female. If it's False, this will generate a female."
```
j-gynn:
```suggestion
        "can_birth": "If the same sex breeding setting is True, the cat will randomly be either male or female. If it's False, this will generate a female."
```

resources/lang/en/screens/event_edit.en.json
```diff
@@ -0,0 +1,91 @@
...
+        "rel_values_comfortable": "comfortable toward r_c is > than:",
+        "rel_values_admiration": "admiration toward r_c is > than:",
+        "rel_values_trust": "trust toward r_c is > than:",
+        "litter": "Instead of being a single cat, this will be a litter of newborns/kittens.",
+        "meeting": "This cat/litter will continue living outside of the Clan.",
+        "exists": "If an existing outsider matches these constraints, they will be used instead of generating a new cat.",
+        "new_name": "This cat will take a new Clan-like name. If you wish to refer to their old name in the event text, use n_c_pre in the abbreviation instead of n_c.",
+        "old_name": "This cat will keep their old name. This will not work with litters, kittens, or newborns.",
+        "mate": "Cat will generate with the same age category as their first listed mate.",
+        "has_kits": "Cat will generate with an age within 14-120 moons.",
```
j-gynn:
```suggestion
        "has_kits": "Cat will be old enough to have kits (between 14 and 120 moons).",
```

scripts/screens/EventEdit.py
j-gynn, line 1: Doesn't follow our naming convention 😠

IDK if we want a different one so we can group dev tools, maybe `_DevEventEdit.py` ? It's just weird that this is the only screen that's not `XyzScreen`

scripts/screens/EventEdit.py
```diff
@@ -0,0 +1,6118 @@
...
+                        if event.ui_element == button and name != self.current_editor_tab:
+                            self.current_editor_tab = name
+                            self.clear_editor_tab()
+                            break
+
+            # TEXT PREVIEW
+            if event.ui_element == self.event_text_element["preview_button"]:
+                # finds what the new preview state should be
+                index = self.preview_states.index(self.current_preview_state)
+                new_index = index + 1 if index + 1 <= 2 else 0
```
j-gynn:
```suggestion
                new_index = (new_index + 1) % 3
```

j-gynn (changes requested):

scripts/screens/EventEdit.py
scribblecrumb: actually shouldn't break it! adding or removing a variable should seamlessly update the editor's options as well.  Admittedly this is untested, but I'm reasonably confident.

scripts/game_structure/ui_elements.py
scribblecrumb: if we don't do it, the bottom tab text layer will try to run away :pensive:

scripts/game_structure/windows.py
scribblecrumb: honest to god i do not know, this was already being used in the window class i copy pasted to use as a template for this one

scribblecrumb: Welp, removed it and nothing broke so :shrug:

scripts/screens/EventEdit.py
scribblecrumb: Shall we table this till after your `game` changes? Would love to also get it hooked up to the biome choosing screen, but that ain't relevant for this PR

scripts/screens/EventEdit.py
scribblecrumb: jsons made o7

scripts/screens/EventEditScreen.py
```diff
@@ -85,158 +67,18 @@ class EventEdit(Screens):
...
-    event_types: dict = {
-        "death": ["murder", "old_age", "mass_death", "war"],
-        "injury": ["war"],
-        "misc": ["murder_reveal", "accessory", "ceremony", "war"],
-        "new_cat": ["war"]
-    }
+    event_types: dict = TYPES
     """Dict with key as event type and value as allowed subtypes for that type."""
 
     # TODO: consider moving some of these into a file that better facilitates new additions
```
j-gynn:
```suggestion
```

scripts/screens/all_screens.py
```diff
@@ -7,7 +7,7 @@
 from .ClanScreen import ClanScreen
 from .ClanSettingsScreen import ClanSettingsScreen
 from .ClearingScreen import ClearingScreen
-from .EventEdit import EventEdit
+from .EventEditScreen import EventEdit
```
j-gynn:
```suggestion
from .EventEditScreen import EventEditScreen
```
can we get the class renamed too? 🥺
(don't do it from my suggestion, do it through IDE lol)

j-gynn (changes requested):
two tiny tweaks!

scripts/screens/all_screens.py
scribblecrumb: As you command o7

j-gynn (approved):
GO FORTH, AND EDIT THY EVENTS!

PR #3573, opened 2025-06-19
[ENHANCEMENT] Event Editor Improvements

scribblecrumb:
## About The Pull Request
Wow I just got the editor merged, surely I would take a break from it. NOPE

This PR is an assortment of bugfixes and improvements for the editor:
- New cat death tag added. Wild that this got missed, but in my defense, our documentation doesn't actually list this tag. Which is something else to fix at some point.
- Some inconsistencies with relationship value naming made it impossible to remove some rel tags from existing events. 
- I had automated the m_c `dies` param to match with the event type (if it's a `death` event, the m_c has to die!) but I forgot to make the display match the initial setting.
- I didn't think to have the location tag automatically flip to just the biome name if a user selected all the camp options. Now it does! Technically wouldn't break anything to leave it the other way, but it *would* make the json messier and harder to read.
- Event text editor is now clickable after entering editor, returning to main menu, and then re-entering the editor
- `some_clan` and `clan` options were forgotten, I've added them to the relevant option lists they're allowed in.
- `transition` subtype was forgotten, another thing we don't have in the documentation
- Event text no longer resets when the editor tab is switched
- `cats_from` and `cats_to` display now uses consistent terminology
- Changed how the rel_value text is determined to have easier future upkeep.

- Event validation upon saving. While the editor did already check that the event_id, text, and weighting were all included, it *didn't* check if our various dict blocks were valid (i.e. that you didn't forget to specify a cat for the injuries to happen to.) Now it does! If required info is missing, an alert message specifying what the issue is will display in a pop up when you attempt to save.
- Param locks!  Now you can click the little lock button that comes with each parameter to "lock" that parameter's info.  When you create a new event, everything will be cleared *except* those locked parameters.  This makes it easier to create many similar events one after the other.  However, if you open an existing event to edit it, all locks will reset to clear room for the existing event's information.
- The look of dark mode is improved. I totally forgot to check this before lol
- New cat block buttons now have a tooltip displaying their current tag list. This makes it easier to see at a glance what each new cat block is adding.
- There's now a search box for searching existing events by event_id! The search pool corresponds to the currently viewed existing events, so if you're viewing a specific type and biome, then it only searches within those parameters.
- New arrow icons now replace some placeholder assets!

Makin that editor better for everyone!

## Proof of Testing
video was too big for github, but it can be found here: https://discord.com/channels/1003759225522110524/1368423551392153600/1385343621712052254


j-gynn:
> Wild that this got missed, but in my defense, our documentation doesn't actually list this tag. Which is something else to fix at some point.

Could this get made as an issue please? Otherwise it'll get lost :3

scribblecrumb:
> Could this get made as an issue please? Otherwise it'll get lost :3

You got it!

scripts/screens/EventEditScreen.py
```diff
@@ -129,6 +131,23 @@ class EventEditScreen(Screens):
...
+    fatal_conditions: list = []
+    """We need this for death history validity checking. This is a list of all conditions that can kill."""
+    for condition in all_possible_injuries:
+        if Condition_Events.INJURIES.get(condition):
+            for age in Condition_Events.INJURIES[condition]["mortality"]:
+                if Condition_Events.INJURIES[condition]["mortality"][age]:
+                    fatal_conditions.append(condition)
+                    break
+                else:
+                    break
```
j-gynn, line 166: i think this should be `continue`...? Otherwise you're only checking mortality for a single age group.

scripts/screens/EventEditScreen.py
```diff
@@ -129,6 +131,23 @@ class EventEditScreen(Screens):
...
+        elif Condition_Events.ILLNESSES.get(condition):
+            for age in Condition_Events.ILLNESSES[condition]["mortality"]:
+                if Condition_Events.ILLNESSES[condition]["mortality"][age]:
+                    fatal_conditions.append(condition)
+                    break
+                else:
+                    break
```
j-gynn, line 173: Same here.

scripts/screens/EventEditScreen.py
```diff
@@ -151,12 +170,12 @@ class EventEditScreen(Screens):
...
-        "amount_up_high_button": Icon.ARROW_RIGHT,
-        "amount_down_low_button": Icon.ARROW_LEFT,
-        "amount_down_mid_button": Icon.ARROW_LEFT,
-        "amount_down_high_button": Icon.ARROW_LEFT
+        "amount_up_low_button": Icon.UP_LOW,
+        "amount_up_mid_button": Icon.UP_MID,
+        "amount_up_high_button": Icon.UP_HIGH,
+        "amount_down_low_button": Icon.DOWN_LOW,
+        "amount_down_mid_button": Icon.DOWN_MID,
+        "amount_down_high_button": Icon.DOWN_HIGH
```
j-gynn: Why does this exist?

scripts/screens/EventEditScreen.py
```diff
@@ -2419,15 +2696,15 @@ def update_relationships_block_options(self):
 
         # CATS
         self.relationships_element["cats_from_list"].set_selected_list(selected_constraints["cats_from"].copy())
-        self.relationships_element["cats_from_info"].set_text(f"selected: {selected_constraints['cats_from']}")
+        self.relationships_element["cats_from_info"].set_text(f"cats: {selected_constraints['cats_from']}")
```
j-gynn: Is this a player-facing string? If so, \*smacks u with newspaper\* localization! (love u)

j-gynn (changes requested):

scripts/screens/EventEditScreen.py
scribblecrumb: Because we have multiple places where we create amount buttons in this configuration.  Having this dict made it easy to standardize their naming and icon assignments.

scripts/screens/EventEditScreen.py
scribblecrumb: okay but consider this: player-facing but also code-adjacent.  i didn't localize any of the info display bc it generally just began with whatever the parameter is already called within the event dicts.  if any localizers are using this editor to write new events, leaving these parts un-localized isn't changing their experience at all compared to them just manually adding events within the jsons.  I *did* localize any instructions that we give, since that's something that would impact their experience far far more. 

also it was a headache to try and do a half localized text string. unless i was missing smth? it was not cooperating with me the few times i attempted it.

scripts/screens/EventEditScreen.py
scribblecrumb: I'm fairly certain we don't have any event-allowed conditions that only kill certain ages? And things are always most lethal with kits, so it seems like a most solid fall-back to me at the very least.  I don't think we have *anything* that wouldn't kill a kit but would kill an adult.

j-gynn (approved):
LGTM

PR #3582, opened 2025-06-22
Certain cats no longer think in 3rd person

Yzzuwyzzuf:
## About The Pull Request

Fixing a situation where cats are able to refer to themselves in 3rd person as a result of them being on both mc and rc sides of thought categories when they only were supposed to be on rc side

Fixes and issue players have been noticing. Not a bug fix per say, just removed few tags that should not have been applied where they were. Leaders will no longer have thoughts about themself in the 3rd person, same with deputies, medicine cats, and medicine cat apprentices as they have been removed from the Mc side

## Proof of Testing

Can't prove a lack of something but the minor changes in the files should speak for themselves

### DIFF
--- a/resources/lang/en/thoughts/alive/general.json
+++ b/resources/lang/en/thoughts/alive/general.json
@@ -207,7 +207,6 @@
             "warrior",
             "medicine cat",
             "mediator",
-            "deputy",
             "leader",
             "elder"
         ],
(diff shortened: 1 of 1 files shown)


archanyhm:
Please fix the PR title to something more meaningful; describe what kind of change this is (bugfix, content, etc) and then what this changes; feel free to take inspiration from other PRs.

Yzzuwyzzuf:
> Please fix the PR title to something more meaningful; describe what kind of change this is (bugfix, content, etc) and then what this changes; feel free to take inspiration from other PRs.

Fixed!

j-gynn (approved):

Issue #3566, opened 2025-06-14
[BUG] Thoughts are broken re: biome, camp & game mode


j-gynn:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

I *believe* that Thoughts don't correctly pull the player's camp, biome and game mode. These rely on `game.switches` variables that are never updated. Whilst there are only a few thoughts that rely on biome (and none that use camp or game mode as far as I can see), this should still be fixed.

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

Just search for `game.switches["biome"]` and you'll see what I mean. Same for `camp_bg` and `game_mode`.

### Commit # or Game Version Number

38d419c877496c0c3d85586d4a30d8e628edb6ba

scribblecrumb:
Did you still want to handle this @rustykitty ?

rustykitty:
> Did you still want to handle this [@rustykitty](https://github.com/rustykitty) ?

If someone else wishes to handle the issue they can go ahead.

PR #3586, opened 2025-06-22
fix biome, camp, gamemode constraints

rustykitty:
## About The Pull Request
Allowing biome, camp, gamemode thought constraints to be actually used.
Other than the original request, to not have thoughts() pull from game.switches, I've also fixed a capitalization issue with the biomes in the thoughts, updating the schema accordingly.

N/A

## Linked Issues
#3566

## Proof of Testing
Screenshot from 2025-06-22 11-40-49
Screenshot from 2025-06-22 11-40-56
Taken in a forest Clan.


j-gynn (changes requested):
Please merge in the latest changes from development!

Also, the rest of the content uses lowercase names (including for season, not just biome) - please make Thoughts match this.

scripts/clan.py
```diff
@@ -32,6 +32,7 @@
     quit,
     clan_symbol_sprite,
     get_living_clan_cat_count,
+    logger,
```
j-gynn: IDK if this is intended or a whoopsie, but we'd like each file to have its own `logger` (`logger = logging.get_logger("__name__")` if i recall the snippet correctly). that way the debug actually points to the right place

scripts/clan.py
```diff
@@ -831,6 +821,20 @@ def load_clan_json(self):
         )
         get_current_season()
 
+        for cat in Cat.all_cats.values():
+            try:
+                # initialization of thoughts
+                cat.thoughts()
+            except Exception as e:
```
j-gynn, line 747: Too broad an exception, can we be more specific what we're expecting to handle here?

scripts/clan.py
```diff
@@ -831,6 +821,20 @@ def load_clan_json(self):
...
+                logger.exception(
+                    f"There was an error when thoughts for cat #{cat} are created."
+                )
+                game.switches[
```
j-gynn: needs updating to the new switches format.

scripts/clan.py
```diff
@@ -831,6 +821,20 @@ def load_clan_json(self):
...
+                    "error_message"
+                ] = f"There was an error when thoughts for cat #{cat} are created."
+                game.switches["traceback"] = e
+                raise
```
j-gynn, line 756: eurgh a lot of clans are going to be crashing if this is added. like, it's not necessarily a bad thing to do! but we're gonna get some Spicy comments that clans ain't loading bcs we're missing coverage for dead cats in some way

scripts/game_structure/load_cat.py
```diff
@@ -280,19 +280,6 @@ def json_load():
...
-        except Exception as e:
-            logger.exception(
-                f"There was an error when thoughts for cat #{cat} are created."
-            )
-            game.switches[
-                "error_message"
-            ] = f"There was an error when thoughts for cat #{cat} are created."
-            game.switches["traceback"] = e
-            raise
-
```
j-gynn, line 362: why move this out of `load_cat`? /genq

j-gynn (changes requested):

scripts/clan.py
rustykitty: I just copied it over from earlier in the Clan load process.

scripts/game_structure/load_cat.py
rustykitty: Thoughts were moved into load_clan so they can be initialized *after* biome, camp, etc. are loaded.

scripts/clan.py
rustykitty: Sorry for forgetting about this! This was moved over from `load_cat` because the constraints will only work once the clan is loaded, so I don't think there will be issues.

scripts/clan.py
rustykitty: I think I've already done that.

rustykitty:
EDIT: nevermind, was on the wrong branch

@j-gynn Are you sure we should convert the constraints to lowercase? It looks like uppercase seasons are used consistently in our JSONs:
image

scripts/clan.py
archanyhm: Since this PR is more than a plain refactor, this should be fixed to catch specific errors or let the error propagate to a place where it can be handled sensibly.

rustykitty:
I think this broke after the merge. If I can't fix it, I'll open a new PR to redo the changes.

PR #3599, opened 2025-06-24
[UI] Update Role Icons

scribblecrumb:
## About The Pull Request
Updates the change role screen icons, as discussed!

Old icons weren't the most indicative of their roles, these are more representational in their symbolism

## Proof of Testing
image

## Changelog/Credits
- Change role screen rank icons have been updated!


j-gynn:
These are smaller than the original icons were - is this an intended change?

scribblecrumb:
Yup! It doesn't affect how they look on the screen. A lot of our old UI assets were essentially made twice as large as they actually needed to be.

j-gynn (approved):

PR #3601, opened 2025-06-24
Removing windows-2019 runner

j-gynn:
## About The Pull Request

Runner is deprecated as of the end of this month, and it's constantly failing all our tests and making me big mad. I want to hurl it into the sun.


archanyhm (approved):
Will serve as a temporary fix until we determine if we actually still need 32-bit builds.

PR #3605, opened 2025-06-25
[BUGFIX] Event Editor text editing

scribblecrumb:
## About The Pull Request
"t_inital" from the inital text liked to tag-a-long when text was being entered.  Now I just kill the initial text once the entry box is inputted.

## Proof of Testing
image
This is in the preview state, where "t_inital" would appear, as you can see, it isn't appearing.

## Changelog/Credits
- "t_initial" no longer invisibly appends itself to the event text entry


j-gynn (approved):
If it works, it ain't stupid.

PR #3606, opened 2025-06-25
[BUGFIX] event_edit reset

scribblecrumb:
## About The Pull Request
For the life of me, I don't know why `game.event_editing` refuses to reset, but here we go. It was already being reset in the EventEditScreen on exit_screen(), but now I've also got it resetting in the StartScreen screen_switches().  Surely it cannot evade this.

Also fixes a crash caused by clicking preview text without an event edit started.

ALSO fixes a crash caused by typing in the search bar.

## Proof of Testing
image
Futzed around in the event editor then timskipped through a clan, all the text adjust looks to be working.

## Changelog/Credits
- Using the event editor should no longer turn off all text adjust in other areas of the game.
- Clicking the preview button before starting an edit no longer crashes
- No longer crashes when using the search bar before the editor is opened


scripts/screens/EventEditScreen.py
```diff
@@ -863,7 +863,9 @@ def handle_event(self, event):
                             break
 
             # TEXT PREVIEW
-            if event.ui_element == self.event_text_element["preview_button"]:
+            if event.ui_element == self.event_text_element[
+                "preview_button"
+            ] and self.event_text_element.get("event_texta"):
```
Thlumyn: Is event_texta intentional vs event_text?

scribblecrumb: Good catch! That's indeed a typo.

j-gynn (approved):

PR #3610, opened 2025-06-26
[DEV TOOL] More event debug configs

scribblecrumb:
## About The Pull Request
This adds new game_config values to help with the testing of short events.  It's currently really hard to get an ensured event to actually occur, due to how much rng we have going on in the background for short events.  

New configs:
`debug_type_override`: you can use this to specify an event type to force. rng will essentially be ignored, every single cat will experience an event of the given type.  Setting this to "injury" also ignores the injury distribution code in order to force the event despite the injury's severity.
`debug_override_requirements`: If set to `true`, all constraints will be ignored and the event will be forced to occur.

Testing short events is a pain, this helps make it less of a pain.

## Proof of Testing
image
With requirements ignored, type "misc" forced, and an event id specified.

image
With type "injury" forced and event id specified.

image
With type "death" forced and event id specified.

image
With type "new_cat" forced and event id specified.

All of these tests were accomplished on the first timeskip! truly far more painless than it used to be.

## Changelog/Credits
Two new game configs for dev testing:
- `debug_type_override`: you can use this to specify an event type to force. rng will essentially be ignored, every single cat will experience an event of the given type.  Setting this to "injury" also ignores the injury distribution code in order to force the event despite the injury's severity.
- `debug_override_requirements`: If set to `true`, all constraints will be ignored and the event will be forced to occur.


j-gynn (approved):

j-gynn:
Fix conflicts & merge when ready!

PR #3616, opened 2025-06-29
[BUGFIX] FutureEvents crash

scribblecrumb:
## About The Pull Request
Ugh, why do the bugs always show up after you merge??
This crash was being caused by non-existent constraints in the involved cat's constraint dict.  Something that wasn't an issue before bc I swear I tested this.  Anyways, this fix just adds a check for the constraint in the given dict. Still a good addition, beyond the bug, since this means we don't spend time hopping into every constraint filter if no constraint was given.

## Proof of Testing
Crash was occurring when cats were murdered:
image
But after my fix, the murder goes off without a hitch.

## Changelog/Credits
- Murders no longer crash the game

### DIFF
--- a/scripts/events_module/event_filters.py
+++ b/scripts/events_module/event_filters.py
@@ -406,7 +406,7 @@ def cat_for_event(constraint_dict: dict, possible_cats: list, comparison_cat=Non
     # run funcs
     allowed_cats = []
     for param in func_dict:
-        if not constraint_dict.get(param):
+        if param not in constraint_dict:
             continue
         allowed_cats = func_dict[param](
             possible_cats, tuple(constraint_dict.get(param))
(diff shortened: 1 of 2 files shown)


scripts/events_module/event_filters.py
```diff
@@ -406,6 +406,8 @@ def cat_for_event(constraint_dict: dict, possible_cats: list, comparison_cat=Non
     # run funcs
     allowed_cats = []
     for param in func_dict:
+        if not constraint_dict.get(param):
```
larkgz:
```suggestion
        if param not in constraint_dict:
```

This is a bit more conventional.

larkgz (dismissed):

j-gynn (approved):

PR #3624, opened 2025-06-29
Partial hearing thoughts

j-gynn:
## About The Pull Request

Adds a sprinkling of new thoughts for partial hearing loss cats. Also one thought about having a "secret language" between two deaf cats was altered because it read as exoticizing. 

Adds a new feature to constrain thoughts with permanent conditions by whether or not they are congenital. Updates the documentation to reflect addition.

Some of the existing deaf thoughts were modified with this change in mind, creating a minor separation between pre-lingually deaf cats (i.e. congenital) and post-lingually deaf cats (i.e. acquired deafness). This change mostly means references to tail signals (the analogue for a signed language) are constrained to congenitally deaf cats, as most sign users are pre-lingually deaf. Few people learn an entirely new language if they go post-lingually deaf, instead preferring to rely on assistive devices like hearing aids and techniques like lipreading. Don't mind my autism kicking in, I just finished my dissertation on this subject lol

More thoughts are always good, and a call to action was posted for writers with relevant experience to add relevant thoughts.

## Proof of Testing
...this is really hard to evidence lmao. I stepped through with a debugger and confirmed that the new thoughts should be added to the pool when appropriate and discarded when not. They were never actually selected from the pool, but damn it I tried


SunlitFable (approved):
Can't speak for the coding, but the writing looks all good! Will wait for a coder to review those changes :]

scribblecrumb (approved):

PR #3628, opened 2025-06-30
Fix hover glitch

### LINKED ISSUE #3619: [UI] Textboxes overlapping some buttons
### Pre-requisites

### Description

Buttons in some Profile Screen boxes (such as the x button or the 'done' button when manually killing a leader) have a partially or mostly obscured hitbox which prevents user from clicking them easily

### Grade

1. Open cat profile
2. Open the change name window
3. 'x' can only be clicked if hovering on edges

1. Open leader profile
2. Open the kill cat window
3. 'Done' button cannot be clicked

_No response_

d5ce6ba1

No

_No response_

_No response_

### Additional context

Shrinking text boxes that are next to the effected buttons removes the error from what I've tried on source

### Screenshots

The x button can only be clicked if hovering over the very top or very right edge of it, hovering over the center of the button yields a text select mouse icon and clicking does nothing
Image

The same happens to the entire 'Done' button here, preventing pressing it at all, the x button has the same issue as above
Image

j-gynn:
## About The Pull Request

Fixing the strange hover glitches that have sprung up after the `uv` migration. There might be more, but this one eradicates the flash warning hover glitch on the event screen tabs. 

Fixes #3619 by making those items just not overlap any more
Fixes another bug I forgot to report where cats on the family tree screen weren't correctly showing their tooltip.

## Linked Issues
Fixes #3619

## Proof of Testing

https://github.com/user-attachments/assets/6c7f299d-23ff-4b3b-a2d3-139e7fa199b4
selecting tabs now has a normal amount of hover rather than rave edition (idk why this isn't embedding any more)

image
Hovering over cats in the family tree now shows tooltip

### DIFF
--- a/scripts/game_structure/ui_elements.py
+++ b/scripts/game_structure/ui_elements.py
@@ -885,6 +885,10 @@ class UISpriteButton:
         )
         del input_sprite
         self.button.join_focus_sets(self.image)
+        self.image.check_hover = self.__image_check_hover
+
+    def __image_check_hover(self, time_delta: float, hovered_higher_element: bool):
+        return False
 
     def return_cat_id(self):
         return self.button.return_cat_id()
--- a/scripts/game_structure/ui_elements.py
+++ b/scripts/game_structure/ui_elements.py
@@ -222,7 +222,8 @@ class UISurfaceImageButton(pygame_gui.elements.UIButton):
 
     def on_hovered(self):
         if self._is_tab and self.tab_movement["hovered"]:
-            self.find_text_layer_pos()
+            if self._is_bottom_tab:
+                self.find_text_layer_pos()
             self.text_layer.set_position(self.text_layer_active_offset)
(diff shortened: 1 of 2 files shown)


scribblecrumb:
Looks good, though I noticed that the right tabs are running into that offset issue (where hovering and unhovering cause the icons to slowly move up on the screen).  Nothing you did caused this, but we might as well fix it now.  Changing the `on_hovered` func to this fixed it:
```py
    def on_hovered(self):
        if self._is_tab and self.tab_movement["hovered"]:
            if self._is_bottom_tab:
                self.find_text_layer_pos()
            self.text_layer.set_position(self.text_layer_active_offset)
        super().on_hovered()
```        
I'm finding it very funny that I previously fixed this problem on the left tab buttons, but somehow fixing those ones did not fix the right tabs lmao

scribblecrumb (approved):

PR #3630, opened 2025-06-30
[BUGFIXES] Event Editor

scribblecrumb:
## About The Pull Request
My struggle never ends.

- Added some extra info checks to the event dict compiler to hopefully clean up empty parameters more.
- Fixes an issue with tags not clearing properly
- Fixes an issue with rel_status param locks for both main and random cat.
- Fixes the odd string sneaking in where ints should be

## Proof of Testing
image
Tags were previously being carried over from event to event, even when they weren't supposed to be.  I clicked through a bunch of events, looking at the tags and even checking some at random to see if their settings would be pulled to the next event. It seems to be clearing the info correctly now!


j-gynn (approved):

j-gynn:
Just as an aside, can you please use new branches for new PRs? The squash commits keep adding previous commits to the log of what's been added and it makes it hard to track what was actually added this time.

scribblecrumb:
> Just as an aside, can you please use new branches for new PRs? The squash commits keep adding previous commits to the log of what's been added and it makes it hard to track what was actually added this time.

Ugh yeah, I noticed that. my days of reusing branches are over, sadge

PR #3633, opened 2025-07-01
[DEV] New filtering system for ShortEvents

scribblecrumb:
## About The Pull Request
This changes how weighting and random_cat assigning operates!

What we prior knew as "weight" has been changed to Frequency, which has a 1-4 scale. 
1. Very Rare
2. Rare
3. Uncommon
4. Common

With a contextual understanding of "An event like this one should happen x times in the span of 10 moons".  A big struggle we have with our old weight system is a lack of real context for writers to gauge how to properly weight their event. This hopefully makes it a bit easier!

It's still a bit of a perspective shift to get used to, as you do need to divorce yourself from the idea of "well a cat shouldn't die every single moon".  This isn't a "how often should injury/death events occur" rarity. It's a "if this type of event happened every moon, how often should you get one like this?"  
I've updated the documentation of ShortEvent to reflect this change and give extra clarification.

How this works under the hood is a simple 1-10 randint.  If the int is 1-4, the event will be common. 5-7, uncommon. 8-9, rare, and 10 is very rare.  This is essentially replicating the "within the span of 10 moons" idea.

Once we've picked what frequency our event needs to be, we'll gather and filter our possible events. If no events are possible with that frequency , we'll move to the next highest frequency (so from rare to uncommon, for example), and so on and so forth. If no events are found by the time we get through the whole scale, then no event occurs.

I've removed the old injury distribution system that essentially worked as another limiter on what events were available. It was meant to ensure cats were getting injuries at rates that made sense for their status, however I feel that it's redundant with the new frequency system.

I mostly updated the jsons to the new frequency system via search and replace, but i did do a skim to manually adjust things that seemed off.  I think it's fine for now, we can adjust as we need to if any events are weirdly common (something else that is much easier to do now!)

This also serves to chunk possible events into smaller lists! I've adjusted `loaded_events` to save lists according to both file path and rarity.

So why the change from "weight" to "frequency"?  Well, I've added an actual weight system that determines the weight automatically during event object generation.  
- `+1` If the event has a specific location
- `+1-3` If the event has a specific season (number is the inverse of the given season list length)
- `+` a variable amount for each cat param given.  These are mostly inverse amounts according to the total options available.  For example, the trait param can add up to 54 to the weight. Of course, that would only be if one single trait was required for the event to trigger.  If, perhaps, there were 10 possible traits for the event, then it would add 44 to the weight.

The ending weight number dictates how many copies of that event will be added to the event list (given, of course, that the event passes all non-cat and m_c filtering.)  During testing this was usually around 2-4, as many of our events have few constraints.  However there were definitely moments where copies would reach into the 70s and sometimes the total event list len would be 700+.

This helps to make the higher constraint events more common without writers having to do any math.  We could even see about giving specific parameters a higher or lower weight impact if we end up feeling like that's warranted.  Traits, admittedly, have the largest weight out of any parameter, simply due to how many traits we have in the game.  We might consider debuffing them a bit, however I'd prefer to just see writers adding in more events with tight constraints.

Now what about the random_cat!  I've changed it so that we're doing exactly the opposite of what we used to.  Rather than picking a cat at random before filtering and trying to find events that fit both main and random cat, now we find all events that fit main cat *then* we see if we can find a random cat.

What occurs is this:
- main cat rng's for an event
- frequency is chosen for the event
- main cat is checked against available events and a new list is created out of the events that pass
- copies of events are added according to their weight
- an event is chosen *at random* from those passing events
- *if* the chosen event needs a second cat, we search through the clan at random to see if any cats fit the constraints of the event
- if a cat does fit, then we use that cat for the chosen event and the event occurs as normal
- if a cat does not fit, we randomly choose a new event and repeat the process until either an event without a second cat is chosen, or until a cat fits the constraints

I've done my best not to make this performance hell and while testing with a 40 cat clan, I actually didn't see random_cat searches happening very often. Typically there would be one or two each moon, generally with a cat found right away.  I also didn't see a big performance change to timeskipping, but tbf my computer rarely struggles with clangen.  I've also ensured that when an event fails to find a cat, we add it's `event_id` to a list of failed ids. Any consecutive copies of that event that get chosen will be discarded immediately without going through the cat search, this way the weighting system isn't bogging us down as much as it could.  I did consider just removing all the events with matching IDs from the list, but honestly I'm not sure if that will be more performant or not?

It's good to keep in mind that this removes the prior weighting system that would just create 20-some copies of the event. I think even with the variable weighting we now have, we're still working with smaller lists and less objects overall.

This changes weighting to have a fixed context for writers and a more reliable effect on the available events.  It also removes the effect that random cats would naturally have on what events were available! After all, with our old system it doesn't matter if a cat has a sibling that could be in a sibling event with them. If the random cat chosen isn't a sibling, then that event won't even be a possibility.  Now, the random cat's identity has no bearing on what events are available!

It also adds an under-the-hood weighting system that doesn't depend on us to manually decide an events weight, but still allows us control in how heavy certain parameters should be considered.

In my time testing I've noticed a much nicer range of events occurring, with events I usually saw very rarely become much more common! They were only rare to begin with because they were being drowned out or they were unnecessarily limited by their random cat constraints.

This also opens up the opportunity for short events to feature more than just two cats, a limitation we've had up till now. I don't think it's something we want to get crazy with (like.. 3 or 4 cats **max**), but that limitation has been a problem in the past.

## Proof of Testing
image
Documentation update

image
A look at the print tests I set up (that aren't included in this PR, ur welcome arch) The 'no cat needed' and 'cat found' messages mark when an event was chosen. You can even see how the weight was incremented up when 0 events were available at lower weights.  My prints also included messages for when cats weren't found, but you don't see any here because a cat was found the first time it was searched for. (fyi, these were done before i changed the name to frequency )

image
And here is the two cat event from that print statement! An entirely appropriate cat was found.

## Changelog/Credits
- The way short events are weighted has changed to be more intuitive and accurate.
- The way random cats are found for short events has changed to be less limiting to the type of events that are available.


j-gynn:
I'm not seeing schema updates to reflect the new name for weight...?

scribblecrumb:
Oof, good catch, I've added that!

scripts/screens/EventEditScreen.py
```diff
@@ -1545,9 +1545,9 @@ def clear_event_info(self):
         self.rank_tag_checkbox = {}
         if not self.param_locks.get("tag"):
             self.tag_info = []
-        self.weight_element = {}
-        if not self.param_locks.get("weight"):
-            self.weight_info = 20
+        self.rarity_element = {}
+        if not self.param_locks.get("rarity"):
+            self.rarity_info = 20
```
j-gynn: pick a rarity, but not 20

scripts/screens/EventEditScreen.py
```diff
@@ -274,9 +274,9 @@ def __init__(self, name=None):
         self.tag_info: list = []
         """Loaded tags"""
 
-        self.weight_element = {}
-        self.weight_info: int = 20
-        """Loaded weight"""
+        self.rarity_element = {}
+        self.rarity_info: int = 20
```
j-gynn: what's the new default?

scribblecrumb: 4!  Darn, thought I got all the 20s lmao

scripts/events_module/short/short_event.py
```diff
@@ -69,22 +102,34 @@ def __init__(
...
             if "not_trait" not in self.r_c:
                 self.r_c["not_trait"] = []
-            if "age" not in self.r_c:
-                self.r_c["age"] = []
+            else:
+                self.weight += len(self.r_c["not_trait"])
             if "backstory" not in self.r_c:
                 self.r_c["backstory"] = []
+            else:
+                self.weight += 1
```
j-gynn: if we have both `if not a` and `else` checks, can we get these swapped around so that it's `if a` and `else`? just for logical flow.

scripts/events_module/generate_events.py
```diff
@@ -131,13 +126,17 @@ def generate_short_events(event_triggered, biome):
...
+                        continue
+
                     event = ShortEvent(
                         event_id=event["event_id"] if "event_id" in event else "",
                         location=event["location"] if "location" in event else ["any"],
                         season=event["season"] if "season" in event else ["any"],
                         sub_type=event["sub_type"] if "sub_type" in event else [],
                         tags=event["tags"] if "tags" in event else [],
-                        weight=event["weight"] if "weight" in event else 20,
+                        frequency=event["frequency"] if "frequency" in event else 4,
```
j-gynn:
```suggestion
                        frequency=event_frequency,
```
we've already determined this on line `119`, should just use `event_frequency`

also, if we've already discarded non-matching frequencies, why do we need to add this to the ShortEvent item?

scripts/events_module/generate_events.py
```diff
@@ -221,7 +220,7 @@ def generate_ongoing_events(event_type, biome, specific_event=None):
                 return event
 
     @staticmethod
-    def possible_short_events(event_type=None):
+    def possible_short_events(event_type=None, frequency=4):
```
j-gynn: why provide a default frequency? Surely it would be better to force whatever is calling this to ensure it has requested a rarity.

scripts/events_module/generate_events.py
```diff
@@ -292,14 +295,10 @@ def filter_possible_short_events(
...
-            if (
-                event.event_id
-                == constants.CONFIG["event_generation"]["debug_ensure_event_id"]
-                and constants.CONFIG["event_generation"]["debug_override_requirements"]
-            ):
+            # if requirements are overridden, allow event through
+            if constants.CONFIG["event_generation"]["debug_override_requirements"]:
                 final_events.append(event)
-                break
+                continue
```
j-gynn, line 303: if we've appended it to the list, we shouldn't continue otherwise we'll add it twice, no?

scripts/events_module/generate_events.py
```diff
@@ -484,20 +432,82 @@ def filter_possible_short_events(
                 if discard:
                     continue
 
-            # ensure ID without requirements override
+            final_events.extend([event] * event.weight)
```
j-gynn, line 437: thought `weight` had been replaced...?

scripts/events_module/generate_events.py
```diff
@@ -484,20 +432,82 @@ def filter_possible_short_events(
...
+                for event in final_events:
+                    if (
+                        event.event_id
+                        == constants.CONFIG["event_generation"]["debug_ensure_event_id"]
+                    ):
+                        chosen_event = event
+                        break
+            # else, pick a random one from the available events
+            elif not chosen_event:
+                chosen_event = random.choice(final_events) if final_events else None
```
j-gynn: i feel like this check for if `final_events` exists should happen much, much earlier.

scripts/events_module/generate_events.py
```diff
@@ -484,20 +432,82 @@ def filter_possible_short_events(
...
+            if constants.CONFIG["event_generation"]["debug_ensure_event_id"]:
+                for event in final_events:
+                    if (
+                        event.event_id
+                        == constants.CONFIG["event_generation"]["debug_ensure_event_id"]
+                    ):
+                        chosen_event = event
+                        break
+            # else, pick a random one from the available events
+            elif not chosen_event:
```
j-gynn, line 459: how can an event already be chosen? /genq

scripts/events_module/short/handle_short_events.py
```diff
@@ -133,56 +133,53 @@ def handle_event(
...
-            sub_types=self.sub_types,
-            allowed_events=self.allowed_events,
-            excluded_events=self.excluded_events,
-            ignore_subtyping=ignore_subtyping,
-        )
+        # choosing frequency
+        # think of it as "in a span of 10 moons, in how many moons should this sort of event appear?"
+        frequency_roll = randint(1, 10)
+        if frequency_roll <= 4:
+            weight = 4
```
j-gynn: rename to `frequency` to align with all the others please, this is so confusing

it *is* doing the same thing as `frequency`, right?

scripts/events_module/short/handle_short_events.py
```diff
@@ -133,56 +133,53 @@ def handle_event(
...
+                other_clan=self.other_clan,
+                freshkill_active=FRESHKILL_EVENT_ACTIVE,
+                freshkill_trigger_factor=FRESHKILL_EVENT_TRIGGER_FACTOR,
+                sub_types=self.sub_types,
+                allowed_events=self.allowed_events,
+                excluded_events=self.excluded_events,
+                ignore_subtyping=ignore_subtyping,
+            )
+            if not chosen_event:
+                # we'll see if any higher weighted events are available
```
```suggestion
                # we'll see if any more common events are available
```
just for my own brain lol

scripts/events_module/short/short_event.py
```diff
@@ -1,20 +1,34 @@
...
     A moon event that only affects the moon it was triggered on.  Can involve two cats directly and be restricted by various constraints.
     - full documentation available on GitHub wiki
     """
 
+    num_of_traits = len(Personality.trait_ranges["normal_traits"].keys()) + len(
+        Personality.trait_ranges["kit_traits"].keys()
+    )
+    num_of_skills = len([*SkillPath])
+
+    num_of_ages = 7  # can't pull this from the Cat file bc of circular imports
```
j-gynn: Can't use `len(CatAge)`?

scripts/events_module/short/short_event.py
```diff
@@ -1,20 +1,34 @@
...
+
+    num_of_ranks = 11
```
j-gynn: Could add a helper function to `CatRank`? Don't think that was merged in when this PR was made, so now you have your own previous work to help!

scripts/events_module/short/short_event.py
```diff
@@ -32,33 +46,56 @@ def __init__(
     ):
         if not event_id:
             print("WARNING: moon event has no event_id")
+
+        self.weight = weight
```
j-gynn: Won't events with pre-existing weights go stratospheric with the auto-weighting? Because all others start at 1...?

scripts/events_module/short/short_event.py
```diff
@@ -1,20 +1,34 @@
...
 class ShortEvent:
     """
     A moon event that only affects the moon it was triggered on.  Can involve two cats directly and be restricted by various constraints.
     - full documentation available on GitHub wiki
     """
 
+    num_of_traits = len(Personality.trait_ranges["normal_traits"].keys()) + len(
+        Personality.trait_ranges["kit_traits"].keys()
+    )
+    num_of_skills = len([*SkillPath])
```
j-gynn:
```suggestion
    num_of_skills = len(SkillPath)
```
Is there a reason this doesn't work? /genq

scripts/screens/EventEditScreen.py
```diff
@@ -274,9 +274,9 @@ def __init__(self, name=None):
         self.tag_info: list = []
         """Loaded tags"""
 
-        self.weight_element = {}
-        self.weight_info: int = 20
-        """Loaded weight"""
+        self.frequency_element = {}
+        self.frequency_info: int = 20
```
j-gynn: defaulting to 20???

j-gynn (changes requested):

j-gynn:
I haven't checked a single one of these events for frequency values to see if they make sense. I don't hate myself enough to try and do that through GitHub's PR viewing framework. Just as a heads-up.

scripts/events_module/generate_events.py
scribblecrumb: Ah, good point

scripts/events_module/generate_events.py
scribblecrumb: continue moves to the next event, so no it shouldn't append twice? am i missing something? /genq

scripts/events_module/generate_events.py
scribblecrumb: It has in that we no longer have it determined by the writers, but it still functions the same in that it adds copies of the event.  Is there a different weighting method you'd like me to use other than copying?

scripts/events_module/generate_events.py
j-gynn: hELP i'm mixing up with `pass`

scripts/events_module/generate_events.py
j-gynn: No, no... I was just surprised to see it here.

scripts/events_module/generate_events.py
scribblecrumb: lmaooo you had me over here like "have i fundamentally misunderstood continue?"

scripts/events_module/short/handle_short_events.py
scribblecrumb: yup, sorry, i've had to rename these twice lmao i'm not surprised i missed a couple

scripts/events_module/short/short_event.py
scribblecrumb: We have no events left with pre-existing weight cus i've replaced them all with frequency, so no.  but regardless that does mean we don't need the weight arg here at all.

scripts/events_module/short/handle_short_events.py
j-gynn: don't sweat it!

scribblecrumb:
I gave them a skim when I first reassigned the frequencies, so it's at least nominally sensible and that's good enough for me.  We may want to think about making some sort of announcement to the beta testers (or? if we dare... the general playerbase) about reporting if they feel an event is too common.  but that may be unwise lmao, if we do then we should prob also make a master issue for it like we have for typos

j-gynn (approved):
Code side looks good. I leave it to you as to whether you'd like to assemble the ~~minions~~ writers to verify the sanity of your new frequencies. If not, merge away! Otherwise, minions assemble~

PR #3634, opened 2025-07-02
Add seizure prone thoughts

dinkwe:
## About The Pull Request

Adds 8 new thoughts for the permanent condition, seizure prone. Of note, while I have seziures, I do not have epilepsy, so while I tried to be general with my seizure-related descriptions, I am just one guy.

Part of the project to add more specific throughts for disabilities/permanent conditions

## Proof of Testing
image

## Changelog/Credits

credits to dinofelisdruid !!

### DIFF
--- a/resources/lang/en/thoughts/alive/general.json
+++ b/resources/lang/en/thoughts/alive/general.json
@@ -6114,5 +6114,40 @@
                 "severe headache"
             ]
         }
+    },
+    {
+        "id": "seizure_prone_gen",
+        "thoughts": [
+            "Is bothered by the sunlight flickering through the trees",
+            "Makes sure to drink water before a training session",
+            "Wonders what {PRONOUN/m_c/subject} {VERB/look/looks} like during a seizure",
+            "Feels like the sun is too bright today",
+            "Feels dizzier than usual today"
+        ],
+        "perm_conditions": {
+            "m_c": [
+                "seizure prone"
+            ]
+        },
+        "random_living_status": [
+            "living"
+        ]
+    },
+    {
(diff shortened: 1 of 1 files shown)


resources/lang/en/thoughts/alive/general.json
```diff
@@ -6114,5 +6114,40 @@
...
+            ]
+        },
+        "random_living_status": [
+            "living"
+        ]
+    },
+    {
+        "id": "seizure_prone_gen_to_alive_gen",
+        "thoughts": [
+            "Tries to explain what a seizure feels like to r_c, but only confuses {PRONOUN/r_c/object} more",
```
j-gynn, line 6139: I think this one is a bit long? Could you possibly split this into just the seizure prone cat "tries to explain what a seizure feels like to r_c" and another from alive gen to seizure prone cat that's like "Is confused by r_c's explanation of what seizures feel like"? It's also nice to have more thoughts where the seizure cat is the "target" but isn't explicitly being handled by a medcat?

j-gynn (changes requested):
other than my one comment, this looks good!

resources/lang/en/thoughts/alive/general.json
dinkwe: It's technically under the character limit I was assigned and I'd rather keep it to one thought because it's one cohesive idea in my mind.

Would "Tries to explain what a seizure feels like, but just confuses r_c" work?

j-gynn: Ah, don't worry. It's fine as it is, it was only a suggestion.

j-gynn (approved):

Issue #2430, opened 2024-06-10
[CODE] Clan-wide relationship changes based on temperament


scribblecrumb:
**Description**
It could be useful to have a way to specify clan-wide relationship changes based on the temperament of the clan as a whole.  For example, a bloodthirsty clan would likely respond differently to the killing of a loner than a gracious clan would.  We can already cause relationships to change on a clan-wide scale in Patrols and ShortEvents, this addition could expand those rel change `cats_to` `cats_from` tags to include `clan_{temperament}` options.

j-gynn:
Closed by #3637

PR #3637, opened 2025-07-03
[FEATURE] Facet tags for cat object gathering

scribblecrumb:
## About The Pull Request
This adds new tags that can be used within any ShortEvent and Patrol parameter block that needs a cat list inputted, such as the injury block or relationships block.  These will likely see the most use in just relationship blocks, though.

These tags allow you to constrain the cats being gathered by a certain personality facet value.  This means it does need to be used in conjunction with other cat tags, typically group tags like `clan` or `patrol`. You can also use multiple tags to constrain across multiple facet ranges.  There is a `low` and `high` version for each facet, with low covering the 0-8 range and high covering the 9-16 range.

Tags are as follows:
- `low_lawful`
- `high_lawful`
- `low_social`
- `high_social`
- `low_stable`
- `high_stable`
- `low_aggress`
- `high_aggress`

ShortEvent and Patrol documentation has been updates, as has the Event Editor.

I've also added these tags to some of our existing ShortEvents, wherever it felt applicable. Murder reveals that feature a clan-wide reveal, breaking the warrior code, fraternizing with twolegs/outsiders, ect.

This makes it far easier for us to apply wide changes across groups of cats in a more controlled manner. Previously, if we wanted a group to have varying reactions, we had to just rely on choosing a random selection of cats within a group. 

I think this also satisfies #2430, it's a slightly different application, but this is a better implementation imo

## Proof of Testing
image
Updated documentation.

image
image
image
Cinnamon broke the warrior code! You can see this now gets added to the relationship log. I just had this event pull down the respect, trust, and comfort the other cat has for Cinnamon, you can see that those values are mostly gone now.

Across the Clan, only Robinstripe and Moonflare had a relationship change from this event.  Robinstripe is adventurous and Moonflare and thoughtful, which both fit into the `high_lawful` `high_stable` constraint I gave this event.

## Changelog/Credits
- Changes across groups of cats can now be done according to each cat's personality facets.
- Some moon events will now apply more specific relationship reactions across groups of cats. For example, if a cat is seen taking food from a Twoleg, their more lawful Clanmates will have a negative reaction  (my favorite event change is with the murder reveal where the murderer tells the whole clan that they had to do it, that the cat they murdered was plotting against them. lawful clanmates may not like that, but more aggressive clanmates might find it justifiable)

### DIFF
--- a/docs/dev/writing/patrols.md
+++ b/docs/dev/writing/patrols.md
@@ -782,6 +782,20 @@ What each parameter does, and what the options are for outcomes.
 | "clan"        | Feelings toward the entire clan are effected                              |
 | "n_c:{index}" | Feelings toward the new cat(s) with the index number {index} are effected |
 
+> Group modifiers: These will modify the cats already being gathered according to the other strings. For example, a block with `"cats_from": ["clan", "low_lawful"]` will gather all the cats in the Clan with a 0-8 lawfulness facet.  These can be combined to get cats with specific ranges of multiple facets.
+
+| modifier     |                                                |
+|--------------|------------------------------------------------|
+| low_lawful   | cats with a 0-8 lawfulness facet are affected  |
+| high_lawful  | cats with a 9-16 lawfulness facet are affected |
+| low_social   | cats with a 0-8 sociable facet are affected    |
+| high_social  | cats with a 9-16 sociable facet are affected   |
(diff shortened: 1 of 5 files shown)


j-gynn (approved):
Think this looks good to me, nice!

PR #3638, opened 2025-07-03
[BUGFIX (kind of?)] Removed Events Potentially Causing Crashes

LaurBee:
## About The Pull Request
Have started getting crashes related to the freshkill pile. Unsure if somehow related to my recently added freshkill events or just bad luck, but am removing said events just to be safe


scribblecrumb:
Could you make an issue with the crash log as well? It might be a simple fix.

LaurBee:
> Could you make an issue with the crash log as well? It might be a simple fix.

Will do, just let me see if I can replicate it again

PR #3657, opened 2025-07-07
[BUGFIX] New Cat creation bugs

### LINKED ISSUE #3654: [BUG] outsider cats from patrols go to starclan when passing away
### Pre-requisites

### Description

i encountered a patrol (gen_med_rescue10). a loner died and three of her kits joined the clan. i moonskipped, and she received a clan burial/death message, and i noticed she ended up in starclan, not the unknown residence. her name is sleepy

### Grade

1. make a clan
2. send out border patrols till an outsider is killed or found dead
3. check starclan
4. moonskip to see the clan burial text

PATROL ID: gen_med_rescue10 | SUCCESS: True

probably occurs w/ every patrol in which an outsider dies

e569c312

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

scribblecrumb:
## About The Pull Request
New kits found on patrol were sometimes generated with kittypet name, despite their new clan status.  This is because previously we never tried to preserve any history of *what* the kit was originally born as, beyond their backstory.  Thus, we never bothered to specify kittypet/loner/ect. in their attribute lists.  Without that specification, they were randomly choosing between all the CatSocials available, then being generated as that social along with an appropriate name for it, then skipping the name change due to our prior assumption of an initial clan name generation.

All that to say, I've now created three new backstory categories to categorize which kitten backstories align with each CatSocial.  Now we'll just check what backstory was chosen from their attribute list and assign a CatSocial according to that.  This honestly isn't *perfect* and we might find kits being generated as initally loners when their parent was a kittypet, and vice versa.  However, the way to fix that is to go through all our new_cat events and add the relevant social tag to every litter attr list.  I don't particularly want to do that and I don't think it matters that much in the long run, this is a pretty small thing to over-complicate and I already feel like I've over-complicated it lmao

We also weren't checking for alive status before adding to the clan, resulting in dead outsiders getting into starclan.  I've added that check now.

Fixes #3654 

## Proof of Testing
image
They've got correct names!  I also jumped through the debugger to see that things were behaving as expected behind the scenes, as they do seem to be.

image
image
Candi didn't make it into cat heaven, as intended

## Changelog/Credits
- Litters will no longer be found with ill-fitting names (rip empresskit)
- Dead outsiders will no longer generate into StarClan

### DIFF
--- a/resources/dicts/backstories.json
+++ b/resources/dicts/backstories.json
@@ -9,6 +9,15 @@
         "outsider_roots_backstories": [
             "outsider_roots1", "outsider_roots2"
         ],
+        "baby_clancat_backstories": [
+            "halfclan1", "halfclan2", "outsider_roots1", "abandoned3"
+        ],
+        "baby_loner_backstories": [
+            "outsider_roots2", "abandoned1", "abandoned2"
+        ],
+        "baby_kittypet_backstories": [
+            "abandoned4"
+        ],
         "half_clan_backstories": [
             "halfclan1", "halfclan2"
         ],
--- a/scripts/utility.py
+++ b/scripts/utility.py
@@ -538,6 +538,22 @@ def create_new_cat_block(
     if bs_override:
         chosen_backstory = choice(stor)
 
+        if (
+            chosen_backstory
+            in BACKSTORIES["backstory_categories"]["baby_clancat_backstories"]
+        ):
+            cat_social = CatSocial.CLANCAT
+        elif (
+            chosen_backstory
+            in BACKSTORIES["backstory_categories"]["baby_loner_backstories"]
+        ):
+            cat_social = CatSocial.LONER
+        elif (
+            chosen_backstory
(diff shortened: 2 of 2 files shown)


j-gynn:
When a litter is generated, they can be tagged with parent labels right? Can't the kits just parasitize the parent's rank, if one exists?

j-gynn (approved):

scribblecrumb:
> When a litter is generated, they can be tagged with parent labels right? Can't the kits just parasitize the parent's rank, if one exists?

I did think about doing that, however it got dicey when I started implementing it.  It just felt too unsure?  Generated cats can be given a parent1 or parent2, or both.  If they have both, how do I decide which to parasitize? Besides that, I'd have to pull the social out of their history, because at this point they've been fully generated and may have joined the clan, thus becoming a CLANCAT.  Not particularly difficult, but just further complicates the matter when handling two parent cats.  

In the end, most of these events do specify a social for the parent and also gives the kits a backstory that matches with that parent's social.  So I felt that it would be simplest to play off that.

Issue #3391, opened 2025-01-30
[CODE] murder reveal text in history is not localized


rustykitty:
**Description**
Pretty much what the title says. I am talking about this text, for example, and a few others in the same function: 
https://github.com/ClanGenOfficial/clangen/blob/5b0f4461bf8bbae278b05bb987388f8f645c750c/scripts/cat/history.py#L675

scribblecrumb:
#3662 addressed this

PR #3662, opened 2025-07-08
[ENHANCEMENT] Murder history changes

scribblecrumb:
## About The Pull Request
This adjusts how we save murder histories to better align with future planned expansion.

I've cut out the extraneous information, which has helped clean up the retrieval of these histories as well.  I've also adjusted their related staticmethods into functions that should be called from the murderer's history.  We now save a bool regarding general clan knowledge of the act as well as individual cat IDs for cats who are aware of the act.  This list means that now we could have the secret spread amongst multiple individuals before it becomes common knowledge.  This is subsequently shown in the murderer and victim's histories, with a True clan_aware overriding the display of aware individuals.

I've also added a function to history specifically for the retrieval of the profile history string!

Additionally, I knocked out some bugs in future events, seeing as this deals heavily with the triggering of those events.  Turns out we weren't removing successfully triggered events from the future event list, which meant murders were being revealed over and over every moon.  So that's fixed now, albeit a little bit jankily, but I promise I have a better implementation in a different PR (that makes other changes to future events that didn't feel appropriate for this PR)
  

Sets us up for future plans of murderer judgement and greatly cleans up the tangle of murder/victim history code.

Addresses #3391

## Proof of Testing
image
Murderer history successfully given
image
Victim history successfully given

image
image
A murderer's history updated when revealed to Clan
image
image
vs. when revealed to an individual

image
image
A victim's history updated when revealed to Clan
image
image
vs. when revealed to an individual

(to note, these screenshots were gathered across multiple murders because i was very inefficient lmao, so don't mind ID's seeming inconsistent w murderer and victim)

## Changelog/Credits
- Future events are now properly removed from the queue if successfully triggered
- Murderer and Victim histories have been adjusted to a new format.  Clan knowledge and individual knowledge of a murder is now saved and displayed in profile histories.


scripts/cat/history.py
```diff
@@ -41,6 +43,20 @@ def __init__(
             self.mentor_influence["skill"] = {}
         if "mentor" in self.mentor_influence:
             del self.mentor_influence["mentor"]
+        # converting old murder saves
+        if self.murder:
+            for killed in self.murder.get("is_murderer", []):
+                if isinstance(killed["revealed"], bool):
+                    new_dict = {"to_clan": False, "aware_individuals": []}
+                    if killed["revealed"]:
+                        new_dict["to_clan"] = True
```
j-gynn:
```suggestion
                    new_dict = {"to_clan": killed["revealed"], "aware_individuals": []}
```

scripts/cat/history.py
```diff
@@ -41,6 +43,20 @@ def __init__(
...
+                    killed["revealed"] = new_dict
+            for death in self.murder.get("is_victim", []):
+                if isinstance(death["revealed"], bool):
+                    new_dict = {"to_clan": False, "aware_individuals": []}
+                    if death["revealed"]:
+                        new_dict["to_clan"] = True
```
j-gynn:
```suggestion
                    new_dict = {"to_clan": death["revealed"], "aware_individuals": []}
```

scripts/cat/history.py
```diff
@@ -416,37 +430,86 @@ def add_scar(self, scar_text, condition=None, other_cat=None):
...
+        :param murderer_id: murderer's cat ID
         """
         if not game.clan:
             return
-        if "is_murderer" not in murderer.history.murder:
-            murderer.history.murder["is_murderer"] = []
+        if "is_murderer" not in self.murder:
+            self.murder["is_murderer"] = []
         if "is_victim" not in victim.history.murder:
             victim.history.murder["is_victim"] = []
```
j-gynn, line 441: just as an aside, i think we might want an "else" statement here that raises the alarm if the victim already has a history of being a victim? Because, y'know, they shouldn't? If they do, they're- they're dead. Y'know?

scripts/events.py
```diff
@@ -315,7 +314,9 @@ def handle_future_events(self):
             if event.moon_delay <= -12:
                 removals.append(event)
             if event.moon_delay <= 0:
-                handle_short_events.trigger_future_event(event)
+                failed = handle_short_events.trigger_future_event(event)
+                if not failed:
+                    removals.append(event)
```
j-gynn:
```suggestion
                if not handle_short_events.trigger_future_event(event):
                    removals.append(event)
```

scripts/events_module/short/handle_short_events.py
```diff
@@ -355,19 +354,26 @@ def trigger_future_event(self, event):
...
+        self.future_event_failed = True
         self.handle_event(
             event_type=event.event_type,
             main_cat=Cat.fetch_cat(event.involved_cats.get("m_c")),
             random_cat=Cat.fetch_cat(event.involved_cats.get("r_c")),
             freshkill_pile=game.clan.freshkill_pile,
             victim_cat=Cat.fetch_cat(event.involved_cats.get("mur_c")),
             sub_type=event.pool.get("subtype"),
-            ignore_subtyping=True if "subtype" in event.pool else False,
+            ignore_subtyping="subtype" not in event.pool,
```
j-gynn, line 369: this does the opposite of the previous. intended?

scripts/events_module/short/handle_short_events.py
```diff
@@ -355,19 +354,26 @@ def trigger_future_event(self, event):
...
         )
 
         self.allowed_events = []
         self.excluded_events = []
 
+        if self.future_event_failed:
+            self.future_event_failed = False
+            return True
+        else:
+            return self.future_event_failed
```
```suggestion
        return False
```

scripts/events_module/short/handle_short_events.py
```diff
@@ -169,6 +170,8 @@ def handle_event(
         # ---------------------------------------------------------------------------- #
         try:
             self.chosen_event = choice(final_events)
+            if self.future_event_failed:
+                self.future_event_failed = False
```
j-gynn: is there a reason you can't just set `future_event_failed` to `False` always? /genq

scripts/cat/history.py
```diff
@@ -416,37 +430,86 @@ def add_scar(self, scar_text, condition=None, other_cat=None):
...
+        :param clan_reveal: set to True if the whole Clan now knows about the murder
+        :param aware_individuals: if only individual cats are learning about the murder, give a list of their cat objects
+        """
+
+        for murder in self.murder["is_murderer"]:
+            if murder["victim"] == victim.ID:
+                if clan_reveal:
+                    murder["revealed"]["to_clan"] = True
+                else:
+                    murder["revealed"]["aware_individuals"].extend(aware_individuals)
```
j-gynn, line 484: I think extending a list with `None` might have unintended consequences? may want to put a `if aware_individuals is None: aware_individuals = []` somewhere

scripts/cat/history.py
```diff
@@ -416,37 +430,86 @@ def add_scar(self, scar_text, condition=None, other_cat=None):
...
+
+        for murder in victim.history.murder["is_victim"]:
+            if murder["murderer"] == murderer_id:
+                if clan_reveal:
+                    murder["revealed"]["to_clan"] = True
+                else:
+                    murder["revealed"]["aware_individuals"].extend(aware_individuals)
```
j-gynn, line 491: as above

scripts/cat/history.py
```diff
@@ -416,37 +430,86 @@ def add_scar(self, scar_text, condition=None, other_cat=None):
...
+    def get_murder_status_text(murder: dict, Cat) -> str:
+        """
+        Returns the complete murder reveal status text for this cat.
+        :param murder: the murder history to pull status text from
+        :param Cat: cat object
+        """
+        text = ""
+        if murder["revealed"]["to_clan"]:
+            text = i18n.t("cat.history.murder_revealed_to_clan", count=1)
+        else:
```
j-gynn: ok it'll be way easier for you to make this change than for me. get rid of the `else` and move everything back out one indent
```suggestion
        if murder["revealed"]["to_clan"]:
            return i18n.t("cat.history.murder_revealed_to_clan", count=1)
```

resources/lang/en/cat/history.en.json
```diff
@@ -48,6 +48,11 @@
         "regular_death": "%{cats} died.",
         "death_cause": "%{death} when {PRONOUN/m_c/subject} %{text}.",
         "murdered": "%{name} murdered %{victims}.",
+        "murder_revealed_to_clan": {
+            "zero": "The Clan is unaware.",
+            "one": "The Clan knows."
+        },
```
j-gynn: i don't like this, if I'm being honest. I did this as a lazy thing for some of the most text-intensive areas (accessories, for example) and am already having regrets. in this case I think it'd be better if it's explicitly `murder_revealed_to_clan` and `murder_not_revealed_to_clan` or something like that. There's no need for it to bastardize the `i18n` pluralization functions. it's not clear to the reader in the python what exactly the number means etc. sowwy

resources/lang/en/cat/history.en.json
```diff
@@ -48,6 +48,11 @@
...
+        "murder_revealed_to_individual": "%{name} knows. ",
```
j-gynn, line 53: is the space after this intentional? should probably be done code-side, so that the burden of sentence separation isn't on the localization teams.

scripts/screens/ProfileScreen.py
```diff
@@ -1438,7 +1438,7 @@ def get_mentorship_text(self):
     def get_text_for_murder_event(self, event, death):
         """Returns the adjusted murder history text for the victim"""
 
-        if switch_get_value(Switch.show_history_moons):
+        if switch_get_value("show_history_moons"):
```
j-gynn: Why?

scripts/screens/ProfileScreen.py
```diff
@@ -1568,7 +1566,7 @@ def get_death_text(self):
                         text = text
 
                     if moons:
-                        text += f" ({i18n.t('general.moons_date', moon=death['moon'])})"
+                        text += f" ({i18n.t('general.moon_date', moon=death['moon'])})"
```
j-gynn, line 1569: good catch; might also be an issue floating around to do with this?

j-gynn (changes requested):

scripts/events_module/short/handle_short_events.py
scribblecrumb: Previous was incorrect.  The intent here is for us to ignore the subtypes events have *if* no subtypes were given.  That way writers can list possible event_ids that have different subtypes and have the filtering ignore the subtype check.

scripts/screens/ProfileScreen.py
scribblecrumb: That's called "me when i didn't quite understand how to get switches", sorry thought I caught all of those lmao

j-gynn: Ah ok, I thought there might've been a missing `Switch` enum 😅

scripts/cat/history.py
scribblecrumb: Why? The intention is that if a clan is aware, we don't show the aware individuals list.  I feel like the aware individuals is only important when they're the *only* ones who know.  Or am I missing something here with what you're trying to fix lmao

j-gynn: because this is what the code does already 🤣 if this check passes, it runs this i18n block, then the next thing it does after is `return text`. so this just shortcuts that? I think?

less indentation means you have to store less information in your brain, easier to balance cognitively.

scribblecrumb: ahhh okay, yeah this was down to me just not properly reading the suggested code, I didn't realize you wanted a return added there.

scripts/cat/history.py
```diff
@@ -443,6 +439,8 @@ def add_murder(self, murderer_id, victim):
             self.murder["is_murderer"] = []
         if "is_victim" not in victim.history.murder:
             victim.history.murder["is_victim"] = []
+        else:
+            print(f"WARNING: victim cat: {victim.ID} has no victim history.")
```
j-gynn:
```suggestion
            print(f"WARNING: victim cat: {victim.ID} already has a murder history - as the victim!")
```
I do also think there's a possible case for an intentional crash here rather than just a debug print. but eh.

j-gynn (changes requested):
just the one

j-gynn (approved):

PR #3669, opened 2025-07-09
[CLEANUP] Declassify handle short events

### LINKED ISSUE #3732: [BUG] same cat appearing twice in a moon event that involves two cats
### Pre-requisites

### Description

exactly as the title says. a cat duplicates itself instead of finding another cat to watch the sunrise with. 

### Grade

1. timeskip until it happens. 

_No response_

dc91309c

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

scribblecrumb:
## About The Pull Request
I didn't plan for this to get as large as it did, but sadly circular imports required it.

This PR gets rid of `handle_short_events.py` entirely. She's gone. Instead, the vast majority of her functions are now just part of the `ShortEvent` class. The `ShortEvent` class now has a func called `execute_event()` that will handle everything that happens after an event is chosen.

As for how we choose an event: I've moved all of the filtering and generation of short events out of `generate_events.py`. Now they're held within `short/generation.py`.  This module has all of the generation/filtering funcs from before but also has the newly added `create_short_event()`, which is what we will now call when we want to make a short event happen.

This also adds prevention for event repeats. We now store a set of all used event_ids.  If we run through all the frequencies without finding usable events, the set is cleared and we repeat finding frequency 4 (most common) events. This hopefully also serves to make the lower frequency events even rarer (something i think we *do* want atm). 

This has come with some additional refactoring to future events! I've separated the `FutureEvent` class file from it's module functions (in order to avoid more circular import bullshit) but also just for organization.  `FutureEvent` objects also have a `triggered` attribute now that helps us detect when a future event should be removed from the queue in a cleaner way than what I slapped together in #3662 

I've also added and fixed up some short event tests! Unfortunately a lot of short event handling is entangled with `game.clan` checks, so I couldn't add all the tests I wanted to. However, some tests are better than none!

I've changed the timeskip test to only test 500 moons.  This is because the test was taking a, frankly, gross amount of time to complete.  Why was it so much longer than before?  Because turns out the various changes made in this declassifying have made it easier for a clan to survive a bulk skipping.  Previously, the test would end up running the vast majority of it's skips on a dead clan.  My theory is that the event repeats making low frequency events even less frequent has led to less severe injuries, and thus less deaths.  I'm of the opinion this is a good thing, as we do generally want clans to be capable of surviving without patrols.  However, it means that the timeskip test now takes like, an hour to complete, because it's not running 800 of it's moonskips on a dead clan.  As such, I've reduced the number of moonskips.  We're still technically testing more moonskips than we were before simply because the clan is actually surviving.

This makes our code base more readable and understandable, *and* helps avoid circular import issues by breaking up massive files into smaller ones.

## Proof of Testing
I've done the usual moonskip spamming and haven't been able to trigger anymore crashes from my changes. 

## Changelog/Credits
- Big backend code structure changes for short events
- Cats can no longer participate in events with themselves
- Added a repeat prevention, this should dramatically reduce the odds of seeing the same event twice in a single moon.


scripts/events_module/future/future_event.py
```diff
@@ -87,6 +14,11 @@ def __init__(
 
         self.involved_cats = involved_cats
 
+        self.triggered = False
+        self.allowed_events = [self.pool.get("event_id")]
```
j-gynn: this will result in `[None]` if there's no event_id. desired?

scripts/events_module/future/future_event.py
```diff
@@ -87,6 +14,11 @@ def __init__(
...
+        self.excluded_events = [self.pool.get("excluded_event_id")]
```
j-gynn: same here

scripts/events.py
```diff
@@ -320,8 +322,18 @@ def handle_future_events(self):
...
             if event.moon_delay <= 0:
-                if not handle_short_events.trigger_future_event(event):
+                create_short_event(
+                    event_type=event.event_type,
+                    main_cat=Cat.fetch_cat(event.involved_cats.get("m_c")),
+                    random_cat=Cat.fetch_cat(event.involved_cats.get("r_c")),
+                    victim_cat=Cat.fetch_cat(event.involved_cats.get("mur_c")),
+                    sub_type=event.pool.get("subtype"),
+                    future_event=event,
+                )
```
j-gynn, line 336: this feels very counterintuitive ngl

tests/test_handle_short_events.py
```diff
@@ -1,97 +1,141 @@
...
+from scripts.cat.pelts import Pelt
 
 
 class TestHandleEvent(unittest.TestCase):
-    pass
+    def setUp(self):
+        self.chosen_event = ShortEvent(event_id="test")
+        self.chosen_event.main_cat = Cat()
+        self.chosen_event.random_cat = Cat()
+        self.pelts = Pelt
```
j-gynn: why is this here?

tests/test_handle_short_events.py
```diff
@@ -1,97 +1,141 @@
...
+        self.chosen_event.execute_event()
+        self.assert_intersection(
+            self.chosen_event.main_cat.pelt.accessory, self.pelts.collars
+        )
 
     def test_notail_cats_do_not_get_tail_accessories(self):
-        self.test.chosen_event.new_accessory = self.pelts.tail_accessories
-        self.test.main_cat.pelt.scars = "NOTAIL"
+        self.chosen_event.new_accessory = self.pelts.tail_accessories
+        self.chosen_event.main_cat.pelt.scars = "NOTAIL"
```
j-gynn: isn't this a list? /genq

scripts/game_structure/constants.py
```diff
@@ -1,5 +1,8 @@
 import ujson
 
+from scripts.clan_resources.herb.herb import HERBS
+from scripts.clan_resources.supply import Supply
```
j-gynn, line 8: edit: are these both Enums in their own files? so long as they don't import any of our business logic into themselves and are fully self-contained, this is ok I guess.

original:
don't import into constants - the whole idea is that this is an area that doesn't get touched by our code so it can go everywhere

j-gynn (changes requested):

scripts/game_structure/constants.py
scribblecrumb: Yeah they don't import any of our scripts.  Though if you'd rather I remove them from here then I can! Don't actually remember what benefit i thought this had

scripts/events.py
scribblecrumb: In what way?

tests/test_handle_short_events.py
scribblecrumb: Ah, holdover from when I copied this from the test accessories

scripts/game_structure/constants.py
scribblecrumb: Ah now I remember, it's cus I moved the SUPPLY_TYPES and SUPPLY_TRIGGERS here, as those feel like relevant constants to have available. Both of those require the enums.

tests/test_handle_short_events.py
scribblecrumb: indeed they are

j-gynn (approved):

Issue #3650, opened 2025-07-06
[ENHANCE] Add a tag for thoughts directed at another cat to exclude newborns


SunlitFable:
### Description

Thoughts can be directed at certain groups of other cats by specifying their age group, but there's no tag to encompass all but newborns from these thoughts. Currently, to exclude just newborns from a thought, writers need to list every other age group, which bloats the tags and decreases readability. Ideally, we should be able to list one tag and have it automatically encompass all age groups except newborns.

scribblecrumb:
Might even be preferable to only allow newborns when *specified*.  Essentially assuming newborns aren't included, rather than requiring writers to add a "no_newborn" tag.

PR #3690, opened 2025-07-11
Seperating kits from newborns

Yzzuwyzzuf:

this is good for the game as it removes newborns from being involved with relationship events as well as limiting the relationship events kits can be involved with preventing warriors from having deep intellectual conversations with kits or asking them for help or judging them for things they realistically are not able to do. this is part 1, admiration and comfort
## Linked Issues

pr is linked to issue #3650 

## Changelog/Credits

adjusted tagging for kits and newborns


scribblecrumb:
Hey Fuzzy!  #3684 made it so that newborns aren't included unless specified. So we don't need to add all ranks except newborn in order to exclude them.

Yzzuwyzzuf:
OK gotcha, I thought it just added the newborn tag  only. I'll remove any unnecessarily added tags then and only keep it for ones that exclude kits too since that has been something that pops up in screenshots chat in the discord from time to time

Yzzuwyzzuf:
> Hey Fuzzy! #3684 made it so that newborns aren't included unless specified. So we don't need to add all ranks except newborn in order to exclude them.

Should be fixed👍

j-gynn (approved):

PR #3706, opened 2025-07-11
Switch game config from json to toml

larkgz:
## About The Pull Request

* Switches `game_config.json` to `game_config.toml`, using tomli for parsing. We can’t use Python’s default tomllib parsing implementation because it was added in 3.11 and our minimum Python version is 3.8. tomli and tomllib are both read-only, but I think that this is fine because this file is intended to be edited by human hands
* Switches `null` values to empty strings, since toml doesn’t support null values. Specifically modified:
    * `debug_ensure_patrol_id`
    * `debug_ensure_patrol_outcome`
    * `debug_ensure_event_id`
    * `debug_type_override`
  Had to modify some code that uses these values if it relied on the fact that the values were `null`. Please double-check I didn't break anything
* Did not add any automated conversion because config files are not a part of save files and because that would make adding comments very hard
* Adding a lot of comments

* Allows comments directly in config file
* Might be easier to understand for users, since JSON matching brackets and the fact that trailing commas are illegal seem to be commonly confusing
* Has a complete and maintained parser in Python, unlike JSON5 and JSONC (as far as I could find when googling anyway)


resources/game_config.json
j-gynn, line 1: this file should be deleted, right?

j-gynn:
Please make sure you run Black so that the formatter test passes (I believe the command is `uv run black .`). I think one file should be deleted but otherwise am happy to approve, assuming no disagreements about the change are raised on Discord.

resources/game_config.json
larkgz: I was thinking of replacing it with a temporary announcement that the file changed.

larkgz:
For sure! I just want to finish documenting it first. Once I do that I’ll run black and also mark the PR as ready to review.

larkgz:
@j-gynn Should be ready now!

j-gynn (approved):

PR #3720, opened 2025-07-13
keep the kits in camp!

Yzzuwyzzuf:
## About The Pull Request
moved thoughts involving things not possible to see in camp from the main general category to the category excluding kits as well as removing unnecessary living status constraints from several categories

this is good for clangen by removing bits of unnecessary code and keeping kits from having thoughts of seeing stuff like twolegs and kittypets 

## Changelog/Credits

kits can no longer go out and explore near twolegs

### DIFF
--- a/resources/lang/en/thoughts/alive/general.json
+++ b/resources/lang/en/thoughts/alive/general.json
@@ -84,16 +84,6 @@
             "Is waiting to watch the sunrise",
             "Is waiting to watch the sunset",
             "Is grooming, grooming, grooming away...",
-            "Saw a pair of Twolegs near camp today!",
-            "Saw a Twoleg petting a loner",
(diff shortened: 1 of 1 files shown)


j-gynn (commented):
Why are you removing the `living` constraint from these thoughts? /genq

Yzzuwyzzuf:
> Why are you removing the `living` constraint from these thoughts? /genq

Because i dont understand why theyre necessary here, the game seemingly is able to understand that unless it specifies a dead cat the other cat in these situations refers to a living one so why have it specify living for these specific segments? If it is a problem I can put them back, it just didn't appear to be necessary when other segments don't require r_c to be specified as alive

j-gynn (approved):

PR #3659, opened 2025-07-08
[FEATURE] Relationships Rework

scribblecrumb:
## About The Pull Request
This PR reworks cat relationships!

Relationships are now defined by 5 values: Like, Respect, Trust, Comfort, and Romance.  Each value has corresponding levels: Extreme Neg, Mid Neg, Low Neg, Neutral, Low Pos, Mid Pos, and Extreme Pos.  Romance is the only outlier here, it just has the Neutral and Pos levels.  Each value has different names used for each of it's levels. This name is how we will refer to them when constraining events.

Values now have the strenum class `RelValue` and their levels have the strenum class `ValueLevel`. `ValueLevel` has plenty of helper funcs to assist in identifying them within relevant categories.

Under the hood, the level of a value is determined by a number.  Negative numbers up to -100 are the Neg levels. Positive numbers up to 100 are the Pos levels.  The ranges covered by each level is determined by a game config `value_intervals`.  Right now, Low are 1-24, Mid are 25-69, and Extreme are 70-100.  0 is Neutral.  This makes the Mid levels the largest range, which seems fitting to me.

You can call both the int and level name of each value using the corresponding `relationship` properties.

All of this has come with quite a bit of refactoring throughout our relationship code.  A lot of filtering has been cleaned up and is now using our communal `event_filters` func, which is wonderful for consistency and general avoidance of repeated code.  This did unfortunately mean that I removed the interaction tests for biome and season filters, as they no longer work without the use of `game.clan`, which our tests cannot access.

All events have been adjusted to the new tags, documentation is updated, and event editor provides tools for adding them.  Old saves convert easily into the new system.

I've also, in general, made nearly all references to rel values consistent in the language used.

## Why This Is Good For ClanGen
This gives all our relationship values both negative and positive aspects that will give us so much room for exploring relationships more fully!

also coincidentally fixes #3160

## Proof of Testing
image
Documentation update

I've also had a beta tester thread going: https://discord.com/channels/1003759225522110524/1391155256322424885 and it looks like everything is working as intended.  I've also stepped through debug view numerous times to ensure things are in order.

## Changelog/Credits
- Relationship UI has gotten an upgrade!
- Relationship values are now Like, Respect, Comfort, Trust, and Romance.  All values except Romance have both negative and positive sides.
- There's multiple levels within each relationship value that cats can achieve, changing their relationships with each other.

PR #3744, opened 2025-07-14
[DEV] Adding not_xyz rel tags

scribblecrumb:
## About The Pull Request
This adds more relationship tags in the form of `not_variable` to mirror our current tags.  This also globalizes the "littermates" tag that thoughts had been using to be usable by all event types that utilize the filter_relationship_type() func.

I've update the short event documentation only at the time, as I'm cognizant of #3659 moving the rel tag lists into one shared tag list page.

The event editor has also been updated with the new tags, their descriptions, and conflicting tags.  The schemas also feature the new tag lists.

`not_parents` has been added as a constraint on one of the newest thought pools added by tobes.

More constraints means writers can get more creative!

## Proof of Testing
image

Mostly tested these via debug window, here you can see the `not_parent` returning False, this is occurring correctly as the cat in question *is* the parent of the other cat.

## Changelog/Credits
- New relationship tags for use by the writers in the majority of event formats: `littermates`, `not_littermates, `not_siblings`, `not_parent`, `not_child`, `not_app`, `not_mentor`.  `littermates` was previously a thoughts-only constraint, it's now available to other event formats like ShortEvents and Patrols.


scripts/utility.py
```diff
@@ -1256,6 +1297,13 @@ def filter_relationship_type(
         if not group[1].ID in group[0].apprentice:
             return False
 
+    if "not_mentor" in filter_types:
+        test_cat = group[0]
+        testing_cats = [cat for cat in group if cat.ID != test_cat.ID]
+
+        if any([inter_cat in test_cat.apprentice for inter_cat in testing_cats]):
+            return False
+
```
j-gynn, line 1299: does not appear to be an equivalency

scripts/utility.py
```diff
@@ -1268,6 +1316,13 @@ def filter_relationship_type(
         if not group[0].ID in group[1].apprentice:
             return False
 
+    if "not_app" in filter_types:
+        test_cat = group[0]
+        testing_cats = [cat for cat in group if cat.ID != test_cat.ID]
+
+        if any([inter_cat in test_cat.mentor for inter_cat in testing_cats]):
+            return False
+
```
j-gynn, line 1314: does not appear to be equivalent to the `app/mentor` constraint above? Not checking for patrol leader or something?

j-gynn (changes requested):

scribblecrumb:
Ended up just moving the patrol lead check to the very start, seems a bit silly that we were doing it in individual if statements.  It seems we want the assumption to be that the patrol lead is *always* the first cat, so might as well just be doing that first thing.

j-gynn (approved):

PR #3745, opened 2025-07-14
Congratulations! It's a Murder!

j-gynn:
## About The Pull Request

Adds a new `sub_type`, `hidden_murder_reveal`. This can be used when you've created a sneaky, sneaky murder that masquerades like a normal death... but we wouldn't have any of those in ClanGen, would we...?! Nah.

Except now we do. Only like two or three for now, but I'm sure other industrious writers with evil in their hearts will dig up more "boring" death events that are ripe for a surprise redrum. 

Also fixes the inconsistency of `sub_type` vs `subtype` that was stopping some events from triggering. 

Also fixes an entirely unrelated strange bug with inheritance encountered during testing where a non-existent/None cat could try and create grandparents.

## Proof of Testing

Oh no, a tragic accident. Poor Furledpaw.
image

Hey, wait a minute...!
image

### DIFF
--- a/docs/dev/writing/shortevents.md
+++ b/docs/dev/writing/shortevents.md
@@ -170,13 +170,14 @@ lowercase season names + "any"
 |------------------------|------------------------------------------|
 | war                    | marks event as only occurring during war |
 
-| misc event subtypes | use                                                               |
-|---------------------|-------------------------------------------------------------------|
-| war                 | marks event as only occurring during war                          |
-| murder_reveal       | marks event as being the reveal of a murder                       |
-| accessory           | marks event as giving an accessory to m_c                         |
-| ceremony            | marks event as being the gifting of an accessory after a ceremony |
-| transition          | marks event as being a cat transitioning their gender             |
+| misc event subtypes  | use                                                                               |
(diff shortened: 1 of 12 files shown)


docs/dev/writing/shortevents.md
```diff
@@ -170,13 +170,14 @@ lowercase season names + "any"
...
-| war                 | marks event as only occurring during war                          |
-| murder_reveal       | marks event as being the reveal of a murder                       |
-| accessory           | marks event as giving an accessory to m_c                         |
-| ceremony            | marks event as being the gifting of an accessory after a ceremony |
-| transition          | marks event as being a cat transitioning their gender             |
+| misc event subtypes  | use                                                                               |
+|----------------------|-----------------------------------------------------------------------------------|
+| war                  | marks event as only occurring during war                                          |
+| murder_reveal        | marks event as being the reveal of a murder                                       |
+| hidden_murder_reveal | marks event as being the reveal of a murder that was disguised as a regular death |
```
scribblecrumb, line 177: You'll also need to add this subtype to `resources/dicts/events/types.json`

scribblecrumb (approved):

PR #3751, opened 2025-07-15
[BUGFIX] Change how on death thoughts are handled

### LINKED ISSUE #3553: [BUG] Not all cats dying in mass death event have on-dead thoughts
### Pre-requisites

### Description

After a mass death event, only half of the cats (in this case, 5 out of 9) have the proper on-dead thoughts, the other 4 cats have regular StarClan thoughts.

### Grade

1. Have mass death events enabled.
2. Get a Clan big enough to have a lot of cats dying during a mass death event (at least 6 cats should die).
3. Skip moons until a mass death event is triggered.
4. Without closing the game, go to StarClan and check the thoughts of all the cats who died during the event.

_No response_

94639b8f6

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

Image

### LINKED ISSUE #3781: [BUG] cat "clings to their litter" despite having no kits
### Pre-requisites

### Description

cat died on moonskip due to yellowcough. her thought on death was "clings to the warmth of her litter" even though she has never had any kits at all, biological or adoptive

### Grade

1. Moonskip
2. Have a cat die and go to StarClan
3. Go to StarClan tab
4. Click on the newest dead cat
5. If that thought is chosen, see error

_No response_

f629af6

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image
Image
Image

what litter, jay... you have no kits...

scribblecrumb:
## About The Pull Request
Discovered in #3750 that on_death thoughts weren't filtered at all.  Fair enough, they were originally kept very simple and had no other cats included.  However, we just had new on_death thoughts added that assumed filtering and so now do not show up in the correct situations.

This PR kinda reworks how we handle on_death thoughts.  The new status class actually cleans things up a bit.
- `leader_death_thought()` was removed, as we now just use the `new_death_thought()` func.
- `new_death_thought()` now uses the same filtering code as our normal thoughts. This means we can also kill `create_death_thoughts()`. 
- within `die()` we no longer call `leader_death_thought()` or `new_death_thought()` instead we use the `thoughts()` func that is already used for our normal thoughts.
- within `thoughts()` we now have two args that let us pull the on_death thoughts instead of normals ones: `just_died` and `lives_left`
- I've also killed the awkward `darkforest` and `isoutside` attributes that were confusing to follow and rather clunky. We now just use the `status.group`.
- exiled cats now get shunted to the unknown residence within the `send_to_afterlife()` function instead of having to be manually moved there after death.

## Proof of Testing
image
image

Oatshade the exiled cat went to the UR and has an appropriate death thought

image

Iceroach the lost cat goes to the SC and has a (semi) appropriate death thought. I don't think we actually have any filters for catching lost cats? Regardless this thought has no filters on it anyways, so technically this is working as intended.

image

Leader losing one life
image

Leader losing all lives

## Changelog/Credits
- Thoughts gained upon death are now filtered the same as normal thoughts


scripts/cat/cats.py
```diff
@@ -654,15 +637,9 @@ def die(self, body: bool = True):
         # mark the sprite as outdated
         self.pelt.rebuild_sprite = True
 
-        # exiled cats are special, cus they get kicked out a heaven
-        if isoutside and self.status.is_exiled():
-            self.status.add_to_group(CatGroup.UNKNOWN_RESIDENCE)
```
j-gynn, line 661: is this now handled in the send_to_afterlife function in status?

scripts/cat/status.py
```diff
@@ -488,7 +488,9 @@ def send_to_afterlife(self, target: CatGroup = None):
             return
 
         # if we have an outsider who has never been a clancat, they go to the unknown residence
-        if self.is_outsider and not self.is_former_clancat:
+        if (self.is_outsider and not self.is_former_clancat) or (
+            self.is_outsider and self.is_exiled(CatGroup.PLAYER_CLAN)
```
j-gynn:
```suggestion
        if self.is_outsider and (self.is_exiled(CatGroup.PLAYER_CLAN) or not self.is_former_clancat
```

scripts/cat/status.py
```diff
@@ -529,15 +531,14 @@ def change_group_nearness(self, group: CatGroup):
...
         """
-        return [
-            entry["standing"]
-            for entry in self.standing_history
-            if entry["group"] == group
-        ]
+        for entry in self.standing_history:
+            if entry["group"] == group:
+                return entry["standing"]
+        return None
```
j-gynn: should probably return `[]` since that's easier to handle downstream whilst still being able to say `if value is None`

j-gynn (changes requested):

scripts/cat/cats.py
scribblecrumb: Yup!

j-gynn (approved):

Issue #3788, opened 2025-07-18
[BUG] Game crashes while selecting cats in mediator screen


KiraDragoness:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

While trying to mediate 2 cats, the game crashes when you try selecting cats or switching the already selected cats.

### Grade

Dire - Still playable, but a major feature is broken

### Steps to reproduce the behavior

1. Have a mediator in your clan
2. Go to the mediating screen
3. Try selecting 2 cats, or switching to another cat after selecting them.

### Commit # or Game Version Number

0b312f81d

### Error log

<img width="604" height="282" alt="Image" src="https://github.com/user-attachments/assets/a981137d-8928-47c4-bc40-eb5839076a53" />

scribblecrumb:
probably fixed in #3659, last I tested on it the mediation screen was working fine

Vickarooni:
i think it crashes when you click too fast, i'm testing it right now and if you click "slowly" it doesn't crash

Yzzuwyzzuf:
my testing shows it only occurs when the cats are related. i get it when i accidentally click on littermates which tracks with the error message talking about relation

j-gynn:
Yes, I know what this bug is - it's only to do with littermates. It's due to #3722, I'll make a fix real quick.

PR #3798, opened 2025-07-20
fixes mediation screen crash #3788

j-gynn:
## About The Pull Request

Fixes the mediation screen crash. Also implements a fix for the heinously overlapping text for siblings (see below) by putting the relationship details into a tooltip. It has a strangely long tool_tip_delay which I couldn't figure out how to reduce?

I tried saying "hover" or something on the "related" label but there really isn't the space. I would very much like this screen to be reworked 😇

Pre-tooltip
image

Post-tooltip
image

### DIFF
--- a/scripts/screens/MediationScreen.py
+++ b/scripts/screens/MediationScreen.py
@@ -556,6 +556,7 @@ class MediationScreen(Screens):
             object_id="#text_box_22_horizleft_spacing_95",
             manager=MANAGER,
         )
+        self.selected_cat_elements["col1" + tag].disable()
 
         mates = False
         if len(cat.mate) > 0:
@@ -573,6 +574,7 @@ class MediationScreen(Screens):
             object_id="#text_box_22_horizleft_spacing_95",
             manager=MANAGER,
         )
+        self.selected_cat_elements["col2" + tag].disable()
 
         # Relation info:
         if related and other_cat and not mates:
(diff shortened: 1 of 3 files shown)


scribblecrumb:
dw i've got plans for reworking this screen

scripts/screens/MediationScreen.py
```diff
@@ -634,14 +640,20 @@ def draw_info_block(self, cat, starting_pos: tuple):
...
+                i18n.t("general.related_text"),
+                ui_scale(pygame.Rect((x + 110, -15), (80, -1))),
+                starting_height=3,
+                object_id="#text_box_22_horizleft_spacing_95",
+                manager=MANAGER,
+                anchors={"top_target": self.selected_cat_elements["col2" + tag]},
+            )
+            self.selected_cat_elements["col2_relation" + tag].set_tooltip(
+                text=i18n.t(relation)
+            )
```
scribblecrumb, line 661:
```python
self.selected_cat_elements["col2_relation" + tag].tool_tip_delay = 0
```
This will fix the tooltip delay. I'm not sure why changing it via themes doesn't work, we should probably make an issue on pygui's github abt it.

I would also suggest disabling the textbox (and other textboxes within the cat view) as that will prevent it from triggering the change to an I cursor.  Not super important, but it helps the UI look cleaner and makes it ever so slightly more apparent that you can hover for more info.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #3804, opened 2025-07-20
[BUGFIX] Dead patrol cats being injured

### LINKED ISSUE #3762: [BUG]cats retain injuries if they got them upon death
### Pre-requisites

### Description

One of my cats was killed by a Coyote and also got scrapes right after which i found odd. Checked their page and they have scrapes. Timeskip does not remove said injury. Skipped 5 moons and it is still retained

### Grade

Get the listed patrol (i had a patrol of i believe 2 cats, maybe 3) and fail it

Patrol id: pln_bord_greenleaf_coyote_small_patrol1 
outcome: SUCCESS: False

4c9d6faf

### Error log

No error given

### Additional context

_No response_

### Screenshots

Image
Image

scribblecrumb:
## About The Pull Request
Added a list comprehension to remove dead cats from the list of cats to injure.

## Linked Issues
Fixes #3762

## Proof of Testing
image
Stepped through debug and the dead cat was successfully removed from list of cats to injure, you can see only one cat object is still in `cats`, which is the remaining alive patrol cat.

## Changelog/Credits
- Cats who die on patrol can no longer be injured after death.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -1844,6 +1844,8 @@ class Cat:
         :param lethal: Allow lethality, default `True` (bool)
         :param severity: Override severity, default `'default'` (str, accepted values `'minor'`, `'major'`, `'severe'`)
         """
+        if self.dead:
+            return
         if name not in ILLNESSES:
             print(f"WARNING: {name} is not in the illnesses collection.")
             return
@@ -1912,6 +1914,9 @@ class Cat:
         :param severity: _description_, defaults to 'default'
         :type severity: str, optional
         """
+        if self.dead:
+            return
+
         if name not in INJURIES:
             print(f"WARNING: {name} is not in the injuries collection.")
             return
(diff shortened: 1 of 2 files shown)


j-gynn:
I wonder if it might be better to put this check in `Cat` in `get_injured`, and possibly in `get_ill` and `get_permanent_condition` too? That way, it covers any time we might give Things to a cat that may have just died. 

I'm supposing that this is a bug introduced by the Status update, so it's not unreasonable that other situations might occur.

scribblecrumb:
I'm actually pretty certain that it wasn't introduced in the status update.  It seems like just a rarely occurring bug due to rng (and probably also a bit of players not noticing it, how often do they check a dead cats profile and conditions?).  The patrol referenced in the issue is one that pulls a random selection of patrol cats to injure, which means that it wouldn't always pull one of the dead cats.  I had to run the patrol numerous times before I actually encountered the problem, and even then it was due to debug view that I noticed.  We don't have many patrols using `multi_cat` for injuries, so it's probably been a longstanding, unnoticed problem.

But you're right that it's a good idea to cover our bases by adding checks to those funcs instead

j-gynn (approved):

PR #3815, opened 2025-07-20
[BUGFIX] Rel constraints

### LINKED ISSUE #3737: [BUG] Mate-only misc event triggers for cats who are not mates
### Pre-requisites

### Description

An event that is supposed to be triggered only for 2 cats that are mates shows up for 2 cats that are not mates nor have any romantic relationship towards each other. One of the 2 cats has a mate, that wasn't mentioned in the event.

### Grade

1. Skip moons
2. Check the misc events and see if this particular event (see screenshot) appears for cats who are not mates.

_No response_

dc91309

### Error log

_No response_

### Additional context

It probably is just a coincidence, as it should not be correlated, but on the same moon, the exact same event triggered for another 2 cats that are actually mates.

### Screenshots

Image

Image

Image

scribblecrumb:
## About The Pull Request
We just needed a `.copy()` bane of my existence

## Linked Issues
Fixes #3737

## Proof of Testing
image
They're mates, they're cuddling.  I forced this event upon the whole clan, only these two (the only mates) got it.

## Changelog/Credits
- Relationship constraints work again

### DIFF
--- a/scripts/events_module/event_filters.py
+++ b/scripts/events_module/event_filters.py
@@ -11,7 +11,7 @@ from scripts.utility import (
 )
 
 
-def event_for_location(locations: list) -> Optional[bool]:
+def event_for_location(locations: list) -> bool:
     """
     checks if the clan is within the given locations
     """
@@ -34,7 +34,7 @@ def event_for_location(locations: list) -> Optional[bool]:
         elif req_biome == game.clan.biome.lower():
             if "any" in req_camps or game.clan.camp_bg in req_camps:
                 return True
-        return False
+    return False
 
 
 def event_for_season(seasons: list) -> bool:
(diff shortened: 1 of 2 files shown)


scripts/events_module/event_filters.py
```diff
@@ -12,7 +11,7 @@
 )
 
 
-def event_for_location(locations: list) -> bool:
+def event_for_location(locations: list) -> Optional[bool]:
```
j-gynn: why,,, is this optional...?

scribblecrumb: pycharm was yellin at me that it returns bool and none, not sure where it's getting the none from but i wanted to make it shut up :shrug:

j-gynn: `line 37` needs to be unindented one level. otherwise it's only checking one location in the entire set.

scribblecrumb: aha! cool, thanks for finding that

j-gynn (approved):

PR #3816, opened 2025-07-20
[BUGFIX] Updated handling of other clan cats

### LINKED ISSUE #3795: [BUG] Cat From Another Clan Invisibly Joins My Clan
### Pre-requisites

### Description

A litter of kits were dropped off and adopted by a queen from another clan. The cat shows up to be seen in the event, when viewed she is listed as a mediator and a clan cat. She is not added to my clan, nor the cats outside the clan, but is available to be selected as a mate.

### Grade

Skip moon
Cat drops off litter of kits

_No response_

0b312f8

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image
Image
Image
Image

scribblecrumb:
## About The Pull Request
I only meant to fix #3795, but I fell down a rabbit hole.

So the issue causing the invisible bug was that we just. weren't telling the lists where other clan cats were meant to go.  We no longer refer to them as "former" clancats in the new status system, rather they're actually assigned a clan!  Since they weren't outsiders and weren't in the player clan, they had no list to belong to, but being clancats let them into the next/previous list with the player clan cats!

So I've implemented the following changes in an attempt the bridge the gap without getting super duper in depth with other clans:
- encountered other_clan cats now display their home clan in their profile, despite being in the COTC list. This ended up being more involved than I expected, since I decided to also display which clan had exiled the exiled cats (even tho technically only player clan can exile anyone atm).  This exile message persists in death!  Home clans also persist in death, so you can see who in an afterlife belonged to what Clan.  Apologies for how tangly the code got here, there was a lot of things to juggle.
- next/prev list now takes into account if a cat is in another clan or in the player clan, as well as accounting for death afterlife, since that was broken too. Afterlife required a second list comprehension, cus for the life of me I couldn't figure out how to handle that in the main comprehension while also letting other clan cats appear next to outsiders.
- Turns out we weren't actually giving other clan cats *clan* names! So that's been fixed here as well.
- I've added a default backstory display for other clan cats! Just a basic "cat lives in other clan"
- This also lead me to changing how exiled and lost backstory displays work, as their backstories were being overwritten by the default outsider backstory (i.e "this cat is a loner").  Now they display their *actual* backstories with a little appending blurb about being lost or exiled.  At some point this should really change, as there's slight wonkiness with these cats having access to thoughts requiring their "hidden" backstory ("this cat came to the clan after leaving their old clan".  I don't think the wonkiness is severe enough to require change *right* now, though.  It would require a lot more work to change it than I think this PR should have.
- other_clan cats are handled on moonskip the same way as outsiders.  This isn't *exactly* ideal, but I think it's a fair enough stopgap until we actually start digging into other_clan simulating stuff.  This at least allows them to age, die, ect.
- other_clan cats are able to access normal clan thoughts, but the only random_cats allowed in their thoughts are other generated cats from their clan!

## Linked Issues
Fixes #3795

## Proof of Testing
image

Exiled message

image

Lost still appears as normal

image

Normal outsiders are also working

image

In death, we still show the exiled status!

image
image

And all cats show their living home clan affiliation.

image

Coldfoot is our new LyreClan cat!

image

Turns out she's a mediator

image

And she appears in the COTC list!

image

Twilightminnow also gave her kits over this moon (i was forcing the event lol), turns out she's also a mediator, but for ShadeClan!

image
image

Exiled/lost backstory display

image

Other clan backstory display

image
image

Other clan cats can have clan thoughts, but only in reference to other cats in *their* clan! (second image has an example of a thought that seems out-of-place, but is being allowed due to this cat's hidden backstory. once again, not so major that it's vital to fix, in fact i think this thought in particular is an easy one for players to dream up a plausible explanation for lol)

image

other clancats can die!

## Changelog/Credits
- Exiled, lost, and other Clan cats now display updated backstories on their profiles.
- Other Clan cats now appear in the COTC list when generated
- Next/Previous cat buttons now work correctly when navigating our various cat lists
- Other Clan cats now display their home Clan on their profile
- When cats have been exiled, their profile now includes which Clan they were exiled from
- When dead, cats still display if they were exiled
- When dead, cats now display what Clan they were part of in life!
- Other Clan cats will now have typical Clan thoughts, but the only cats they can include in their thoughts are generated cats from the same Clan.
- Other Clan cats will now age and die as intended.


scripts/cat/cats.py
```diff
@@ -1512,7 +1512,14 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
         all_cats = self.all_cats.copy()
         all_cats.pop(self.ID)
-        other_cat = choice(list(all_cats.keys()))
+        # this makes sure other_clan cats don't think about player_clan cats
+        if not self.dead and self.status.is_other_clancat:
+            all_cats = {
+                ID: c
+                for ID, c in all_cats.items()
+                if c.status.group == self.status.group
+            }
```
j-gynn: i don't like that we have to recompute this every time a thought happens - could we possibly try and pass it in through the `one_moon` function? Whilst leaving it optional as `None` so that it can be computed when needed for new cats etc.?

scripts/cat/cats.py
```diff
@@ -1538,43 +1545,46 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
-                    other_cat = None
-                    break
+        other_cat = None
+        if all_cats:
+            other_cat = choice(list(all_cats.keys()))
+            # for cats inside the clan
+            if where_kitty == "inside":
+                dead_chance = getrandbits(4)
+                while (
+                    other_cat == self.ID
```
j-gynn: why don't we just do `all_cats.pop(self.ID, None)` on line 1548? The `None` argument acts like a default in the same way as we do for `.get()` to avoid KeyError

scripts/cat/cats.py
```diff
@@ -1538,43 +1545,46 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
+                    and len(all_cats) > 1
+                    or (all_cats.get(other_cat).dead and dead_chance != 1)
+                    or (other_cat not in self.relationships)
+                ):
+                    other_cat = choice(list(all_cats.keys()))
```
j-gynn: can we `pop` the chosen cat from this list? bcs it wasn't suitable

scripts/cat/cats.py
```diff
@@ -1538,43 +1545,46 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
+                    i += 1
+                    if i > 100:
+                        other_cat = None
+                        break
+            # for dead cats
+            elif where_kitty in ("starclan", "hell", "UR"):
+                while other_cat == self.ID and len(all_cats) > 1:
```
j-gynn: This `while` only ever triggers if we choose the cat themselves? If we `pop` them at the start, we don't even need this loop?

scripts/events_module/outsider_events.py
```diff
@@ -26,29 +28,29 @@ def killing_outsiders(cat: "Cat"):
...
+                    f"but as they drift away, they hope to see "
+                    f"familiar starry fur on the other side."
+                )
+                death_history = (
+                    "m_c died while being lost and trying to get back to the Clan."
                 )
+            else:
+                social = i18n.t(f"general.{cat.status.social}", count=1)
+                text = (
+                    f"Rumors reach your Clan that the {social}, "
```
j-gynn, line 47:
```suggestion
                    f"Rumors reach your Clan that the {social} "
```

j-gynn (changes requested):

scripts/events_module/outsider_events.py
scribblecrumb: Comma is actually correct here!  It's a non-essential clause.  Basically, if a name doesn't change the meaning of a sentence, but is instead supplementary, you add a comma to mark it as such.

scripts/cat/cats.py
scribblecrumb: gonna be real, this whole `where_kitty` if elif stuff wasn't written by me lol, I'll see if I can clean it up tho

scripts/cat/cats.py
scribblecrumb: The thing is, sense its result is individual to each cat, I think we kinda have to do it here.  I suppose we could make up a smaller list of just other clan cats and pass it in to be further filtered down to the correct other clan?

scripts/cat/cats.py
```diff
@@ -1528,64 +1530,52 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
+            if self.status.is_clancat:
+                # we want to limit how often dead cats are thought about
+                thinking_of_dead_cat = getrandbits(4) != 1
                 while (
-                    other_cat == self.ID
-                    and len(all_cats) > 1
-                    or (all_cats.get(other_cat).dead and dead_chance != 1)
-                    or (other_cat not in self.relationships)
+                    len(all_cats) > 1
+                    or (other_cat.dead and thinking_of_dead_cat)
```
j-gynn: shouldn't this be `not` thinking of dead cat?

scripts/cat/cats.py
```diff
@@ -1528,64 +1530,52 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
         other_cat = None
         if all_cats:
-            other_cat = choice(list(all_cats.keys()))
+            other_cat = choice(all_cats)
             # for cats inside the clan
-            if where_kitty == "inside":
-                dead_chance = getrandbits(4)
+            if self.status.is_clancat:
+                # we want to limit how often dead cats are thought about
+                thinking_of_dead_cat = getrandbits(4) != 1
```
j-gynn: this gives a 75% chance to be thinking of a dead cat?

scripts/cat/cats.py
j-gynn: also think this can't be `or`, it has to be `and` with a nested `or`. otherwise, you can have this repeating over and over again with the same invalid cat because `len(all_cats) > 1` is False but the only remaining cat makes one of the others `True`

scripts/cat/cats.py
```diff
@@ -1528,64 +1530,52 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
-            elif where_kitty in ("starclan", "hell", "UR"):
-                while other_cat == self.ID and len(all_cats) > 1:
-                    other_cat = choice(list(all_cats.keys()))
+
+                    other_cat = choice(all_cats)
+
                     i += 1
                     if i > 100:
                         other_cat = None
                         break
```
j-gynn: could move this lot up to before `1512` and combine so that you don't repeat it?

scripts/cat/cats.py
```diff
@@ -1528,64 +1530,52 @@ def thoughts(self, just_died=False, lives_left: int = 0):
...
             # it appears as for now, kittypets and loners can only think about outsider cats
-            elif where_kitty == "outside":
-                while (
-                    other_cat == self.ID
-                    and len(all_cats) > 1
-                    or (other_cat not in self.relationships)
-                ):
-                    other_cat = choice(list(all_cats.keys()))
+            elif self.status.is_outsider:
+                while len(all_cats) > 1 or (other_cat not in self.relationships):
```
j-gynn:
```suggestion
                while len(all_cats) > 1 and (other_cat not in self.relationships):
```

j-gynn (changes requested):

scripts/cat/cats.py
scribblecrumb: Oh you know what, I should have changed this to a `== 1`. ugh, the old code was not readable when it came to this

scripts/events_module/outsider_events.py
j-gynn: I thought the use of the word "the" makes "the rogue" like a title or an adjective or something?
"Rumors reached your Clan that the rogue has died recently" doesn't work to me.  It should be "the rogue Linda died", or "a rogue, Linda, died". Anyway, irrelevant to the content of this actual PR i suppose

j-gynn: ah yeah that's what's screwing me over. it should be "a" not "the"

scribblecrumb: ahh, well i guess you could fix that in #3838 ;)

j-gynn (approved):

PR #3838, opened 2025-07-23
pronounify outsider_events.py

j-gynn:
## About The Pull Request

Localizes and pronounifies the outsider events that show when cats die. Also technically expands this system, as now you can have multiple outsider deaths. This is **not** the ShortEvent format, there is no constraint aside from the flavour of outsider, and there is no way to make any death more or less common. I just thought it'd be nice for some ✨ Options ✨

I spotted hardcoded text, I obliterated.

## Proof of Testing

Working death event.
image


j-gynn:
oops.

scribblecrumb (approved):

scripts/events_module/outsider_events.py
```diff
@@ -22,32 +23,56 @@ class OutsiderEvents:
...
+                elif cat.status.is_lost(group_id):
+                    text = random.choice(deaths["other_clan_lost"])
+                    death_history = i18n.t(
+                        "events.death.outsider_deaths.history.other_clan_lost"
+                    )
+                else:
+                    text = random.choice(deaths["other_clan"])
+                    death_history = i18n.t(
+                        "events.death.outsider_deaths.history.other_clan"
+                    )
```
scribblecrumb: Okay, problem.  So when a cat is lost or exiled, they aren't considered a clancat, thus `is_other_clancat` won't find them.  However, `is_former_clancat` *will* find them (along with lost/exiled player clan cats, but at this point in the logic we've handled those two already). They also won't have a group_ID, since their group is currently None.  So what you'll want to do is go backwards through `group_history` till you find a not-none group_ID (might be a handy func to add to `status` tbh, something like: `get_last_valid_group_ID`?)

scribblecrumb (changes requested):

scripts/events_module/outsider_events.py
j-gynn: scribble i loathe you /silly

j-gynn:
Proof of otherclan cats successfully dying 
image
image

scripts/events_module/outsider_events.py
```diff
@@ -22,32 +23,56 @@ class OutsiderEvents:
...
+
+                clanname = [
+                    c for c in game.clan.all_other_clans if c.group_ID == group_id
+                ][0].name
+                clanname = i18n.t("general.clan", name=clanname)
+                text = text.replace("o_c_n", clanname)
+                death_history = death_history.replace("o_c_n", clanname)
+            elif not cat.status.is_outsider:
```
scribblecrumb: if *not* outsider? shouldn't it be the opposite, if this is meant to catch kittypets/rogues/loners?

j-gynn: ...oh fuck you're right.

scribblecrumb (approved):

PR #3842, opened 2025-07-23
Add displayname to Clan save data

larkgz:
## About The Pull Request

* Adds `displayname` field to Clan save data and Clan objects.
* `displayname` is used in UI to refer to the Clan's name instead of `name` (which is now effectively a unique ID referring to the Clan)
* Removes check during Clan creation that a Clan name doesn't exist. Instead, it allows it, sets the `displayname` to the existing name, and generates a new Clan ID using a uuid. The new Clan ID includes the name of the Clan for help with locating it when looking through folders.

I wanted to keep this PR small so it's simpler to review.  I'm hoping that I (or someone who is good at UI) can make improvements with separate PRs in the future.

This PR does not:
* Add any UI for changing `displayname` in-game  (side-note: this hypothetical future UI might want to make sure that the player's Clan not be named the same as any neighbouring Clans to help avoid confusion?)
* Display the `displayname` in the list of Clans when switching
image

* Makes it possible to rename Clans

## Proof of Testing
image

### DIFF
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -241,7 +241,7 @@ class Clan:
         save_cats(game.clan.name, Cat, game)
         number_other_clans = randint(3, 5)
         for _ in range(number_other_clans):
-            other_clan_names = [str(i.name) for i in self.all_clans] + [game.clan.name]
+            other_clan_names = [str(i.name) for i in self.all_clans] + [game.clan.displayname]
             other_clan_name = choice(
                 names.names_dict["normal_prefixes"] + names.names_dict["clan_prefixes"]
             )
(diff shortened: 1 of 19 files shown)


scripts/clan.py
```diff
@@ -76,14 +76,21 @@ def __init__(
...
+        displayname="",
     ):
         if name == "":
             return
 
         if starting_members is None:
             starting_members = []
 
+        # name is the unique id of the clan. i'm sorry if this is confusing...
         self.name = name
```
scribblecrumb, line 89: Definitely confusing ^^; any chance we could rename this attr to something like `self.ID` or `self.UUID`

scribblecrumb (changes requested):

scripts/clan.py
larkgz: I would prefer making refactors like this in a separate PR so it could be focussed on separately.

scribblecrumb (approved):

PR #3874, opened 2025-07-24
[FEATURE/CONTENT] Collar Palette Switching

scribblecrumb:
## About The Pull Request
This updates our collars to use palette maps for their color variations.  This means that we can quickly and easily add new colors and collar styles without significantly increasing future scope for sprite additions.

In the process, I've also updated the look of our collars to be more pleasing and in-line with our current art style guidelines.  In the process of ensuring old collars convert into similar new colors, I've ended up also increasing the collar variants we have! 

What has changed wrt to sprite files:
- We now have a single collar spritesheet, this holds every collar style variation. I'll refer to style_types and styles when talking about these. Each row represents a style_type: BOW, LEATHER, and NYLON respectively.  Each style_type has a multiple variations, which I call styles.
- You'll notice that these are colored mostly in grayscale with some elements brightly colored (or, in the case of gradient styles, in a heinous clown color scheme). The true colors that will be seen in-game will be mapped onto these colors.
- Within our sprites folder, we now have a palettes folder! This is where you can see the color palettes for each style. They are named accordingly: `collars{STYLE]_palette`.  
- If you open up a palette, you'll see that the top row of pixels matches the colors we see within the spritesheet, these pixels should never be changed unless you've made edits to the spritesheet.  The following rows represent individual color schemes!  You can add new rows to add new colors. 

So, after we've added new colors to the palette png, how do we add it in game?  We now have the handy dandy `collar_sprite_data.json` in `resources/dicts/sprites`.  The important part of this file is `style_data`.  You'll see that it lists a dict for each style_type and within the style_type are lists for each style.  This is where you would add the name of your new color.  Note that the list order matches the order of the rows in the palette png.  This file will be pulled in `sprites.py` and used to create all the collar sprites!  It will also be compiled into the collar list in `pelts.py`, so we don't have to worry about adding new collar sprite names in there.

In the end, the only thing you have to do to add a new color is edit a .png and a .json!  Easy peasy.

It's a bit more involved if you want to add a new style or style_type, but imo still relatively easy.  You would add the sprites for your new style into the spritesheet, using the corresponding row or starting a new row.  You should use grayscale colors when possible.  I brightly colored any distinctive parts, such as "metal" parts of the collars.  This just makes it easier for human eyes to recognize what parts are meant to be what.  Then you'll need to make a new palette .png following our naming convention.  You add all the colors you used in the spritesheet to the top row and add any color palettes you want below!  Then the .json will need to be edited with either a new list or a dict depending on if you're adding a new style or style_type.  

You've already noticed that the collar sprite names are much different.  They're piecemeal in the .json and are compiled together during sprite generation to create the name that will be used in-game.  I've tried to land on a naming convention that will be relatively scale-able and clear: `{STYLE_TYPE}_{STYLE}_{color}`.  Some names include multiple style descriptors, such as `LEATHER_BELL_SPIKE_crimson`. The color is lowercase to assist with style detection for profile display.

Speaking of profile display, we no longer display color as that could easily become unwieldy over time.  We already have some pretty long display names without colors.  I've set it up so that we display the collar style.  This means that, for example, `LEATHER_GRADIENT_rainbow` becomes "colorful leather collar" on profiles.

Makes it easier for us to keep up with sprite workload as scope increases.

## Proof of Testing

https://github.com/user-attachments/assets/9e0ae5a4-b218-4434-a37a-b2e5883a93a8

image

Acc name display!

image

You can see the acc name within this cat's save file to the right.  This is the old acc name.  The cat still displays the associated acc, meaning that save conversion is indeed working!

image

Documentation!

## Changelog/Credits
- We now handle the generation of collars in a much more scale-able way!  uh... sorries in advance to the mods this is gonna affect ^^;  This currently *is not* high-res mod friendly (it *could* be done, but it would probably be a pain in the ass), so high-res modders should stick to an older version.  Eventually we'd like to add a game config setting that would switch it to the old way of handling collars so that you can use that.
- The look of bows and collars has been updated!  And some new collar variations have been added :> If I counted right, we now have 99 collar/bow variants.


scribblecrumb:
Documentation added! this baby is ready to go

docs/dev/art/pallete-maps.md
j-gynn, line 1: misspelled "palette" in the file name <3

docs/dev/art/pallete-maps.md
```diff
@@ -0,0 +1,62 @@
...
+        "NYLON_BELL_GRADIENT": {
+            "zero": "colorful belled nylon collar",
+            "many": "colorful belled nylon collars",
+            "one": "colorful belled nylon collar"
+        }
+```
+
+`"zero"` is the profile display name while `"many"` and `"one"` are the plural and singular forms used when referencing this accessory in event text.
+
+## Adding a New Style_type
```
j-gynn:
```suggestion
## Adding a New `style_type`
```

scripts/cat/pelts.py
```diff
@@ -335,75 +334,21 @@ class Pelt:
...
-        "BLACKNYLON",
-        "SPIKESNYLON",
-        "WHITENYLON",
-        "PINKNYLON",
-        "PURPLENYLON",
-        "MULTINYLON",
-        "INDIGONYLON",
-    ]
+
+    collar_accessories = []
```
j-gynn, line 338: i know these *are* collars, but I can't help but wonder whether this should be `neck_accessories`, since all the other slots are based on body placement?

j-gynn (changes requested):

scripts/cat/pelts.py
scribblecrumb: I don't think so, as these lists correlate to the lists used for event reference.  Being able to mark an event as giving a collar, specifically, is more useful than marking it as giving a neck acc.  While atm our neck accs are *only* collars, we shouldn't be *restraining* neck accs to only be collars in the future.

j-gynn: Ahh I see, I didn't think of it that way!

j-gynn (approved):

j-gynn:
good luck modders 😘

PR #3889, opened 2025-07-26
[BUGFIX] Unequal dead_for counts

### LINKED ISSUE #3875: [BUG] dead moon count incorrect
### Pre-requisites

### Description

emberlight and haredew died in the same event. emberlight acts correctly, like he just died this moon. haredew acts like he died last moon

### Grade

1. make a clan
2. timeskip till a cat dies from an event (unsure if it is just this specific event or not)
3. check one of the cat's moons in death

_No response_

d6f1a983

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

Image

scribblecrumb:
## About The Pull Request
So! This was caused by double death events.  Essentially, if the second cat who dies hasn't already had their events loop, then when they *do* have their loop, they will increment their moons_as.  This means that they end up saying they've been dead for 1 moon even though they haven't.

To fix this I've added an attribute to status called `died_this_moon` that gets set to True when the cat changes to an afterlife. If this attr is True, then the cat won't increment their moons_as. It gets set back to False right after the increment check so that they'll start incrementing again next moon.

## Linked Issues
Fixes #3875

## Proof of Testing
image
image

0 moons on first moon of death

image

1 moon upon timeskip, we're incrementing again!

## Changelog/Credits
- The second dead cat within a single event no longer starts their dead_for count at 1


larkgz:
This fix seems to introduce another bug. Steps to reproduce: 
* Kill a cat. The cat is dead for 0 moons
* Timeskip one moon. The cat is still dead for 0 moons

scribblecrumb:
Ugh irritating how we have this stuff set up. Should be fixed now

larkgz (approved):

Issue #2087, opened 2024-02-12
removed


ZtheCorgi:
removed

ZtheCorgi:
also, tbh, i have been playing with the has_med_app check basically disabled in my game and i like the experience a lot better. maybe it could be changed so that, if has_med_app is true, it just reduces the chance of a cat becoming a medicine app rather than disallowing it outright? it's extremely helpful on large clans especially

<- guy who has 44 medicine cats

Lixxis:
I'm gonna change the chance after the next release, so the change causes no issues for the release. The skills seems to not trigger any error or game break, therefore it will as well remain for now.

PR #3913, opened 2025-07-30
Enhancement: apprentice ceremony tweaks and improvements

ZtheCorgi:
## About The Pull Request

-medicine cat and mediator apprenticeships now take into account skills as well as personality traits
-updated list of compatible traits and added incompatible traits for both medicine cats and mediators
-skills and traits now influence apprenticeships more strongly than disabilities do
-has_old_med and very_old_med checks no longer compound, and very_old_med procs properly
-a medicine cat apprentice can be made even when there is one already, it simply lowers the chances

it updates apprentice ceremonies to take advantage of the skillpath feature, which has been needed for a while. apprenticeship choices will feel more meaningful if skills and traits match up more often, especially now that the impact of being disabled is lower. additionally, it makes it more likely that a cat could become a medicine app out of passion, rather than solely out of need, which provides more diverse storytelling opportunities without overrunning the player with medicine apps.

## Linked Issues

#2087

## Proof of Testing

i have been playing with these changes for a long time and have not run into problems. definitely no crashes, and the rate at which medicine cats and mediators take on new apprentices feels good to me. i actually can't even play without these changes in my personal game anymore lol.

below are some screenshots showing the changes in action:

Screenshot from 2025-07-30 11-41-16
Screenshot from 2025-07-30 11-39-52

in the first set of screenshots here, i have enough medicine cats, and no mediators. the chance number is 1 in [chance]. in the prints, you can see that scrubkit got a bonus chance for being ambitious, but was made a warrior because the chance still wasn't high enough. on the other hand, you can also see that rockkit got the opposite treatment for being bloodthirsty.

Screenshot from 2025-07-30 14-30-41
Screenshot from 2025-07-30 14-30-14

in these screenshots, i only have a single medicine cat who is 151 moons old. as you can see, the chance is made much higher by this! even still, auburnkit's adventurous personality kept it a coin toss as to whether they'd become a medicine app. then, their adventurous personality and no compatible skills further prevented them from being likely to become a mediator apprentice, so they got to become a warrior app.

even though their sibling elderkit has the running skill, which would make them more likely to be a warrior, their secondary skill in combination with the heightened chances made it a guarantee that they would become a medicine cat apprentice instead!

now that elderpaw is a medicine apprentice, flamekit still has a pretty high initial chance to become a medicine app since there aren't enough healthy to care for the clan. however, because elderpaw is a medicine apprentice, that chance is further lowered, and flamekit does not become a medicine app.


ZtheCorgi:
let me know if i should remove the testing prints! its a quick fix, i just thought they might be useful for monitoring in case any tweaks need to be made to chances ^^

scripts/events.py
```diff
@@ -1245,43 +1246,141 @@ def perform_ceremonies(self, cat):
...
+                                print(f"Senior medicine cat")
+                                print(f"Not enough healthy medicine cats")
+                                print(f"Chance updated to {chance}")
+                                if has_med_app:
+                                    chance = int(chance * 3.5)
+                                    print(f"There is a medicine cat apprentice")
+                                    print(f"Chance updated to {chance}")
+
+                    if very_old_med == med_cat_list:
                         # These chances apply is all the current medicine cats are very old.
```
scribblecrumb: It looks like we only use `very_old_med` and `has_elder_med` when checking if all the current med cats qualify as elder/very old.  since we're never actually using these variables by themselves, but rather always comparing them to the `med_cat_list`, I think we might as well just be making them bools from the beginning.

You could do something like:
```python
elder_meds = all([c.age == CatAge.SENIOR for c in med_cat_list])
very_old_meds = all([c.moons >= 150 for c in med_cat_list])
```

This would help with readability and would mean you can remove that beginning `if not very_old_med` check.  Instead you can just go from `if elder_meds` to `elif very_old_meds` and so on.

scripts/events.py
```diff
@@ -1245,43 +1246,141 @@ def perform_ceremonies(self, cat):
...
-                        if has_med:
-                            chance = int(chance / 2.22)
-                        else:
-                            chance = int(chance / 13.67)
-                    elif very_old_med == med_cat_list:
+                    print(f"Medcat app {cat.name} starting chance: {chance}")
+                    if not very_old_med:
+                        if has_elder_med == med_cat_list:
+                            # These chances apply if all the current medicine cats are elders.
+                            if has_med:
```
scribblecrumb: I'd also like if we could rename this to `has_working_med`.  While reading I kept misinterpreting this as "a medicine cat exists" rather than "a medicine cat can work".

scripts/events.py
```diff
@@ -1245,43 +1246,141 @@ def perform_ceremonies(self, cat):
...
                     # These chances will only be reached if the
                     # Clan has at least one non-elder medicine cat.
                     elif not has_med:
                         chance = int(chance / 7.125)
+                        print(f"Not enough healthy medicine cats")
+                        print(f"Chance updated to {chance}")
+                        if has_med_app:
+                            chance = int(chance * 3.5)
+                            print(f"There is a medicine cat apprentice")
+                            print(f"Chance updated to {chance}")
```
scribblecrumb: It seems we always increase by the same modifier when `has_med_app` is true.  Rather than repeating this code block in three places, could it be moved out of the nested `if` statements to just always take place when a med app is present, regardless of other various scenarios?

scripts/events.py
```diff
@@ -1245,43 +1246,141 @@ def perform_ceremonies(self, cat):
...
-                            chance = int(chance / 2.22)
-                        else:
-                            chance = int(chance / 13.67)
-                    elif very_old_med == med_cat_list:
+                    print(f"Medcat app {cat.name} starting chance: {chance}")
+                    if not very_old_med:
+                        if has_elder_med == med_cat_list:
+                            # These chances apply if all the current medicine cats are elders.
+                            if has_med:
+                                chance = int(chance / 2.22)
```
scribblecrumb: To fix that issue you mentioned before wrt to chances becoming 0, you could utilize the `max()` func to prevent the int from becoming 0.

For example:
`chance = max(1, int(chance / 2.22))` will return whichever number is higher, 1 or the resulting int from our chance modifying.

scripts/events.py
```diff
@@ -1245,43 +1246,141 @@ def perform_ceremonies(self, cat):
...
+                        print(f"{cat.name} {cat.skills.primary.skill}")
+                        print(f"Chance updated to {chance}")
+
+                    if cat.is_disabled():
+                        chance = int(chance / 1.3)
+                        print(f"{cat.name} is disabled")
+                        print(f"Chance updated to {chance}")
 
                     if chance == 0:
                         chance = 1
```
scribblecrumb: Could shorten this to just:
```python
chance = max(chance, 1)
```

scripts/events.py
```diff
@@ -1300,33 +1399,101 @@ def perform_ceremonies(self, cat):
...
-
-                        # Only become a mediator if there is already one in the clan.
-                        if (
-                            mediator_list
-                            and not has_mediator_apprentice
-                            and not int(random.random() * chance)
-                        ):
-                            self.ceremony(cat, CatRank.MEDIATOR_APPRENTICE)
+
+                        if not mediator_list:
```
scribblecrumb: Instead of checking for `not mediator_list` to make warrior apps, passing to an `else` and then further nesting in another `else` for the final creation of warrior apps, could we instead get rid of this `if not meditor_list` block and have the following `else` instead be `if mediator_list`?  Then, we can follow it with a new `else` statement where warrior apps are made.  

This way we avoid some of the nesting hell we have going on, we remove a repetitive code block, and we increase readability.

scripts/events.py
```diff
@@ -1300,33 +1399,101 @@ def perform_ceremonies(self, cat):
...
+                                chance = int(chance / 1.3)
+                                print(f"{cat.name} is disabled")
+                                print(f"Chance updated to {chance}")
+
+                            if chance == 0:
+                                chance = 1
+                            print(f"Final mediator app chance {cat.name}: {chance}")
+
+                            # Only become a mediator if there is already one in the clan.
+                            if mediator_list and not int(random.random() * chance):
```
scribblecrumb: the `if mediator_list` check can be removed here if my prior suggestion is implemented.

scribblecrumb (changes requested):
Some suggestions! Mostly ways to combat repetative code and increase readability.

ZtheCorgi:
thanks for the suggestions! i definitely wanted to try to make it less of a mess, but was stuck on how to do so without breaking things, so i appreciate it. Will work on this uhh,, when my body and brain cooperate

ZtheCorgi:
alright, that should be all the changes that were requested, i believe! i have tested to make sure this code still functions properly ^^ i can get screenshots if needed though. there are a couple things i'm still unsure if i did the right way but it is certainly less scuffed than before lol

scripts/events.py
```diff
@@ -1243,45 +1231,149 @@ def perform_ceremonies(self, cat):
...
-                    elif not has_med:
-                        chance = int(chance / 7.125)
-                    elif has_med:
-                        chance = int(chance * 2.22)
+                    elif not has_working_med:
+                        chance = max(1, int(chance / 7.125))
+                        print(f"Not enough healthy medicine cats")
+                        print(f"Chance updated to {chance}")
+
+                    elif has_working_med:
```
scribblecrumb, line 1276: This could be changed to just a straight `else:`

scripts/events.py
```diff
@@ -1243,45 +1231,149 @@ def perform_ceremonies(self, cat):
...
+                        SkillPath.HEALER,
+                        SkillPath.STAR,
+                        SkillPath.DREAM,
+                        SkillPath.CLAIRVOYANT,
+                        SkillPath.GHOST,
+                        SkillPath.CAMP,
+                    ]:
+                        chance = max(1, int(chance / 2))
+                        print(f"{cat.name} {cat.skills.primary.skill}")
+                        print(f"Chance updated to {chance}")
```
scribblecrumb, line 1343: Rather than repeating the list of qualifying skills, I think it would be better to create a list variable for them like:
```python
qualifying_skills = [
SkillPath.OMEN,
SkillPath.PROPHET,
SkillPath.HEALER,
SkillPath.STAR,
SkillPath.DREAM,
SkillPath.CLAIRVOYANT,
SkillPath.GHOST,
SkillPath.CAMP]
```
then change the `if` statements to check against this list variable, so like:
`if cat.skills.secondary and cat.skills.secondary.path in qualifying_skills:`

This means any future changes to this list will only have to be changed in one spot, rather than two.

scripts/events.py
```diff
@@ -1243,45 +1231,149 @@ def perform_ceremonies(self, cat):
...
-                        chance = int(chance / 7.125)
-                    elif has_med:
-                        chance = int(chance * 2.22)
+                    elif not has_working_med:
+                        chance = max(1, int(chance / 7.125))
+                        print(f"Not enough healthy medicine cats")
+                        print(f"Chance updated to {chance}")
+
+                    elif has_working_med:
+                        chance = max(1, int(chance * 2.22))
```
scribblecrumb, line 1277: We shouldn't need to use `max()` when we multiply chance.  Division is where we chance the creation of a 0. As long as we're always doing division within `max()`, we know that the lowest number we'll multiply will always be 1.

scripts/events.py
```diff
@@ -1294,39 +1386,113 @@ def perform_ceremonies(self, cat):
...
+                                print(f"{cat.name} {cat.personality.trait}")
+                                print(f"Chance updated to {chance}")
+
+                            if cat.skills.secondary and cat.skills.secondary.path in [
+                                SkillPath.MEDIATOR,
+                                SkillPath.SPEAKER,
+                                SkillPath.INSIGHTFUL,
+                                SkillPath.CLAIRVOYANT,
+                                SkillPath.LORE,
+                            ]:
```
scribblecrumb, line 1450: Same critique here as with the med cat qualifying skills

scripts/events.py
```diff
@@ -1294,39 +1386,113 @@ def perform_ceremonies(self, cat):
...
+                                self.gain_accessories(cat)
+                                print(f"Made {cat.name} warrior apprentice")
+
+                        # this part should be reached whenever there are
+                        # no mediators
                         else:
                             self.ceremony(cat, CatRank.APPRENTICE)
                             self.ceremony_accessory = True
                             self.gain_accessories(cat)
+                            print(f"Made {cat.name} warrior apprentice, second check")
```
scribblecrumb, line 1495: Instead of a double check for creating warrior apps, how about we just move this out of the nesting a bit? We could align it with line `1245`'s indent and have a new `if` statement where we check if the cat is still a kitten rank like:

`if cat.status.rank == CatRank.KITTEN:`

Then assign them as a warrior apprentice.

scribblecrumb (changes requested):
Much improved! However the new ease of reading means I've found more things we could clean up ;)

ZtheCorgi:
yep, i figured. lol i didn't even think about that with max(), duh, that makes sense. i just kinda stuck it in front of everything

aaand yeah i was hoping you might suggest a more elegant way to make sure cats actually become warrior apprentices at the end of everything, i was getting a bit tired. thanks again!

scripts/events.py
ZtheCorgi: ah yeah, i wanted to do that. ill probably do more descriptive names since ill need different qualifying skills for medicine cats and mediators

scribblecrumb: eh idk about that, as long as the lists are being made right before their separate sections, there's no reason to give them different names as they'd never be referred to at the same time

ZtheCorgi:
closed until such time as there is a development server without a community side on which to communicate with fellow devs.

PR #3931, opened 2025-08-08
[CLEANUP] Killing windows.py

scribblecrumb:
## About The Pull Request
windows.py was becoming more and more of a monster, it's far past time to break her up. I've created a new `windows` folder within `ui` that holds each window class within a separate file.  I've also created a base window class to reduce repetition wrt to input blocking and exit button handling.

Very little about the classes themselves were changed beyond removing redundant code that's now covered by our base class. Some of them could definitely use refactoring, but that's not the purpose of this PR.  While the line changes certainly makes it look like lots happened here, I really did just copy-paste the old window classes into new files.

I was slowly losing my mind every time I had to venture into windows.py. It was hard to navigate and was only going to continue getting worse.

## Proof of Testing
image
image
image
image
image
image
image

## Changelog/Credits
- Window classes have been moved out of windows.py and into their own individual scripts within `ui/windows`


archanyhm:
Not necessarily a change request, more like a prompt for thoughts: Should we call all Window classes <SomePurpose>Window? The only thing really making clear that this is a Window we're dealing with is the import, so seeing something like `GameOver` doesn't really convey clearly that we're handling a Window instance here. What do you think? @scribblecrumb @j-gynn

scribblecrumb:
I'm not against it. We're certainly inconsistent on the matter, we do have some of the windows already named with Window and a couple with Popup.

Would you prefer that in this PR? I was planning to make a new PR after this one that would include the visual changes I've made to windows (cleaner borders and darkening everything behind the window), so I could just include the renaming in that one instead.

j-gynn (approved):

j-gynn:
I'm saying merge this in now and we can look at refactoring later (keeping the scope of a PR as small as practical)

j-gynn:
...that being said, please hold as I resolve the merge conflicts.

PR #3947, opened 2025-08-11
Fix pronoun test

j-gynn:
## About The Pull Request
*god* i hope this works. It should fix the pronoun test that was failing because sometimes we got trans kitties sneaking in.

also fixes a bug i noticed where medicine cats were under-estimating their skillz

## Proof of Testing
i ran the test 100,000 times and it didn't give me a failure once so if there's still errors i'm moving to the moon

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -244,11 +244,8 @@ class Cat:
 
         # age and status
         if status_dict is None and moons is None:
-            if disable_random:
-                self.age = CatAge.NEWBORN
-                self.status.generate_new_status(age=self.age)
-            self.age = choice([*CatAge])
-            self.status.generate_new_status(age=self.age)
+            self.age = CatAge.NEWBORN if disable_random else choice([*CatAge])
+            self.status.generate_new_status(age=self.age, disable_random=disable_random)
         elif moons is not None:
             self.moons = moons
             if moons > 300:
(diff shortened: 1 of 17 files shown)


scripts/cat/cats.py
```diff
@@ -241,7 +244,10 @@ def __init__(
 
         # age and status
         if status_dict is None and moons is None:
-            self.age = choice(list(CatAge))
+            if disable_random:
+                self.age = CatAge.NEWBORN
+                self.status.generate_new_status(age=self.age)
```
scribblecrumb: why don't we use the new status `disable_random` arg here?

scripts/cat/status.py
```diff
@@ -125,6 +125,7 @@ def generate_new_status(
         social: CatSocial = None,
         group: CatGroup = None,
         rank: CatRank = None,
+        disable_random=False,
```
scribblecrumb, line 128: Doesn't seem like you actually added this arg as True to any calls, do we actually need it?

scribblecrumb (changes requested):

scripts/cat/cats.py
j-gynn: ...this is what happens when you're coding at 11:30pm.

scripts/cat/status.py
j-gynn: it's nested waaaay down into `get_rank_from_age`, where it's used to force apprentice or warrior instead of picking a choice.

scribblecrumb: No I saw that, I meant that none of the calls to this func were setting the arg to True

scribblecrumb (approved):

PR #3965, opened 2025-08-13
[BUGFIX] Kittens inheriting lost status

### LINKED ISSUE #3926: [BUG] Kits born from taken-away birthing parent have full clan names
### Pre-requisites

### Description

If a pregnant cat is taken away by Twolegs and then gives birth, the resulting kits of the litter have a full warrior name, instead of using the -kit suffix.

### Grade

1. Have a cat in your Clan get pregnant
2. Get that pregnant cat to be taken away by Twolegs
3. Wait until their litter is born
4. Check the newborn kits' display names

_No response_

0c535153a

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

Image

### LINKED ISSUE #4040: [UI] Outsider sprites not aging properly (sometimes?)
### Pre-requisites

### Description

Sometimes outsider sprites do not age properly when moons are skipped. As can be seen in the attached screen recording- the same file can be used to either show the sprite aging- or not aging. When the magnifying class is clicked the correct sprite is shown regardless of what is shown in the normal cat page. (attached file for ease of testing)

### Grade

1. Have a clan with an outsider that is about to change sprites
2. save
3. repeat moon until their sprite does not change

_No response_

67db60d3

No

_No response_

_No response_

### Additional context

https://www.youtube.com/watch?v=9mmJZY7BqQE

[saves - outsider kit glitch.zip](https://github.com/user-attachments/files/22453919/saves.-.outsider.kit.glitch.zip)

### Screenshots

_No response_

scribblecrumb:
## About The Pull Request
This makes our handling of the kittens of lost cats much more robust.

- Kittens are now marked as lost if their birthing parent is lost.  NOTE: These kittens *will not* be given any prior Clan group history, meaning they aren't considered former Clancats. This is an important determiner when identifying them for handling.
- These kits will have both the new CatStanding.UNKNOWN and the CatStanding.LOST toward the player clan.
- CatStanding.UNKNOWN is used to determine if the Clan can interact with them via the Leader's den.
- The birthing parent can choose to abide by traditional Clan names, or to break from them. Their choice is indicated in the birth event.
- A new birth event specifically for lost cats was added, as the more general one didn't really work with their situation.
- If the parent keeps with Clan names, the kittens will gain appropriate suffixes as they reach appropriate ages.  There are no ceremony events written for these, though I think psuedo-ceremonies by the parent could be a fun thing to add later.
- These lost children can now return to the Clan with special event text, where they introduce themselves as being the kitten of the original lost parent.
- Added a blocker to prevent lost newborns and kitten from returning to the Clan (felt weird for them to make it alone, esp if the parent was still living and lost)
- Fixed a bug where outsider sprites wouldn't update with age.

## Linked Issues
Fixes #3926 

## Proof of Testing
image

Lost parent has kittens and leaves traditional naming behind.

image

Lost kid returns (this kid's parent *did* choose to name traditionally and Braveblink is currently an adult)

image

The "lost" kittens aren't seen in the leader's den

## Changelog/Credits
- Kittens are now marked as lost if their birthing parent is lost.  NOTE: These kittens *will not* be given any prior Clan group history, meaning they aren't considered former Clancats. This is an important determiner when identifying them for handling.
- These kits will have both the new CatStanding.UNKNOWN and the CatStanding.LOST toward the player clan.
- CatStanding.UNKNOWN is used to determine if the Clan can interact with them via the Leader's den.
- The birthing parent can choose to abide by traditional Clan names, or to break from them. Their choice is indicated in the birth event.
- A new birth event specifically for lost cats was added, as the more general one didn't really work with their situation.
- If the parent keeps with Clan names, the kittens will gain appropriate suffixes as they reach appropriate ages.  There are no ceremony events written for these, though I think psuedo-ceremonies by the parent could be a fun thing to add later.
- These lost children can now return to the Clan with special event text, where they introduce themselves as being the kitten of the original lost parent.
- Added a blocker to prevent lost newborns and kitten from returning to the Clan (felt weird for them to make it alone, esp if the parent was still living and lost)
- Fixed a bug where outsider sprites wouldn't update with age.


scripts/cat/names.py
```diff
@@ -265,6 +265,25 @@ def __repr__(self):
...
+            age_to_rank = {
+                CatAge.NEWBORN: CatRank.NEWBORN,
+                CatAge.KITTEN: CatRank.KITTEN,
+                CatAge.ADOLESCENT: CatRank.APPRENTICE,
+            }
+            rank = None
+            if self.cat.age in age_to_rank:
+                rank = age_to_rank[self.cat.age]
+            if rank:
+                return self.prefix + self.names_dict["special_suffixes"][rank]
```
j-gynn, line 281:
```suggestion
                rank = age_to_rank[self.cat.age]
                return self.prefix + self.names_dict["special_suffixes"][rank]
```
unless I'm misreading this, there's no way for `rank` to be non-null where the previous statement is `False`.

j-gynn (changes requested):
One very ancient comment to review.

j-gynn (approved):

PR #3968, opened 2025-08-13
[BUGFIX] grief assignment

### LINKED ISSUE #3966: [BUG] cats aren't grieving enough when another cat dies
### Pre-requisites

### Description

when a cat dies, there appears to be a lack of any emotional response from any of the other cats even if they liked the cat. no grief events, no grief thoughts. this has happened three times in my save but i only recorded one of them. i'm not sure if this is a frequency issue or the code is actually broken so i'm keeping it as an independent issue. 

### Grade

Dire - Still playable, but a major feature is broken

1. kill a cat that other cats have high green bars towards
2. timeskip
3. see error

_No response_

33aa1f59

### Error log

_No response_

### Additional context

_No response_

### Screenshots

two of the cats with the highest green bars for the dead in my most recent instance, plenty of other cats had high green bars too but these were the highest. there were no grief events or grief thoughts at all.

Image

Image

scribblecrumb:
## About The Pull Request
So my last bugfix for this fucked up a bit, but I swear it's working now!
Firstly, we weren't finding the rel_type correctly.  We also weren't checking for mid_pos or mid_neg! I've made mid_pos a 50/50 to give major grief chances.

Also unrelated to this bug, I realized that the rosemary use was actually making grief *more* common? So fixed that.

There was also a crash occurring when grief strings were accessed.

## Linked Issues
Fixes #3966

## Proof of Testing
image

We be grieving

## Changelog/Credits
- Cats can now grieve again.
- Rosemary treatment of body now makes grief *less* common, which was the original intention.
- Game no longer crashes when retrieving certain grief strings.


scripts/cat/cats.py
```diff
@@ -704,12 +704,15 @@ def grief(self, body: bool):
             # find what tier of rel they had for each type
             tiers: list[RelTier] = rel_with_dead.get_reltype_tiers()
             for tier in tiers:
-                rel_type = [k for k in rel_type_tiers if tier in k]
+                rel_type = [k for k in rel_type_tiers if tier in rel_type_tiers[k]]
                 if tier.is_extreme_pos:
                     very_high_types.extend(rel_type)
+                elif tier.is_mid_pos:
+                    list_to_extend = choice([very_low_types, very_high_types])
```
larkgz: `list_to_extend` doesn't seem to be used for anything.

larkgz (changes requested):

larkgz (approved):

PR #3973, opened 2025-08-14
[BUGFIX] Rel attr setting

scribblecrumb:
## About The Pull Request
Mostly just more instances of improperly settng rel attrs.  Also realized that we weren't *actually* adding to romance during romance events, nor were we applying rel increases when the intensity was anything other than high.... ugh

## Proof of Testing
Using debug mode to display what's happening:
image

Romance before interaction

image

Romance after the interaction

image

This is an interaction where respect is increased, here's respect before the increase

image

And respect after.

## Changelog/Credits
- Relationship events now properly apply relationship changes

### DIFF
--- a/scripts/cat_relations/relationship.py
+++ b/scripts/cat_relations/relationship.py
@@ -308,14 +308,24 @@ class Relationship:
             # so a negative interaction will affect all values to a negative degree
             # and a positive interaction will affect all values to a positive degree
 
+            if rel_type == RelType.ROMANCE:
+                self.romance += amount
+
             for rel_out in (
                 RelType.LIKE,
                 RelType.RESPECT,
                 RelType.TRUST,
                 RelType.COMFORT,
             ):
-                setattr(self, rel_out, choice(buffs) if rel_type != rel_out else amount)
-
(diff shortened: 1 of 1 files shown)


scripts/cat_relations/relationship.py
```diff
@@ -308,14 +308,24 @@ def interaction_affect_relationships(
...
                 RelType.COMFORT,
             ):
-                setattr(self, rel_out, choice(buffs) if rel_type != rel_out else amount)
-
+                setattr(
+                    self,
+                    rel_out,
+                    getattr(self, rel_out) + choice(buffs)
+                    if rel_type != rel_out
+                    else amount,
```
larkgz: It seems like this can still set the relationship to exactly `amount` (which seems to be how much the number is supposed to change by), is this intended?

scribblecrumb: Ah I think I was assuming the if else was isolated to `choice(buffs)` and didn't include the `getattr()`.  I've added parentheses now so that it works as intended.

larkgz (approved):

PR #3987, opened 2025-08-20
[BUGFIX] Only cats hating each other being eligible for affairs

KiraDragoness:
## About The Pull Request
Apparently, only cats who hated each other were passing the filter to be eligible as possible affairs, which was causing unmated love affairs to be basically impossible to trigger, and in turn making litters with outsiders much more common than they should.

## Proof of Testing
I debugged the problem adding a print with the list of possible affairs, with this fix the list doesn't appear as empty anymore:
image

## Changelog/Credits
BUGFIX: Cats can now have affairs with other cats without hating them


scripts/events_module/relationship/pregnancy_events.py
```diff
@@ -714,7 +714,7 @@ def get_second_parent(cat, clan):
                         if not p_rel.opposite_relationship:
                             p_rel.link_relationship()
                         p_rel_opp = p_rel.opposite_relationship
-                        if p_rel_opp.like < -20 and p_rel.like < -20:
+                        if not p_rel_opp.like < -20 and not p_rel.like < -20:
```
scribblecrumb: I think this would be easier to read if we just flipped the `<`, so:

```suggestion
                        if p_rel_opp.like > -20 and p_rel.like >  -20:
```

scribblecrumb (approved):

PR #4005, opened 2025-09-09
[FEATURE] Afterlife

larkgz:
## About The Pull Request

* Adds “afterlife acceptance” to history.
  * Cats who die will have flavour text relating to which afterlife they were sent to, with different feelings depending on if they were accepted to the afterlife without debate, accepted with some controversy, were rejected by the afterlife, or are kits.
  * The history text was written by Tiri and Snap, though I messed with the tenses a little. The tenses might still be wrong. 
* Adds simple “afterlife affinity” system. 
  * By default, a cat goes to where the guide is, but negative affinity with that afterlife may cause a cat to go to the opposite one instead.
  * Afterlife affinity is **not** currently modified anywhere in-game.
  * The system where an afterlife is determined may be changed according to further discussion, so just consider this a placeholder system for now.

* Starts work of expanding the afterlife in ClanGen

## Proof of Testing

image
image


scripts/cat/cats.py
```diff
@@ -496,6 +498,47 @@ def dead(self, die: bool):
...
+                        )
+                        self.status.send_to_afterlife(CatGroup.STARCLAN_ID)
+                        return
+                    # fine, they can go to the dark forest, but some cats don't like it
+                    self.history.add_afterlife_acceptance(
+                        CatGroup.DARK_FOREST, contentious=True
+                    )
+                # dark forest thinks this cat is ok
+                else:
+                    self.history.add_afterlife_acceptance(CatGroup.DARK_FOREST)
```
scribblecrumb:
```suggestion
                if game.clan.instructor.status.group == CatGroup.STARCLAN:
                    affinity = self.starclan_affinity
                    accepted_group = CatGroup.STARCLAN
                    rejected_group = CatGroup.DARK_FOREST
                    rejected_ID = CatGroup.DARK_FOREST_ID
                else:
                    affinity = self.dark_forest_affinity
                    accepted_group = CatGroup.DARK_FOREST
                    rejected_group = CatGroup.STARCLAN
                    rejected_ID = CatGroup.STARCLAN_ID
                    
                # afterlife does not like this cat
                if affinity < 0:
                    # might send them to the opposite afterlife instead
                    if random() < abs(affinity / 100):
                        self.history.add_afterlife_acceptance(
                            rejected_group, rejected=True
                        )
                        self.status.send_to_afterlife(rejected_ID)
                        return
                    # fine, they can go to afterlife, but some cats don't like it
                    self.history.add_afterlife_acceptance(
                        accepted_group, contentious=True
                    )
                # afterlife thinks this cat is ok
                else:
                    self.history.add_afterlife_acceptance(accepted_group)
```

This might be a preferable setup rather than essentially repeating the code block.

scripts/cat/history.py
```diff
@@ -410,6 +417,109 @@ def add_death(self, death_text, condition=None, other_cat=None):
...
+            "dark_forest_contentious_4",
+            "dark_forest_contentious_5",
+        ]
+        dark_forest_kit_text = [
+            "dark_forest_kit_0",
+            "dark_forest_kit_1",
+            "dark_forest_kit_2",
+            "dark_forest_kit_3",
+            "dark_forest_kit_4",
+        ]
```
scribblecrumb: I think it would be preferable to have these lists in a json.

scripts/cat/history.py
```diff
@@ -410,6 +417,109 @@ def add_death(self, death_text, condition=None, other_cat=None):
...
+        elif afterlife == CatGroup.DARK_FOREST:
+            if is_kit:
+                self.afterlife_acceptance = random.choice(dark_forest_kit_text)
+            elif contentious:
+                self.afterlife_acceptance = random.choice(dark_forest_contentious_text)
+            elif rejected:
+                self.afterlife_acceptance = random.choice(dark_forest_rejected_text)
+            else:
+                self.afterlife_acceptance = random.choice(dark_forest_default_text)
+
```
scribblecrumb: If the above lists are transferred to a json, then we could use fstrings like `random.choice(f"{afterlife}_kit_text"` instead of this repetition

scribblecrumb (changes requested):

scripts/cat/history.py
```diff
@@ -410,6 +423,46 @@ def add_death(self, death_text, condition=None, other_cat=None):
...
+        :param is_kit: `True` if the cat is a kit. Gives kinder acceptance text referring to kits.
+        :param contentious: `True` if the acceptance is supposed to be contentious. Afterlife will seem iffy about the cat.
+        :param rejected: `True` if cat is rejected from `guide_afterlife`. They will go to the opposite one instead.
+        """
+
+        afterlife = None
+        if guide_afterlife == CatGroup.STARCLAN:
+            afterlife = "starclan"
+        elif guide_afterlife == CatGroup.DARK_FOREST:
+            afterlife = "dark_forest"
```
scribblecrumb, line 443: the `CatGroup` enums will already return a string, so we don't need this extra step of creating the `afterlife` variable, we can just use `guide_afterlife`

scribblecrumb (changes requested):

scripts/cat/history.py
larkgz: I think that it's more clear this way, and I would prefer not to rely on enum strings in case that they change.

scribblecrumb (approved):

PR #4011, opened 2025-09-11
[TEST] Adding a bulk timeskip test

scribblecrumb:
## About The Pull Request
Lago asked for a test that just runs 1000 timeskips, I also wanted it so I made it. All in all it was surprisingly simple.

Helps catch timeskip crashes before they get merged!  1000 timeskips is bound to catch even the rarer crashes.

I think eventually we'll want to expand this out to test a variety of setting configurations, but for now it's better than nothing.

## Proof of Testing
image

It actually fails atm due to the current grief crash! So it works lmao

## Changelog/Credits
- Added a new test that runs 1000 timeskips to check for crashes

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -9,6 +9,7 @@ from scripts.clan import Clan
 from scripts.clan_package.settings import switch_clan_setting, set_clan_setting
 from scripts.events import events_class
 from scripts.game_structure import game
+from scripts.utility import get_living_cat_count
 
 
 class TestEvents(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_events.py
```diff
@@ -0,0 +1,62 @@
...
+                        [
+                            CatRank.KITTEN,
+                            CatRank.APPRENTICE,
+                            CatRank.WARRIOR,
+                            CatRank.WARRIOR,
+                            CatRank.ELDER,
+                        ]
+                    )
+                )
+                for _ in range(10)
```
larkgz, line 43:
```suggestion
                for _ in range(7)
```
To better match the amount of cats when making a new Clan.

larkgz (changes requested):

tests/test_events.py
scribblecrumb: Is that necessary though? I honestly just went with a round number for the heck of it, as well as figured that a larger number wouldn't hurt for suitability of the Clan.  idk that we *need* to replicate a player's starting Clan for this.

larkgz: That's fair.

larkgz (approved):

PR #4026, opened 2025-09-15
[CI] Fix Build Job Failure

larkgz:
## About The Pull Request

Currently, the build job is failing because broth on itch.io is returning invalid 404s. There isn't really anything we can do about that, since it's an issue on itch.io's end. However, we can at least stop the job from failing by not trying to download butler when we're not using it.

More green checkmarks are nice.


larkgz:
Sorry for force push, I forgot to switch branches while fixing another bug.

scribblecrumb (approved):

PR #4038, opened 2025-09-21
[CRASHFIX] Newborn conversion

### LINKED ISSUE #4037: [BUG] Game crash when conversing old save file with newborns
### Pre-requisites

### Description

After the new newborn sprites update, I tried loading an old Clan that had some newborns, and the game crashed when I tried going into the page where the newborn sprites had to be loaded. So I assume is a problem witht the conversion?

### Grade

Dire - Still playable, but a major feature is broken

1. Have a Clan with newborns before the newborn sprites update
2. Update the game
3. Load the old Clan with newborns
4. Try visualizing any of the newborns, either through the cat list screen or the individual cat screen.

_No response_

67db60d35

### Error log

Image

### Additional context

_No response_

### Screenshots

_No response_

### LINKED ISSUE #4051: [UI] newborn sprites default to same pose upon game reload
### Pre-requisites

### Description

clan_cats.json doesn't seem to specify the cat's newborn sprite unlike the other age stages, so even if newborns are generated with differing poses, upon saving & reloading the game, they all default to the same pose (the Lump)

### Grade

1. have pregnant cats in your clan
2. moonskip until the kits are born
3. view the kits (make sure they have different poses)
4. save the game & reload
5. view the kits again
6. see error

_No response_

f5216cc

No

_No response_

_No response_

### Additional context

_No response_

### Screenshots

Image

the kits look like this when first born...

Image

but they've become Lumps when the game is reloaded!

there is no "sprite_newborn" line in clan_cats.json, which is what i suspect is causing this.

Image

scribblecrumb:
## About The Pull Request
Gonna be real, I don't understand why this fixes the issue, but it does.  Without the `if age == "para_adult"` statement, the newborn int isn't converted.  with it, the conversion happens. My brain can't fathom why the para statement seemed to be blocking the newborn one, considering that it didn't block any of the other age checks.

Also switched it to convert to "newborn2", since that's the *actual* old newborn. but kept a little check in there to ensure mods that don't use multiple newborn poses would still be safe.

Also fixed an issue with newborn pose being saved and retrieved.

## Linked Issues
Fixes #4037 
Fixes #4051

## Proof of Testing
image

whistlekit is converting without a crash (ignore the "kitten" rank, i save edited this one's age so that the sprite could be tested. i didn't bother to edit the status too)

image

Before save

image

After re-opening the game

## Changelog/Credits
- Old newborns being converted to the new poses no longer crash
- Saved newborn sprites are now retrieved correctly


scripts/cat/pelts.py
```diff
@@ -253,7 +253,20 @@ def __init__(
...
+                    newborn_sprite,
+                    kitten_sprite,
+                    adol_sprite,
+                    adult_sprite,
+                    senior_sprite,
+                    para_adult_sprite,
+                ]
+                if isinstance(x, int)
+            ]
+        ):
```
larkgz, line 266:
```suggestion
        if any(
            isinstance(x, int) or x is None
            for x in [
                newborn_sprite,
                kitten_sprite,
                adol_sprite,
                adult_sprite,
                senior_sprite,
                para_adult_sprite,
            ]
        ):
```
* I think this order is a little more clear
* Probably also want to run conversions if any sprites are missing
* Apparently, according to pylint, the list comprehension inside of `any` is not required ([use-a-generator / R1729](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-a-generator.html))

scripts/cat/pelts.py
```diff
@@ -270,13 +283,17 @@ def __init__(
...
-                if self.length == "long":
-                    self.cat_sprites["para_adult"] = "para_adult_long0"
-                else:
-                    self.cat_sprites["para_adult"] = "para_adult_short0"
+                if age == "para_adult":
+                    if self.length == "long":
+                        self.cat_sprites["para_adult"] = "para_adult_long0"
+                    else:
+                        self.cat_sprites["para_adult"] = "para_adult_short0"
+                    continue
```
larkgz, line 289:
```suggestion
                if age == "para_adult":
                    if self.length == "long":
                        self.cat_sprites[age] = "para_adult_long0"
                    else:
                        self.cat_sprites[age] = "para_adult_short0"
                    continue
```
Let's make it a little more consistent with the rest of the checks.

scripts/cat/pelts.py
```diff
@@ -270,13 +283,17 @@ def __init__(
                     break
```
larkgz:
```suggestion
                    continue
```
`break` here prevented newborn code from running sometimes.

larkgz (changes requested):

larkgz (approved):

PR #4041, opened 2025-09-21
[BUGFIX] Herb amount not detected properly

larkgz:
## About The Pull Request

`herb_supply.entire_supply` returns a dictionary where all the herbs in the game are the keys, and the amounts of herbs owned by the player are the values. However, some checks worked under the assumption that the dictionary only contained herbs that the player owned, and were not functioning correctly as a result. 

This adjusts the checks to work with the actual structure of the dictionary.

## Proof of Testing
* Have no rosemary in Clan
* Kill cat
* No rosemary treatment in log

## Changelog/Credits

* Fixes issue where cats were using herbs that they didn't actually have for some treatments
* Fixes issue where events that were supposed to occur when cats had no herbs weren't able to be triggered (though, none exist)

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -766,7 +766,7 @@ class Cat:
                 if (
                     body
                     and not body_treated
-                    and game.clan.herb_supply.entire_supply["rosemary"] > 0
+                    and game.clan.herb_supply.entire_supply["rosemary"]
                 ):
                     body_treated = True
                     game.clan.herb_supply.remove_herb("rosemary", -1)
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2030,8 +2030,8 @@ class Cat:
             ):
                 clan_herbs = {
                     herb
-                    for herb, herb_count in game.clan.herb_supply.entire_supply.items()
-                    if herb_count > 0
+                    for herb, clan_has_herb in game.clan.herb_supply.entire_supply.items()
+                    if clan_has_herb
                 }
                 needed_herbs = {"horsetail", "raspberry", "marigold", "cobwebs"}
                 usable_herbs = list(needed_herbs.intersection(clan_herbs))
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -766,7 +766,7 @@ class Cat:
                 if (
                     body
                     and not body_treated
-                    and "rosemary" in game.clan.herb_supply.entire_supply
+                    and game.clan.herb_supply.entire_supply["rosemary"] > 0
                 ):
                     body_treated = True
                     game.clan.herb_supply.remove_herb("rosemary", -1)
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2028,7 +2028,11 @@ class Cat:
                 )
                 != 0
             ):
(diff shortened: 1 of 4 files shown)


scripts/cat/cats.py
```diff
@@ -766,7 +766,7 @@ def grief(self, body: bool):
                 if (
                     body
                     and not body_treated
-                    and "rosemary" in game.clan.herb_supply.entire_supply
+                    and game.clan.herb_supply.entire_supply["rosemary"] > 0
```
scribblecrumb: wouldn't:
```suggestion
                    and game.clan.herb_supply.entire_supply["rosemary"]
```
work? Since if the supply is zero, it should return false anyways

scripts/cat/cats.py
```diff
@@ -2028,7 +2028,11 @@ def get_injured(self, name, event_triggered=False, lethal=True, severity="defaul
                 )
                 != 0
             ):
-                clan_herbs = set(game.clan.herb_supply.entire_supply.keys())
+                clan_herbs = {
+                    herb
+                    for herb, herb_count in game.clan.herb_supply.entire_supply.items()
+                    if herb_count > 0
```
scribblecrumb:
```suggestion
                    if herb_count
```
Same here, since 0 is falsey and the herb supplies should never contain negatives, just checking `if herb_count` should be sufficient.  if that doesn't feel clear enough for future readers, you could renamed `herb_count` to `herbs_present`

scripts/clan_resources/herb/herb_supply.py
```diff
@@ -277,7 +277,7 @@ def get_overall_rating(self):
         """
         returns the rating of given supply, aka how "full" the supply is compared to clan size
         """
-        if not self.entire_supply:
+        if self.total <= 0:
```
scribblecrumb, line 280: Same here, self.total should never be negative anyways, so we can just check `if not self.total`

scripts/events_module/event_filters.py
```diff
@@ -222,7 +222,7 @@ def event_for_herb_supply(trigger, supply_type, clan_size) -> bool:
 
     herb_supply = game.clan.herb_supply
 
-    if not herb_supply.entire_supply and "empty" in trigger:
+    if herb_supply.total <= 0 and "empty" in trigger:
```
scribblecrumb, line 225: same thing

scripts/screens/MedDenScreen.py
```diff
@@ -593,7 +593,7 @@ def draw_med_den(self):
         herb_list = []
         herb_supply = game.clan.herb_supply
 
-        if not herb_supply.total:
+        if herb_supply.total <= 0:
```
scribblecrumb, line 596: Was this check not working? Unless I'm misunderstanding the way this translates to bool values, this change doesn't actually change this check.

scribblecrumb (changes requested):

scripts/clan_resources/herb/herb_supply.py
larkgz: I think that checking `self.total <= 0` is more readable for this value than checking for `not self.total`.

scripts/screens/MedDenScreen.py
larkgz: It's just for readability and consistency.

scribblecrumb (approved):

PR #4043, opened 2025-09-22
[FEATURE] Afterlife temperaments

scribblecrumb:
## About The Pull Request
This adds afterlife temperament values.  These currently don't affect the game in any way, we're just getting this stuff in piecemeal so that PRs are manageable.

This adds an `Afterlife` class which is used for the new `game.starclan` and `game.dark_forest` attributes.  This class tracks facets and temperament.  By default, all facets are set to 8, meaning that an empty afterlife will have an average facet default and stoic temper.

In order to avoid handling an insane number of dead cats and to ensure that the "important" dead ones are the ones with influence, we only allow leaders, deputies, medicine cats, and guides to affect facets.  

Rather than fully re-calcing facets on each timeskip, we keep track of cats who have changed afterlife via `game.updated_afterlife_cats`.  When we timeskip, each cat is checked that they're in:
- A new afterlife (to avoid weird interactions when a cat is flip flopped a bunch between afterlives)
- An appropriate rank (or the guide)
Then the cat is added to the `afterlife.influencing_cats` set and we do some math to add each of their facets to the afterlife's facet averages.  If the cat was removed from an afterlife, then they're also removed from it's `influencing_cats` and facet average

Continues on our journey to making affinities do things!  Once we know afterlife tempers, we can start determining what each afterlife thinks about each cat, and how that affects their affinity to that afterlife.

## Proof of Testing
I'll use the debugger to step through this, since nothing displays these tempers in-game yet. StarClan is started with a size of 5.

This was also done without the rank check, as that woulda been an annoying thing to account for when checking that the basic adding/removing statements are working.  However, I did test afterwards with the rank check and confirmed that only leaders, meddies, and deputies could make it through and change the facets.  I just wanted this debugging to display that I'd accounted for many different situations.

Cat exiled and sent to UR
image

image
image

SC and DF are unchanged (and you can see that DF is defaulting to average 8 values, since it's currently empty)

UR cat sent to SC
image

image
image

SC changed, DF did not

SC cat sent to DF
image

image
image

SC and DF changed, SC lost the cat and DF gained it

Cat killed and sent to SC
image

image
image

SC changed, DF did not

Cat killed and sent to DF (after initially being sent to SC)
image

image
image

DF changed, SC did not

Set of cats to update facets with is reset after all updates are applied
image

Now with the rank check on:
dead deputy
image
image

Changed SC as intended

Dead warrior
image
image

Gets continued instead of adjusting facets

## Changelog/Credits
- Adds temperaments & facets to StarClan and the Dark Forest based on the high ranking cats within each (guide, leaders, deputies, medicine cats).  Currently has no affect on gameplay and is not viewable in-game.


scripts/clan.py
```diff
@@ -1309,41 +1295,117 @@ def __repr__(self):
...
+            )
+
+    def new_average(self, old_avg: int, new_value: int, removal: bool) -> int:
+        if removal:
+            old_size = self.size + 1
+            new_value = -new_value
+        else:
+            old_size = self.size - 1
+
+        return int(((old_avg * old_size) + new_value) / self.size)
```
larkgz: Converting to int here will result in quickly accumulating roundoff errors, and may also give different averages for the same numbers depending on the order that they're added.

scripts/clan.py
```diff
@@ -1309,41 +1295,117 @@ def __repr__(self):
...
+                name,
+                self.new_average(_int, getattr(cat.personality, name), removal=removal),
+            )
+
+    def new_average(self, old_avg: int, new_value: int, removal: bool) -> int:
+        if removal:
+            old_size = self.size + 1
+            new_value = -new_value
+        else:
+            old_size = self.size - 1
```
larkgz: Can you pass in `old_size` as a parameter instead of recalculating it here?

scripts/clan.py
```diff
@@ -1309,41 +1295,117 @@ def __repr__(self):
...
+            ],
+            constants.facet_types,
+        ):
+            setattr(
+                self,
+                name,
+                self.new_average(_int, getattr(cat.personality, name), removal=removal),
+            )
+
+    def new_average(self, old_avg: int, new_value: int, removal: bool) -> int:
```
larkgz: * Please add a docstring to this function.
* If it's only going to be used as a helper in this class, please add an underscore to the name.
* I think `new_average` is a bit unclear. Maybe something like `_add_to_average` or something?

scripts/clan.py
```diff
@@ -1309,41 +1295,117 @@ def __repr__(self):
...
+                self.aggression,
+                self.stability,
+            ],
+            constants.facet_types,
+        ):
+            setattr(
+                self,
+                name,
+                self.new_average(_int, getattr(cat.personality, name), removal=removal),
+            )
```
larkgz: This is kind of hard to understand at a glance. 

Since it's all fixed attributes, I don't think that using setattr, getattr, and zip is strictly necessary here. There'll be some repetition, but I think that's a fair exchange for better readability.

scripts/clan.py
```diff
@@ -1309,41 +1295,117 @@ def __repr__(self):
...
+
+    @stability.setter
+    def stability(self, val):
+        self._stable = val
+
+    @property
+    def temperament(self) -> str:
+        return get_temper_alignment(self.sociability, self.aggression)
+
+    def change_facets(self, cat: Cat, removal: bool = False):
```
larkgz: * Please add a docstring to this function.
* Might want to rename `removal` to `is_removal` so it's more quickly understandable as a boolean?
* Maybe rename to `change_facets_by_cat`? `change_facets` sort of feels like it's setting all the facets to a given value.

scripts/events.py
```diff
@@ -99,6 +98,43 @@ def one_moon(self):
...
         game.clan.age += 1
         get_current_season()
+
+        # update afterlife tempers
+        for c in game.updated_afterlife_cats:
+            if not c.status.newly_joined_group:
+                continue
+            if (
+                not c.status.rank
+                in (
```
larkgz:
```suggestion
                c.status.rank
                not in (
```
Same thing but a bit easier to read.

scripts/cat/status.py
```diff
@@ -357,6 +357,13 @@ def is_other_clancat(self) -> bool:
     def is_leader(self) -> bool:
         return self.rank == CatRank.LEADER
 
+    @property
+    def newly_joined_group(self) -> bool:
```
larkgz: `joined_group_this_moon` might be clearer?

scripts/events.py
```diff
@@ -99,6 +98,43 @@ def one_moon(self):
...
+                # this cat never had their facets added to this group, so we won't try to remove them
+                continue
+
+            prior_group = c.status.group_history[-2]["group"]
+
+            if prior_group == CatGroup.STARCLAN_ID:
+                game.starclan.change_facets(c, removal=True)
+            elif prior_group == CatGroup.DARK_FOREST_ID:
+                game.dark_forest.change_facets(c, removal=True)
+        game.updated_afterlife_cats.clear()
```
larkgz: Not fully sold on this approach (i.e. looping over just updated_afterlife_cats and then checking history to correct any removals). It feels a bit like premature optimization that's more difficult to maintain. 

We already iterate over every cat, so imo it's fine to just recalculate the afterlife facets from everyone each moon.

larkgz (changes requested):
Please read the last comment first, since it's the most important.

scripts/events.py
scribblecrumb: Doing that math every moon regardless of whether or not the afterlife population changed seems wasteful though.  And players who turn off fading inevitably end up with large afterlives (even if fading is on, they can get pretty big), which would mean a lot of extraneous calculation occurring.

I do agree that the history checks are a little annoying to deal with. A different approach could be storing a list (actually a set would prolly be best) of IDs for each cat who influences the facets. Then we use the len for the average calc and we just check for the cat ID's presence when checking if their influence needs to be removed from the group.  That would be more maintainable in the long run, and easier to understand at a glance.

scripts/clan.py
```diff
@@ -1309,41 +1307,149 @@ def __repr__(self):
...
-                if cat.dead_for in range(
-                    self.forgotten_stages[f][0], self.forgotten_stages[f][1]
-                ):
-                    fade_level = f
-        white.fill((255, 255, 255, fade_level))
-        return white
+        if is_removal:
+            new_value = -new_value
+
+        return ((old_avg * old_size) + new_value) // new_size
```
larkgz: Unfortunately, this is still going to result in accumulated roundoff errors. The issue is that it's using the old average, which is already based off a rounded value, and then rounding the value again.

It might make sense to keep the totals (e.g. `self._total_aggression`) instead of retrieving them by multiplying `old_avg * old_size`.

scripts/clan.py
```diff
@@ -1309,41 +1307,149 @@ def __repr__(self):
         return f"{self.name}Clan"
 
 
-class StarClan:
+class Afterlife:
     """
-    TODO: DOCS
+    Currently just used for tracking temperament & facets. All facets default to 8 if dead_cats is empty.
```
larkgz:
```suggestion
    Currently just used for tracking temperament & facets. All facets default to 8 if influencing_cats is empty.
```

larkgz (changes requested):

scripts/clan.py
```diff
@@ -491,10 +492,22 @@ def load_clan(self):
                 Switch.error_message, "There was an error loading the clan.json"
             )
 
+        # can't put this in post initialization bc guide isn't made before that func
+        self.load_guide_influence()
```
larkgz: For some reason, it seems like the guide is being loaded into the afterlife influence another time, prior to this function being called. I have no idea where.

I think either:
* Remove this function
* Check when adjusting facets by cat if the cat is already an influencer or not. I guess one way of doing this would be checking if the length of the set is the same after trying to add/remove a cat.

scripts/clan.py
```diff
@@ -1309,41 +1307,150 @@ def __repr__(self):
...
+
+        self._law: int = 0
+        self._social: int = 0
+        self._aggress: int = 0
+        self._stable: int = 0
+
+        self.total_aggression: int = 0
+        self.total_lawfulness: int = 0
+        self.total_sociability: int = 0
+        self.total_stability: int = 0
```
larkgz: Can you rename these variables with an underscore in front?

scripts/clan.py
```diff
@@ -1309,41 +1307,150 @@ def __repr__(self):
...
+    def aggression(self) -> int:
+        if not self.influencing_cats:
+            return 8
+        else:
+            return self._aggress
+
+    @aggression.setter
+    def aggression(self, val):
+        self._aggress = val
+
```
larkgz, line 1340: Now that I think about it, I'm not sure if this class even needs public setters for law, sociability, aggression, and lawfulness, since setting these values from outside will throw off the average calculations.

Possibly print an error message instead?

larkgz (dismissed):

scripts/clan.py
scribblecrumb: Oh! Theory: I bet this is happening when the guide is a leader/dep/medicine cat.  Since those are the qualifiers for influencing temperament, they get included during the initial cat loading process.  Then when the guide is loaded, they get added to the temper *again*.  This isn't an issue for guides who are any other rank.  

So I think I will go ahead and check if the cat is already an influencer before adjustments are made, as this function is still necessary for those non-leadership role guides.

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
+    @property
+    def temperament(self) -> str:
+        return get_temper_alignment(self.sociability, self.aggression)
+
+    def adjust_facets_by_cat(self, cat: Cat, is_removal: bool = False):
         """
-        TODO: DOCS
+        Adjusts the afterlife's facet averages according to the facets of the given cat
+        :param cat: The cat object adjust facets by
+        :param is_removal: Set True if the cat's facets are being removed from the afterlife's
```
archanyhm: Perhaps rename `is_removal` to `do_remove`?

scripts/cat/personality.py
```diff
@@ -4,12 +4,14 @@
...
+from scripts.game_structure import constants
+
 
 class Personality:
     """Hold personality information for a cat, and functions to deal with it"""
 
-    facet_types = ["lawfulness", "sociability", "aggression", "stability"]
-    facet_range = [0, 16]
+    facet_types = constants.facet_types
+    facet_range = constants.facet_range
```
archanyhm, line 14: Not sure what the point is to move them into the `constants` file. Are they used anywhere else?

scripts/cat/status.py
```diff
@@ -357,6 +357,13 @@ def is_other_clancat(self) -> bool:
     def is_leader(self) -> bool:
         return self.rank == CatRank.LEADER
 
+    @property
+    def joined_group_this_moon(self) -> bool:
```
archanyhm: Perhaps rename to `did_join_group_this_moon` instead.
(More consistent with `is_`, `has_`, `do_`-nomenclature for boolean-members)

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
+        if not self.influencing_cats:
+            return 8
+        else:
+            return self._aggress
+
+    @aggression.setter
+    def aggression(self, value):
+        print(
+            "ERROR: Afterlife aggression cannot be set manually as it is meant to be calculated from the currently dead cats."
+        )
```
archanyhm: Raise exception instead.

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
+        if not self.influencing_cats:
+            return 8
+        else:
+            return self._social
+
+    @sociability.setter
+    def sociability(self, value):
+        print(
+            "ERROR: Afterlife sociability cannot be set manually as it is meant to be calculated from the currently dead cats."
+        )
```
archanyhm: Raise exception instead

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
+        if not self.influencing_cats:
+            return 8
+        else:
+            return self._law
+
+    @lawfulness.setter
+    def lawfulness(self, value):
+        print(
+            "ERROR: Afterlife lawfulness cannot be set manually as it is meant to be calculated from the currently dead cats."
+        )
```
archanyhm: Raise exception instead

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
+        if not self.influencing_cats:
+            return 8
+        else:
+            return self._stable
+
+    @stability.setter
+    def stability(self, value):
+        print(
+            "ERROR: Afterlife aggresstabilitysion cannot be set manually as it is meant to be calculated from the currently dead cats."
+        )
```
archanyhm: Raise exception instead

scripts/events.py
```diff
@@ -100,6 +99,45 @@ def one_moon(self):
...
+            # now do same for DF
+            elif (
+                c.status.group == CatGroup.DARK_FOREST
+                and c.ID not in game.dark_forest.influencing_cats
+            ):
+                game.dark_forest.adjust_facets_by_cat(c)
+                if c.ID in game.starclan.influencing_cats:
+                    game.starclan.adjust_facets_by_cat(c, is_removal=True)
+
+        game.updated_afterlife_cats.clear()
```
archanyhm: Pull into smaller helper function(s) since `one_moon` is already quite hefty.

scripts/clan.py
```diff
@@ -1309,41 +1307,160 @@ def __repr__(self):
...
-        fade_level = 0
-        if cat.dead:
-            for f in self.forgotten_stages:  # pylint: disable=consider-using-dict-items
-                if cat.dead_for in range(
-                    self.forgotten_stages[f][0], self.forgotten_stages[f][1]
-                ):
-                    fade_level = f
-        white.fill((255, 255, 255, fade_level))
-        return white
+        return total // num_of_influencers
```
archanyhm, line 1442: Probably rename to something that expresses more clearly that this relates to facets.

archanyhm (changes requested):

scripts/cat/personality.py
scribblecrumb: At the moment, no (though I think there was an iteration of this PR where I was using this it in the Afterlife class, which likely prompted this change), but we've been moving this sort of information to `constants` lately. I personally like having it consolidated in one file, but I understand if we'd rather not in this case.

archanyhm: I have to think about this some more. Given that this is relatively easy to refactor down the line, I'll mark this as resolved and make a note to bring it up during the code-style-guidelines-topic that we were talking about recently.

scripts/clan.py
```diff
@@ -491,10 +492,22 @@ def load_clan(self):
...
+        self.load_guide_influence()
         load_clan_settings()
 
         return version_info
 
+    @staticmethod
+    def load_guide_influence():
+        """
+        Adds guide's facet influences to their current afterlife
+        """
```
archanyhm: Potential nitpick, but does `load_guide_influence` really describe best what the function does? Perhaps `add_guide_facet_influence` or `apply_guide_facet_influence` would be more in-line?
(This is not a change-request; just an idea)

archanyhm (approved):
Not sure if @larkgz has anything else to say, but lgtm.

scripts/clan.py
archanyhm: (So if you disagree, feel free to just resolve this)

PR #4049, opened 2025-09-23
[BUGFIX] Negative Rel Events

scribblecrumb:
## About The Pull Request
There was an old magic string still hanging about preventing a bool value from working as intended, leading to negative events not having a negative impact on relationship values.  This fixes that and renames an attribute to make it's use clearer.

I also realized that relationships on the rel screen weren't sorted by absolute value, meaning that cats with negative rels were always at the end of the list (after cats with empty/low relationships).  Fixed that as a little QoL thing.

## Proof of Testing
image

Meant to be a positive change, returns a positive value to change rels by

image

Meant to be a negative change, returns a negative value to change rels by

image

Sungale no longer at the end of the line

## Changelog/Credits
- Negative relationship events now affect relationships negatively instead of positively
- On the relationship screen, cats with negative relationships are no longer sorted at the end of the list. Instead, cats are sorted by absolute total rel value, meaning that negative rel values are treated and sorted like positive ones.

### DIFF
--- a/scripts/cat_relations/relationship.py
+++ b/scripts/cat_relations/relationship.py
@@ -244,12 +244,12 @@ class Relationship:
 
         return process_text(string, cat_dict)
 
-    def get_value_change_amount(self, value_change: bool, intensity: str) -> int:
+    def get_value_change_amount(self, is_positive: bool, intensity: str) -> int:
         """Finds and returns the int amount that the relationship type will change by according to given intensity and additional modifiers
 
         Parameters
         ----------
-        value_change : bool
+        is_positive : bool
             True if the relationship value is positive, False if negative.
         intensity : str
             the intensity of the affect
@@ -261,7 +261,7 @@ class Relationship:
         """
         # get the normal amount
         amount = constants.CONFIG["relationship"]["value_change_amount"][intensity]
-        if value_change == "decrease":
+        if not is_positive:
             amount = amount * -1
 
         # take compatibility into account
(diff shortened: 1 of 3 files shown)


larkgz:
`change_according_dictionary` in `relationship.py` also seems to call the function using magic strings, can you fix that as well?

scripts/cat_relations/relationship.py
```diff
@@ -350,7 +350,10 @@ def change_according_dictionary(self, dictionary: dict) -> None:
         for key, value in dictionary.items():
             if value == "neutral":
                 continue
-            amount = self.get_value_change_amount(value, "low")
+
+            amount = self.get_value_change_amount(
+                is_positive=True if value == "positive" else False, intensity="low"
+            )
```
larkgz, line 356:
```suggestion
            amount = self.get_value_change_amount(
                is_positive=value == "positive", intensity="low"
            )
```

scripts/events_module/relationship/romantic_events.py
```diff
@@ -203,7 +203,7 @@ def start_interaction(cat_from, cat_to):
         relationship.used_interaction_ids.append(chosen_interaction.id)
 
         # affect relationship - it should always be in a romantic way
-        value_change = "increase" if positive else "decrease"
+        value_change = True if positive else False
```
larkgz:
```suggestion
        value_change = "increase" if positive else "decrease"
```
`value_change` is used later for some strings, so let's keep this and just pass `positive` into `interaction_affect_relationships` as-is, since it's already a boolean.

scripts/cat_relations/relationship.py
```diff
@@ -278,13 +278,13 @@ def get_value_change_amount(self, value_change: bool, intensity: str) -> int:
...
     def interaction_affect_relationships(
-        self, value_change: bool, intensity: str, rel_type: str
+        self, is_positive: bool, intensity: str, rel_type: str
     ) -> None:
         """Affects the relationship according to the chosen types.
 
         Parameters
         ----------
-        value_change : str
+        is_positive : str
```
larkgz:
```suggestion
        is_positive : bool
```

larkgz (changes requested):

larkgz (approved):

PR #4050, opened 2025-09-24
[FEATURE] Failed murders

scribblecrumb:
## About The Pull Request
This adds "failed" murders and updates the murder chance calculation.

Failed murders are events that occur when a murder *nearly* occurred, aka when the chance to carry out the murder is 20 or less.  When that chance is low enough, a failed murder event is triggered. These events are meant to sound a bit ominous and hint toward the potential murderer's intentions without being super overt. 

The chance calc to *consider* a murder hasn't been changed, however the chance calc to *carry out* a murder has been adjusted as such (keep in mind chance number is x/100 chance. a lower chance number means murder is more likely):
- Both mid neg and extreme neg rel tiers can cause a cat to be considered as a target
- All neg tier levels adjust the chance is their own ways. Each extreme tier lowers by 20, each mid tier by 10, and each low tier by 5,
- The last rel event being high negative now lowers chance by 20 instead of 50
- The last rel event being mid negative now lowers chance by 10 instead of 20
- Chance is lowered by the aggression facet's value
- Chance is lowered by the stability + lawfulness facets difference from 16 (so a lawfulness of 3 will lower the chance by 13)

(this was built on #4049 so that i wouldn't have to suffer trying to get murders to happen)

I think this creates an environment for more "understandable" murders. Previously, I think the presence of a high negative event in the log had an overly large weight. Cats generally shouldn't be murdering after a single high negative occurs.  Mid and low neg tiers were also being given the same weight, which I don't think reflects the relationship well.  We also just didn't take personality into account at all past the initial "will this cat consider a murder" chance, which seemed a missed opportunity.

Failed murder events have the benefit of "setting up" a possible event without actually setting a future event in stone.  Perhaps a cat just felt particularly murderous one moon and then calmed the fuck down, but also perhaps they're truly scoping out their victim. This gives the players more opportunity to build a storyline and to "investigate" murders if they're so inclined.

## Proof of Testing
image

Got a variety of murders to show up.  These felt decently spread out over the moons and seemed to nicely reflect relationships and events.

image

One of the failed murder events!

image

Aaaand uh oh.... looks like eventually someone got the courage to go through with it

## Changelog/Credits
- Adjusted murder chance calculation to better reflect relationship and personality
- Added "failed" murder events when a murder *nearly* occurs, but doesn't quite hit the chance. These might give you a hint as to who is contemplating drastic actions...


SunlitFable (approved):
writing all looks good! not confident speaking on the code though </3

archanyhm (approved):

Issue #3724, opened 2025-07-13
[BUG] cat thoughts mention they did actions they physically cannot do


cawcaw420908:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

a lonesome cat's thought says she "went out to hunt by herself" even though she had a mangled leg at the time and thus physically cannot do that 

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. have one or more lonesome cats
2. timeskip until one of those cats gets an injury
3. check the cat's thoughts
4. chance of error. 
5. if no error repeat 1-4 until you get the error

### Commit # or Game Version Number

bf1619b5

### Screenshots

<img width="731" height="543" alt="Image" src="https://github.com/user-attachments/assets/fa2a2648-e272-4d0d-b3ac-a7cdfc169ba2" />

cawcaw420908:
yet another example of this with a different cat

<img width="732" height="551" alt="Image" src="https://github.com/user-attachments/assets/9c43f411-9210-4c41-aee7-aaff6a3fa8dd" />

cawcaw420908:
instance 3
<img width="729" height="542" alt="Image" src="https://github.com/user-attachments/assets/20a85d2a-ae49-442d-8d9e-169ce0b4448d" />

cawcaw420908:
instance 4

<img width="729" height="549" alt="Image" src="https://github.com/user-attachments/assets/46126cec-6567-42f1-a41e-5425ae125cb3" />

cawcaw420908:
instance 5

<img width="736" height="559" alt="Image" src="https://github.com/user-attachments/assets/1accee82-6314-420a-bcd1-10fec0457e68" />

cawcaw420908:
instance 6

<img width="746" height="584" alt="Image" src="https://github.com/user-attachments/assets/aa06e9ab-f272-4678-a74c-967491e6e7e4" />

j-gynn:
I am unassigning @CL0WNTH0UGHTS to reopen this up for someone else to look at.

amilleah:
I tried to split thoughts into gen_ and working_ but found that it duplicated/ expanded the existing files which also makes it error-prone. I suggest instead adding `activity_type` tag to individual events, defaulting as plain string in thoughts:

```json
"thoughts": [
    "Isn't afraid of anything",
[...]

PR #4064, opened 2025-09-29
[FEATURE] New Thought Tags to Exclude Conditions

CL0WNTH0UGHTS:
## About The Pull Request
Adds new tags/logic to thoughts in order to exclude thought from cats with certain conditions/severity of conditions. These tags are:
* "not_{condition}" - If the cat(s) have a illness/injury of {condition}, then the thought is removed from their possible pool of thoughts.
* "not_any" - If the cat(s) have at least one illness/injury, then the thought is removed from their possible pool of thoughts.
* "not_{severity}" - If the cat(s)' illness/injury has a severity of {severity}, then the thought is removed from their possible pool of thoughts

Tags previously reported thoughts which would indicate a cat is going on patrol or leaving the camp with not_{severity} tags, in order to prevent cats with major/serve injuries from receiving them.

Removes the immersion breaking issue of cats with serious injuries, like head trauma or broken back, mentioning going on patrol or leaving camp. In theory, these cats wouldn't be going anywhere until their injury is healed and have been properly monitored. Also, allows writers to write more nuanced thoughts on injuries/illnesses; What if a cat has one condition but not another, what if the main cat has fleas but random cat doesn't, what if a cats grief stricken but the other isn't? 

## Linked Issues
#3724

## Proof of Testing
image
image

## Changelog/Credits
- Adds new options to thoughts, allowing them to be excluded based on condition and severity
- Fix (Some) thoughts mentioning patrolling or going outside, when the cat has a serious illness/injury

### DIFF
--- a/scripts/cat/thoughts.py
+++ b/scripts/cat/thoughts.py
@@ -184,8 +184,34 @@ class Thoughts:
                         and "any" not in thought["has_injuries"]["m_c"]
                     ):
                         return False
-                else:
-                    return False
+                    if [
+                        i
+                        for i in injuries_and_illnesses
+                        if f"not_{i}" in thought["has_injuries"]["m_c"]
+                        or (
+                            "not_any" in thought["has_injuries"]["m_c"]
+                            and injuries_and_illnesses
+                        )
+                    ]:
+                        return False
+
+                    for severity in ["minor", "major", "severe"]:
+                        if f"not_{severity}" in thought["has_injuries"]["m_c"]:
+                            if [
+                                i
+                                for i in injuries_and_illnesses
+                                if (
+                                    i in main_cat.injuries
+                                    and main_cat.injuries[i]["severity"] == severity
+                                )
+                                or (
+                                    i in main_cat.illnesses
+                                    and main_cat.illnesses[i]["severity"] == severity
+                                )
+                            ]:
+                                return False
+            else:
+                return False
 
             if "r_c" in thought["has_injuries"] and random_cat:
                 if random_cat.injuries or random_cat.illnesses:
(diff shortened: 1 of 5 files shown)


scripts/cat/thoughts.py
```diff
@@ -184,6 +184,32 @@ def cats_fulfill_thought_constraints(
                         and "any" not in thought["has_injuries"]["m_c"]
                     ):
                         return False
+                    if [
+                        i
+                        for i in injuries_and_illnesses
+                        if f"not_{i}" in thought["has_injuries"]["m_c"]
+                        or (
+                            "not_any" in thought["has_injuries"]["m_c"]
+                            and injuries_and_illnesses
```
larkgz, line 193: I think this part can be made a separate statement for readability:
```py
if injuries_and_illnesses and "not_any" in thought["has_injuries"]["m_c"]:
  return True
```

scripts/cat/thoughts.py
```diff
@@ -184,6 +184,32 @@ def cats_fulfill_thought_constraints(
                         and "any" not in thought["has_injuries"]["m_c"]
                     ):
                         return False
+                    if [
+                        i
+                        for i in injuries_and_illnesses
+                        if f"not_{i}" in thought["has_injuries"]["m_c"]
```
larkgz, line 190: Since we don't need to go through the whole list to check for this, a list comprehension is not necessary here. You can just return after you see the first instance of the illness.

For example, something like: 
```py
for i in injuries_and_illnesses:
  if f"not_{i}" in thought["has_injuries"]["m_c"]:
    return False
```

scripts/cat/thoughts.py
```diff
@@ -184,6 +184,32 @@ def cats_fulfill_thought_constraints(
...
+                                if (
+                                    i in main_cat.injuries
+                                    and main_cat.injuries[i]["severity"] == severity
+                                )
+                                or (
+                                    i in main_cat.illnesses
+                                    and main_cat.illnesses[i]["severity"] == severity
+                                )
+                            ]:
+                                return False
```
larkgz, line 212: Similarly to previous comments, I think a list comprehension here is also not necessary.

scripts/cat/thoughts.py
```diff
@@ -201,6 +227,32 @@ def cats_fulfill_thought_constraints(
                         and "any" not in thought["has_injuries"]["r_c"]
                     ):
                         return False
+                    if [
```
larkgz, line 230: See above comments

larkgz (changes requested):
A few things:
* Nice work on editing the JSON schema to reflect the changes. Thanks for that!
* There are some slight issues when it comes to using list comprehensions in places where it's not really necessary. I'm fairly certain that your logic is right, this is for some optimization and readability.
* Would it be possible to also add documentation somewhere under `docs/dev/writing/thoughts.md`?

scribblecrumb:
Also, a reminder that you'll need to update the Thoughts documentation your new tags! This is found at `docs/dev/writing/thoughts.md`

CL0WNTH0UGHTS:
Converting to Draft in order to fix requested changes + push upstream. Sorry for the month(s) long delay

scribblecrumb:
Closing as stale! The thought format has been changed, I would suggest referencing https://clangen.io/docs/dev/writing/thoughts/ for info on what it looks like now. You're welcome to adapt your changes into the new format and make a new PR!

PR #4073, opened 2025-10-04
[ENHANCEMENT] text in language settings buttons is no longer hardcoded

rustykitty:
## About The Pull Request
Previously, we had Spanish and German hardcoded on the language settings. This PR changes that, so with the exception of English, the buttons on the language settings page pull from a dict to be generated.

While experimenting with getting Chinese fonts to work, I noticed that the setting was hardcoded and not necessarily very easy to change to add a new language (copy paste, change button styles, etc).

## Linked Issues
N/A

## Proof of Testing
image


scribblecrumb (changes requested):
Always happy to see something get un-hardcoded! However, could we also consider moving the additional_langs dict out of the scripts? As it is, it's only nominally better than hardcoding. I'd love to see a json in the `lang` folder instead so that it's extra easy to locate and change for translators.

rustykitty:
image

scribblecrumb (approved):

PR #4077, opened 2025-10-05
[REFACTOR] JSON schemas to pydantic models.

archanyhm:
## About The Pull Request
Currently all resources are validated against JSON schemas, which are hand written. This is not necessarily super ergonomic and pydantic natively supports dumping models to schemas.

Makes adjusting the schemas easier

## Proof of Testing
See pipeline


pyproject.toml
```diff
@@ -57,3 +58,13 @@ discord = [
...
+
+[project.scripts]
+dump_model_schemas = "scripts.dump_model_schemas:main"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["scripts/"]
```
larkgz, line 70: What's hatchling being used for here?

tests/test_json_model_compliance.py
```diff
@@ -0,0 +1,86 @@
+"""
+Tests that JSON files are correct according to schemas.
+
+ Please do not put the *unittest* in the tests/unittest GitHub action.
+ It is only for local use.
+HOWEVER,
+ Please keep the raw python script, so it can be run by the GitHub action.
```
larkgz: I don't think this comment applies anymore, since this file no longer runs as a script.

scripts/dump_model_schemas.py
```diff
@@ -0,0 +1,25 @@
+import json
```
larkgz, line 2: Can you add a docstring for what this file does and how to run it easily?

tests/test_json_model_compliance.py
```diff
@@ -0,0 +1,86 @@
...
+
+    yield from chain.from_iterable(
+        RESOURCES_DIR.glob("lang/*/events/" + glob) for glob in INCLUSION_GLOBS
+    )
+
+
+@pytest.mark.parametrize(
+    "thought_file",
+    all_thought_files(),
+    ids=lambda thought_file: f'"{str(thought_file.relative_to(os.getcwd()))}"',
```
larkgz: * cwd can also be retrieved with `Path.cwd()`, if you want to skip the import of `os`
* The fstring will automatically convert the variable to a string, so `str()` is not required here
* Could these lambdas be moved to a single helper function since they're all the same?

scripts/models/common/relationship_status.py
```diff
@@ -0,0 +1,87 @@
...
+    runs_from_only = "runs_from_only"
+    fears_only = "fears_only"
+    avoids_only = "avoids_only"
+    considers_only = "considers_only"
+    seeks_out_only = "seeks_out_only"
+    prefers_only = "prefers_only"
+    relies_on_only = "relies_on_only"
+
+
+class RelationshipStatus(RootModel):
```
larkgz: I know that it's like this in the original schema, but you can probably remove the wrapping root model here and just use the enum.

scripts/models/patrol/herb.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+from enum import Enum
+
+
+class Herb(Enum):
```
larkgz: Could this have a different name, since it refers exclusively to non-specific herbs?

scripts/dump_model_schemas.py
```diff
@@ -0,0 +1,25 @@
...
+from scripts.models.patrol.patrol_schema import PatrolSchema
+from scripts.models.shortevent.short_event_schema import ShortEventSchema
+from scripts.models.thought.thought_schema import ThoughtSchema
+
+
+def dump_model_schema(config: BaseModel, output_name: str):
+    """Creates the JSON schemas based on the passed Pydantic model and writes it to a file"""
+    model_schema = config.model_json_schema()
+
+    with open(output_name, "w") as json_file:
```
larkgz:
```suggestion
    with open(output_name, "w", encoding="utf-8") as json_file:
```
Pylint wants encoding to be explicit

scripts/models/patrol/patrol_schema_item.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    max_cats: int = Field(
+        ..., description="Maximum total number of cats for this patrol"
+    )
+    min_max_status: Optional[Dict[str, List[int]]] = Field(
+        None,
+        description="Allows specification of the minimum and maximum number of specific types of cats that are allowed on the patrol.",
+    )
+    weight: int = Field(
+        ...,
```
larkgz, line 52:
```suggestion
        20,
```

scripts/models/patrol/patrol_schema_item.py
```diff
@@ -0,0 +1,77 @@
...
+        None,
+        description='If patrol_art contains gore, this line can hold a clean version. The existence of a non-empty string in this parameter marks the patrol art in "patrol_art" as explicit.',
+    )
+    min_cats: int = Field(
+        ..., description="Minimum total number of cats for this patrol."
+    )
+    max_cats: int = Field(
+        ..., description="Maximum total number of cats for this patrol"
+    )
+    min_max_status: Optional[Dict[str, List[int]]] = Field(
```
larkgz:
```suggestion
    min_max_status: Optional[Dict[str, Tuple[int, int]]] = Field(
```
Also, the keys should be one of: 
```
            "medicine cat",
            "warrior",
            "leader",
            "deputy",
            "apprentice",
            "medicine cat apprentice",
            "healer cats",
            "normal adult",
            "all apprentices"
```

scripts/models/patrol/outcome.py
```diff
@@ -0,0 +1,77 @@
...
+from scripts.models.patrol.herb import Herb as PatrolHerb
+from scripts.models.patrol.history_text import HistoryText
+from scripts.models.patrol.injury_item import InjuryItem
+from scripts.models.patrol.prey import PreyEnum
+from scripts.models.patrol.relationship import Relationship
+
+
+class Outcome(BaseModel):
+    text: str = Field(..., description="Displayed outcome text.")
+    weight: int = Field(..., description="Controls how common an outcome is.")
```
larkgz:
```suggestion
    weight: int = Field(20, description="Controls how common an outcome is.")
```

scripts/models/patrol/dead_cat.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+from enum import Enum
+
+
+class DeadCats(Enum):
```
larkgz: Can this be renamed to be more about leader lives specifically?

scripts/models/patrol/patrol_schema_item.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+    biome: List[BiomeEnum] = Field(
+        ..., description="Controls the biome(s) the patrol appears in"
+    )
+    season: List[SeasonEnum] = Field(
+        ..., description="Controls the season(s) the patrol appears in."
+    )
+    types: List[PatrolType] = Field(..., description="Controls the type of patrol.")
+    tags: List[
+        Union[PatrolTag, Annotated[str, StringConstraints(pattern=r"^clan:(.+)$")]]
```
larkgz: Can this union be moved into the PatrolTag model?

scripts/models/shortevent/history_text.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Optional, List, Union, Annotated
+
+from pydantic import BaseModel, Field, StringConstraints
+
+from scripts.models.shortevent.cats import Cats
+
+
+class HistoryText(BaseModel):
+    cats: Optional[
+        List[Union[Cats, Annotated[str, StringConstraints(pattern=r"^n_c:[0-9]+$")]]]
```
larkgz: Can this union be moved into the Cats model? It's exactly the same  as the one used in injuryitem, so I think it's ok.

scripts/models/shortevent/mc.py
```diff
@@ -0,0 +1,48 @@
...
+)
+from scripts.models.common.skill import Skill
+from scripts.models.common.status import Status
+
+
+class MC(BaseModel):
+    age: Optional[List[Union[CatAge, str]]] = Field(
+        None, description='List of ages m_c can be. If they can be anything, use "any".'
+    )
+    status: Optional[List[Union[Status, str]]] = Field(
```
larkgz: This should be a union between the statuses and the literal `any`.  Similarly applies for the other ones in this model.

scripts/models/patrol/injury_item.py
```diff
@@ -0,0 +1,26 @@
...
+from scripts.models.common.gather_cat import GatherCat
+from scripts.models.common.illness import Illness
+from scripts.models.common.injury import Injury
+from scripts.models.common.perm_condition import PermCondition
+from scripts.models.common.scar import Scar
+
+
+class InjuryItem(BaseModel):
+    cats: Optional[List[GatherCat]] = Field(None, description="Which cats are injured.")
+    injuries: Optional[List[Union[Injury, Illness, PermCondition, str]]] = Field(
```
larkgz: Should accept the literal `non_lethal` but not generic `str`.

scripts/models/thought/illness_injury_any.py
```diff
@@ -0,0 +1,12 @@
...
+from typing import Union
+
+from pydantic import RootModel
+
+from scripts.models.common.illness import Illness
+from scripts.models.common.injury import Injury
+
+
+class IllnessInjuryAny(RootModel):
+    root: Union[Illness, Injury, str]
```
larkgz: Should accept the literal `any` but not generic `str`.

scripts/models/thought/perm_condition_any.py
```diff
@@ -0,0 +1,11 @@
...
+
+from typing import Union
+
+from pydantic import RootModel
+
+from scripts.models.common.perm_condition import PermCondition
+
+
+class PermConditionAny(RootModel):
+    root: Union[PermCondition, str]
```
larkgz: Should accept the literal `any` but not generic `str`.

scripts/models/thought/status_any.py
```diff
@@ -0,0 +1,11 @@
...
+
+from typing import Union
+
+from pydantic import RootModel
+
+from scripts.models.common.status import Status
+
+
+class StatusAny(RootModel):
+    root: Union[Status, str]
```
larkgz: Should accept the literal `any` but not generic `str`.

scripts/models/shortevent/rc.py
```diff
@@ -0,0 +1,45 @@
...
+from scripts.models.common.all_trait import AllTrait
+from scripts.models.common.backstory import Backstory
+from scripts.models.shortevent.rc_relationship_status import RcRelationshipStatus
+from scripts.models.common.skill import Skill
+from scripts.models.common.status import Status
+
+
+class RC(BaseModel):
+    age: Optional[List[Union[CatAge, str]]] = Field(
+        None, description='List of ages r_c can be. If they can be anything, use "any".'
```
larkgz: This should be a union between the ages and the literal `any`. Similarly applies for the other ones in this model.

scripts/models/shortevent/short_event_schema_item.py
```diff
@@ -0,0 +1,67 @@
...
+        None, description="Controls the biome and camp the event appears in."
+    )
+    season: Optional[List[SeasonEnum]] = Field(
+        None, description="List of seasons in which the event may occur."
+    )
+    sub_type: Optional[List[EventSubTypeEnum]] = Field(
+        None, description="List of sub_types for this event."
+    )
+    tags: Optional[
+        List[Union[Tags, Annotated[str, StringConstraints(pattern=r"^clan:(.+)$")]]]
```
larkgz: Can this union be moved into the Tags model?

scripts/models/shortevent/supply.py
```diff
@@ -0,0 +1,26 @@
...
+    type: Optional[Union[Herb, SupplyType]] = Field(
+        None, description="Indicates the supply being affected."
+    )
+    trigger: Optional[List[EventTriggerEnum]] = Field(
+        None,
+        description="Indicates when the event can trigger. Must include all possible trigger times.",
+    )
+    adjust: Optional[
+        Union[
+            SupplyAdjust,
```
larkgz: Can this union be moved into the SupplyAdjust model?

larkgz (changes requested):
* Curious about why some classes have Enum in the name while some don't (e.g. `SeasonEnum`)?
* Apparently use of the ellipsis in Field is discouraged according to the Pydantic documentation: https://docs.pydantic.dev/latest/concepts/fields/
* I don't like the `None` default values for the optional fields, since it causes the schema to set `null` as default. Can they be removed? Default values should probably only be provided when an actual default value is specified.
* For some reason, all optional fields in the generated JSON schema are incorrectly set to allow `null`. Can this be fixed?
* It looks like the conversion didn't catch anything defined using `const` and instead just turned it all into the more general `str`. I tried commenting where that happens, but I might have missed some. Please double check if possible.

pyproject.toml
archanyhm: To use `project.scripts`, a build system with build targets has to be defined. Hatchling is relatively sane, so that's what I decided to use here.

I think `project.scripts` is a bit neater, because overall it'll allow us to make tools easily available via `uv` without having to rely on typing out the paths to the scripts, while giving us a way to logically separate the scripts from the rest of the code. This isn't done right now, because it makes little sense to have `scripts/scripts` or `scripts/bin`, but in a future refactoring PR where all of the ClanGen related code goes into `src/ClanGen`, everything could neatly fit into `src/ClanGen/scripts` or the likes. Alternatively we could also provide shell scripts, a Makefile or some other task runner to make execution of scripts possible, but I feel like this would just make life more complicated due to OS constraints (e.g. having to write POSIX-Shell-Scripts for *nix systems, PowerShell/Batch for Windows) or requiring yet another third party tool `make` or something like Nix.

scripts/dump_model_schemas.py
archanyhm: I agree with documenting how to run this script, but I don't really see why that should be a docstring for the file. Those only really seem to be useful when they're imported somewhere, and since this is an entrypoint script, that isn't the case. Additionally, I think once you stumble across this script (without looking for it), figuring out how to run it is relatively trivial. I can add a section to that in the README.md or a related file, however, on what it does and how to access it easily.

archanyhm:
> Curious about why some classes have Enum in the name while some don't (e.g. `SeasonEnum`)?

Mostly to blame on the output of the [datamodel-code-generator](https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/) — I only renamed enums when it included numbers (e.g. for the m_c and r_c relationship enums). Personally I'm a bit iffy on the entire "name includes type" scheme, so calling everything `ThingEnum` isn't necessarily my favourite, but we should probably find some sort of consensus on this.

---

> Apparently use of the ellipsis in Field is discouraged according to the Pydantic documentation: [docs.pydantic.dev/latest/concepts/fields](https://docs.pydantic.dev/latest/concepts/fields/)

Discouraged, yes. In our case I think we just need to choose what is more convenient. I don't think the models are necessarily there to be inspected regularly, they're more of an implementation detail for modeling the JSON Schemas and consequently the resources, so I value convenience more. If I understood it correctly, the alternative to:
```py
class SomeModel(BaseModel):
	required_field: str = Field(…, description="I'm a field!")
```
is
```py
class SomeModel(BaseModel):
	required_field: Annotated[str, Field(description="I'm a field")]
```
Which I find a bit awkward to read. But again, this is my personal opinion, so I'm more than fine with changing this.

---

> I don't like the `None` default values for the optional fields, since it causes the schema to set `null` as default. Can they be removed? Default values should probably only be provided when an actual default value is specified.

The answer is — unfortunately — "it depends". I don't think there's a stabilized way to easily express that a field should be optional in the sense that it either is included and conforms to a value, or it is not included (as opposed to either being `SomeValue` or `null`/`None`). Thankfully, as of Pydantic `2.12`, an experimental sentinel value of `MISSING` is introduced and can be used to express just this. While this is still not stabilized, I don't think there's any disadvantage for us in using it as of now.

Instead of marking a value as `Optional[str]` or `str | None`, it is marked as `str | MISSING` with a default of `MISSING`, which mirrors the JSON Schema optional fields closest. I've tested this by adjusting the `stat_skill` field in the `Outcome` class as discussed on Discord:
CleanShot 2025-10-08 at 10 41 20@2x
image
image
It is also not marked as required in the resulting schema:
image
And last, the resulting diff when changing from the previous `Optional[…]` to `… | MISSING = MISSING`
CleanShot 2025-10-08 at 10 43 21@2x

---

> For some reason, all optional fields in the generated JSON schema are incorrectly set to allow `null`. Can this be fixed?

See above

---

> It looks like the conversion didn't catch anything defined using `const` and instead just turned it all into the more general `str`. I tried commenting where that happens, but I might have missed some. Please double check if possible.

With `const` you're referring to literal strings as allowed values or something different? Can you elaborate?

scripts/models/patrol/patrol_schema_item.py
archanyhm: Resolved it differently to match the previous JSON Schema.
```py
    weight: int = Field(
        ...,
        description="Controls how common a patrol is. Normal patrols would be around 20. Lower numbers are less common and higher numbers are more common.",
        json_schema_extra={
            "default": 20
        },  # Necessary so that JSON Schema still shows a default without making the field optional
    )
    ```

File: scripts/models/patrol/patrol_schema_item.py [hunk 22]
```diff
@@ -0,0 +1,92 @@
...
+        None,
+        description='If patrol_art contains gore, this line can hold a clean version. The existence of a non-empty string in this parameter marks the patrol art in "patrol_art" as explicit.',
+    )
+    min_cats: int = Field(
+        ..., description="Minimum total number of cats for this patrol."
+    )
+    max_cats: int = Field(
+        ..., description="Maximum total number of cats for this patrol"
+    )
+    min_max_status: Union[Dict[MinMaxStatusDictKey, tuple[int, int]], MISSING] = Field(
```
larkgz: If we're supporting 3.8 with this (which I think we are since `Union` is being used instead of `|`) then `Tuple` has to be imported from `typing` instead of using `tuple`.

scripts/dump_model_schemas.py
larkgz: Sure. 

I think either add it to the README.md, or make a separate README file in `schemas` about how the files there aren't edited manually and should be generated using this script. I kind of prefer the second one, but that's just a slight preference.

larkgz (changes requested):
> Discouraged, yes. In our case I think we just need to choose what is more convenient. I don't think the models are necessarily there to be inspected regularly, they're more of an implementation detail for modeling the JSON Schemas and consequently the resources, so I value convenience more.

I don't love either of these, but I guess I prefer the first one.

> Thankfully, as of Pydantic 2.12, an experimental sentinel value of MISSING is introduced and can be used to express just this. While this is still not stabilized, I don't think there's any disadvantage for us in using it as of now.

Yeah, agreed here. 

> With const you're referring to literal strings as allowed values or something different? Can you elaborate?

When specifying a JSON schema, you can define properties using `const` to restrict a property to a single value https://json-schema.org/understanding-json-schema/reference/const. The converter doesn't seem to have caught any of these; it turned them all into `string`. I think we probably caught them all though? Hopefully?

pyproject.toml
larkgz: Makes sense to me. Can you add a comment on the toml for this since it's not immediately obvious?

(Sorry for deleting this and then posting basically the same thing again. I got confused because GitHub was showing it to me twice. But apparently they were the same comment.)

scripts/models/patrol/patrol_schema_item.py
archanyhm: On it. I've ran several inspections and lints while adjusting the types and neither of them (PyCharm, Pylint, Ruff) seemed to have caught this. I love this ecosystem.

scripts/models/patrol/outcome.py
larkgz, line 1: It seems like a few days ago, min_max_status was also added to outcomes. Could this be added?

scripts/models/patrol/patrol_schema_item.py
```diff
@@ -0,0 +1,92 @@
...
+        MISSING,
+        description="Allows specification of the minimum and maximum number of specific types of cats that are allowed on the patrol.",
+    )
+    frequency: int = Field(
+        ...,
+        description="Controls how common a patrol is. Normal patrols would be around 20. Lower numbers are less common and higher numbers are more common.",
+        json_schema_extra={
+            "default": 20
```
larkgz:
```suggestion
        description="Controls how common a patrol is. 4 is the most common, 1 is the least.",
        json_schema_extra={
            "default": 4
```

scripts/models/patrol/outcome.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class Outcome(BaseModel):
+    text: str = Field(..., description="Displayed outcome text.")
+    frequency: Annotated[
+        int,
+        Field(
+            description="Controls how common an outcome is.",
+            json_schema_extra={
+                "default": 20
```
larkgz:
```suggestion
                "default": 4
```

larkgz (approved):

PR #4085, opened 2025-10-10
[DEV] Declassify Events class

rustykitty:
## About The Pull Request
Convert the Singleton Events class into a module.

N/A

## Linked Issues
N/A

## Proof of Testing
image
Fresh moonskip.

## Changelog/Credits
N/A

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -64,6 +64,7 @@ CEREMONY_TXT = None
 WAR_TXT = None
 ceremony_lang = None
 war_lang = None
+ceremony_id_by_tag = {}
 
 def one_moon():
     """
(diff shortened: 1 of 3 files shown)


scribblecrumb (approved):
Very nice! Just resolve the merge conflict and then I think this is good to go.

Chinch-Bug:
There appears to be an infinite loop happening now, given the several hours of runtime for each attempt at tests (which usually take about a minute). It looks like the call to create a new cat short event was moved out of the if statement it was supposed to be in, so I suspect the issue is a conga-line of infinitely invited cats.

rustykitty:
> There appears to be an infinite loop happening now, given the several hours of runtime for each attempt at tests (which usually take about a minute). It looks like the call to create a new cat short event was moved out of the if statement it was supposed to be in, so I suspect the issue is a conga-line of infinitely invited cats.

Hmmm, I'll take a look at it. It was working fine before I merged with upstream.

archanyhm:
Next time, please run the test suite locally before pushing.

PR #4086, opened 2025-10-10
[BUGFIX] event generation

### LINKED ISSUE #4080: [BUG] accessory event repeats several times and incorrectly tags extra cats as involved
### Pre-requisites

### Description

a cat gives another an accessory. i scroll down a little and the exact same event with the exact same cats is repeated multiple times, but when i click the involved cats button on the duplicate, it incorrectly lists more random cats that have nothing to do with the event. the more it copies itself, the more cats it lists.

### Grade

1. moonskip
2. get the gen_misc_any_accessory2 event
3. check if it is duplicated with the exact same cats
4. click "involved cats" button on both to compare
5. see error

_No response_

25e9a72

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

first instance of the accessory event. simple enough, just a cat gifting an accessory another.

Image

but scroll down a little further, and not only is this the same event with the same cats and same accessory, but now it's adding two unrelated cats to the "involved" list. 

Image

third instance, still in the same moon, and now it's listing the same cats as before plus some extra ones.

Chinch-Bug:
## About The Pull Request

I've had reports of the new collar strings not being handled correctly in the event text itself. When testing if it worked correctly in basegame, I noticed that on the current version trying to force an accessory event always failed with no misc events found, despite having bypass requirements on.
So I've copied the logic from the profile screen in order to handle the collar accessory name in accessory events, and edited the filter_events function, which seemed to be only checking the first possible event

Debugging events is more doable now! (Also, I assume the loop not completing before selecting an event before probably made things extra repetitive, but I can't verify that)

## Linked Issues

should also fix #4080

## Proof of Testing

image

### DIFF
--- a/scripts/events_module/short/short_event.py
+++ b/scripts/events_module/short/short_event.py
@@ -84,6 +84,7 @@ class ShortEvent:
         self.sub_type = sub_type if sub_type else []
         self.tags = tags if tags else []
         self.text = text
+        self.text_template = text
         self.new_accessory = new_accessory if new_accessory else []
         self.m_c = m_c if m_c else {"age": ["any"]}
         if self.m_c:
@@ -215,6 +216,12 @@ class ShortEvent:
         :param other_clan: the object for the other clan involved in this event
         """
         self.additional_event_text = ""
+        self.text = self.text_template
+        self.all_involved_cat_ids: list[int] = []
+        self.new_cats = []
+        self.multi_cat_objects = []
+        self.dead_cat_objects = []
+        
         if other_clan:
             self.other_clan_name = f"{other_clan.name}Clan"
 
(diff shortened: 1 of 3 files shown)


scripts/events_module/short/short_event.py
```diff
@@ -215,6 +216,12 @@ def execute_event(self, other_clan=None):
         :param other_clan: the object for the other clan involved in this event
         """
         self.additional_event_text = ""
+        self.text = self.text_template
+        self.all_involved_cat_ids: list[int] = []
+        self.new_cats = []
+        self.multi_cat_objects = []
+        self.dead_cat_objects = []
```
scribblecrumb: Instead of redefining these, let's just use `.clear()`. This way attributes like `self.new_cats` will still have their documentation attached and we don't have to retype any of them (potentially annoying if we were to ever change what type smth like `all_involved_cat_ids` accepts, we'd have to change it in two places instead of one!)

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #4147, opened 2025-12-24
[CONTENT] Kit Relationship Events

tobes03:
## About The Pull Request

More relationship events that modify kits' relationships. I've focused on child/parent relationships. I was especially looking for places where it would make sense to increase one value but decrease another to promote more dynamic relationships (ie a strict parent might increase respect but lower comfort, while an overly easygoing parent might increase comfort but decrease trust)

Fills a need for more content to shape kits' relationships from early on, create more dynamic relationships, and generate relationship events that correspond to those cats' relationships.

## Proof of Testing

image


resources/lang/en/events/relationship_events/normal_interactions/like/increase.json
```diff
@@ -304,7 +344,10 @@
 	{
 		"id": "platonic_kit_inc_med5",
 		"interactions": [
-			"m_c ask r_c what it's like to be in training."
+			"m_c ask r_c what it's like to be in training.",
```
scribblecrumb:
```suggestion
			"m_c asks r_c what it's like to be in training.",
```

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #4156, opened 2025-12-28
[AUDIO] Updated music system

### LINKED ISSUE #3850: [AUDIO] audio setting does not save when you re-open the game
### Pre-requisites

### Description

I usually turn my audio off and assume that setting is automatically saved like all the other settings but every time the game is opened its back on

### Grade

Turn off sound
Close game
Open game

_No response_

d6f1a983

### Additional context

_No response_

### Screenshots/Video

_No response_

### LINKED ISSUE #3919: [AUDIO] Clangen cannot boot without an audio device detected.
### Pre-requisites

### Description

I was asked to assist someone who cannot play clangen due to the fact that Clangen cannot find their laptops audio output. The only solution seems to be for them to connect external devices such as AirPods, headphones, or speakers. The game will not even run at all without these external devices connected to their computer.

I have done some research on the official server and it would seem that many other users have encountered this problem and many people cannot play the game without an external device connected to their computer. There should be some code that allows the game to run without the need for an audio output. Music is not completely necessary for the game experience.

Image

This what they said about their computer: “(I was running the game on) Stable, on itch. I tried redownloading as win64 cause I had the windows 10 version and I’ve got windows 11 now”

They were only able to play the game once AirPods were connected to the laptop.

### Grade

Game-breaking - makes it impossible to play the game

Have no audio output detected on the computer. 
Launch clangen.

Most likely a hardware issue with the computer. But, clangen should still be accessible.

_No response_

0.12.2 Stable, on itch. 

### Additional context

I was not the one who encountered the bug, but nobody had reported the bug until now. I couldn’t let this sort of thing go under the radar like that.

### Screenshots/Video

Image

### LINKED ISSUE #4098: [AUDIO] clangen becomes unresponsive after muting the game
### Pre-requisites

### Description

sometimes when I try to mute ClanGen the application freezes and I have to force quit it. No warnings appear in the console. this doesn't happen all the time when i mute the game, only a few times, but i havent found anything in common with those few times.

### Grade

1. do some stuff inside a save
2. go back to the main menu
3. press the mute button
4. chance of error???? (step 1 may be optional)

_No response_

bcfeec54

### Additional context

_No response_

### Screenshots/Video

_No response_

scribblecrumb:
## About The Pull Request
This PR has been cooking for a while, it completely overhauls our audio system to better achieve the effect we want.

I've added a doc page giving a basic overview of the system, I'd suggest reading that to get an idea of the structure we're now working with.

The goal here was to create an audio system that allowed us to play both music and ambiance at the same time (not an easy feat with pygame's audio options). In the end, I've utilize the `pygame.mixer.music` to handle our main ambiance tracks (as these are always playing and thus need to be memory efficient) with occasional ambiance overlays playing via the sound channels.  Another dedicated sound channel is used to play music tracks at random intervals. Much of the timing here is controlled via timer threads.

This is something that's been in discussion for a long time and finally brings us closer to our goals for the audio.

fixes #4098

## Proof of Testing
[files too big so discord link](https://discord.com/channels/1003759225522110524/1101283866544701470/1454931829718126839)

## Changelog/Credits
- Unique ambient noise tracks added for all biomes. Certain camps and seasons have unique sound bites that will play at random intervals.
- 6 new music tracks added. These will play at random intervals during the game.
- Music, Ambiance, and Sound Effect volumes can now all be adjusted individually
- A lot of backend work to make this happen! Our audio system has been made more robust through a lot of painful wrestling and headaches.
- New information added to our audio page on the doc site!


scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+
+        self.playlist_dict: dict = {}
+
+        self.current_playlist: list = []
+
+        self.biome_playlist: list = []
+        self.season_overlay_playlist: list = []
+        self.camp_overlay_playlist: list = []
+
+        self.volume: int = game_setting_get("ambiance_volume") / 100
```
j-gynn: we're calling this an `int` but then doing division of a number, which might result in a float

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+        self.load_playlists()
+
+    def load_playlists(self):
+        """
+        Loads the ambiance playlists
+        """
+        # loading playlists
+        try:
+            with open("resources/audio/ambiance.json", "r", encoding="utf-8") as f:
+                self.playlist_dict = ujson.load(f)
```
j-gynn, line 52: Would prefer this to load to a temporary variable if it's not valid in its preloaded form

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            changed = True
+            self.start_overlay()
+
+        if changed:
+            self.ready_playlist(playlist)
+            self.fade_out()
+
+    def ready_playlist(self, playlist):
+        """
+        loads and plays random file from playlist, queues up next track
```
j-gynn, line 96: add docustring for what `playlist` should be in this format:

`:param playlist: [description]`

thank you

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+
+    def stop_overlay(self, fadeout: bool = True):
+        """
+        Stops any overlay sounds from playing
+        :param fadeout: set True if playing sounds should fade, False if they should stop immediately
+        """
+        if fadeout:
+            fade = 300
+        else:
+            fade = 0
```
j-gynn:
```suggestion
        fade = 300 if fadeout else 0
```

can this variable align with `fade_ms` found elsewhere?

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            self.camp_silence_timer.cancel()
+        if self.season_timer and self.season_timer.is_alive():
+            self.season_sound.fadeout(fade)
+            self.season_timer.cancel()
+        elif self.camp_silence_timer:
+            self.season_silence_timer.cancel()
+
+    def play(self, track, fade_ms=1000):
+        """
+        plays the given track and sets volume
```
j-gynn, line 137: `:param:` for `track` and `fade_ms` please

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            )  # don't want to repeat current track, so we take it out
+            options = playlist_copy
+        else:
+            options = self.current_playlist
+
+        try:
+            self.queued_track = random.choice(options)
+
+        except IndexError:
+            print("WARNING: playlist is empty")
```
j-gynn:
```suggestion
            logger.warning("Playlist is empty")
```

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+        if not self.queued_track:
+            return
+
+        self.play(self.queued_track, fade_ms=3000)
+        self.set_queued()
+
+    @staticmethod
+    def fade_out(fadeout=2000):
+        """
+        fades the ambiance out, by default the fade is 2 seconds
```
j-gynn, line 185: `:param:` for `fadeout` please

also consider changing it to `fade_ms` to align with other variables describing the same thing

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+    def unmute(self):
+        """
+        unpauses current ambiance track, then double checks if the track is appropriate for the screen before changing
+        if necessary
+        """
+        pygame.mixer.music.unpause()
+        self.check()
+
+    def change_volume(self, new_volume):
+        """changes the volume, int given should be between 0 and 100"""
```
j-gynn: `:param:` for `new_volume` please

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+        Finds the clan's biome and returns the appropriate playlist
+        """
+        try:
+            biome = game.clan.biome
+        except AttributeError:
+            biome = "Forest"
+
+        try:
+            camp: str = game.clan.camp_bg
+            index = int([x for x in camp if x.isdigit()][0]) - 1
```
j-gynn, line 238: Why do we do this when we just use the `camp1` format everywhere else? if it's for storing a unique name, can't we prepend the biome like `Forest_camp1` or `Forest:camp1`? seems like a lot of effort for little reward

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+
+    @staticmethod
+    def get_busy() -> bool:
+        """
+        Check if ambiance is playing.
+        """
+        return pygame.mixer.music.get_busy()
+
+    @property
+    def overlay_volume(self):
```
j-gynn, line 283: could you add a docustring for what this is? I know it's technically a variable but it's kind of confusing what it does <3

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            return
+        self.season_silence_timer = AudioTimer(
+            random.randint(20, 50), self.play_season_overlay
+        )
+        self.season_silence_timer.daemon = True
+        self.season_silence_timer.start()
+
+    def play_season_overlay(self):
+        """
+        Plays the season overlay and beings its timer.
```
j-gynn:
```suggestion
        Plays the season overlay and begins its timer.
```

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        self.season_sound = random.choice(self.season_overlay_playlist)
+        self.season_sound.set_volume(self.overlay_volume)
+        if pygame.mixer.find_channel():
+            print("played season overlay")
```
j-gynn:
```suggestion
            logger.info("season overlay started")
```

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            return
+        self.camp_silence_timer = AudioTimer(
+            random.randint(20, 50), self.play_camp_overlay
+        )
+        self.camp_silence_timer.daemon = True
+        self.camp_silence_timer.start()
+
+    def play_camp_overlay(self):
+        """
+        Plays the camp overlay and beings its timer.
```
j-gynn:
```suggestion
        Plays the camp overlay and begins its timer.
```

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        self.camp_sound = random.choice(self.camp_overlay_playlist)
+        self.camp_sound.set_volume(self.overlay_volume)
+        if pygame.mixer.find_channel():
+            print("played camp overlay")
```
j-gynn:
```suggestion
            logger.info("camp overlay started")
```

scripts/game_structure/audio/ambiance.py
```diff
@@ -0,0 +1,352 @@
...
+            self.camp_sound.play(fade_ms=4000)
+            self._start_camp_overlay_timer(self.camp_sound.get_length())
+        # TODO: what happens if no channel found?
```
j-gynn, line 360: What's been done about this?

scripts/game_structure/audio/audio_manager.py
```diff
@@ -0,0 +1,75 @@
...
+        if not self.ambiance.get_busy():
+            self.ambiance.check()
+            self.ambiance.play_queued()
+
+        if not self.music.get_busy():
+            self.music.check()
+
+    def check(self, fade_out: bool = False):
+        """
+        Checks that background audio is appropriate for the current screen
```
j-gynn, line 40: `:param:` for `fade_out` please

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def choose(self):
+        """
+        chooses music from the appropriate playlists and sends it to be loaded
+        """
+        self.find_playlist()
+
+        if not self.current_playlist:
+            logger.error("Music track list is empty, check the music.json!")
+            chosen_track = "resources/audio/music/Generations.mp3"  # making this default just in case
```
j-gynn: This should probably crash instead of attempting to recover.

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        elif len(self.current_playlist) == 1:
+            chosen_track = self.current_playlist[0]
+        else:
+            if self.last_track_name in self.current_playlist:
+                self.current_playlist.remove(self.last_track_name)
```
j-gynn: this permanently removes it from `current_playlist`. When does `current_playlist` get refreshed/reset?

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+            chosen_track = choice(self.current_playlist)
+
+        self.loaded_track = pygame.mixer.Sound(chosen_track)
+        self.current_track_name = chosen_track
+        print(chosen_track)
+
+    def find_playlist(self):
```
j-gynn, line 97: docustring pls

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+                    )
+                )
+            if self.available_music.get(f"{biome.casefold()}_playlist"):
+                self.current_playlist.extend(
+                    self.available_music.get(f"{biome.casefold()}_playlist")
+                )
+
+    def check(self, fade_out: bool = False):
+        """
+        checks if loaded music is appropriate for the given screen and stops playback if needed
```
j-gynn, line 133: `:param:` for `fade_out` please

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        # the mute button it likely won't be noticeable
+        self.check()
+
+        if self.loaded_track:
+            self.channel.unpause()
+            # making sure all timers are cleared first
+            self._stop_timers()
+            self._start_music_timer(self.remaining_time_of_paused_track)
+
+    def fade_out(self, fadeout=2000, delay=randint(30, 300)):
```
j-gynn, line 202: I don't think you should have a default value that's a random integer function but if PyCharm didn't scream at you it's probably fine?

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        between the two, otherwise the fade will get cut off or weird duplication will occur.
+        """
+        if self.channel and self.channel.get_busy():
+            self.channel.fadeout(fadeout)
+            self._start_silence_timer(delay)
+            self.music_timer.cancel()
+
+    def change_volume(self, new_volume):
+        """
+        changes the voume, int given should be between 0 and 100
```
j-gynn: `:param:` for `new_volume` please

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        elif new_volume < 0:
+            new_volume = 0
+
+        self.volume = new_volume / 100
+        game_setting_set("music_volume", new_volume)
+        self.loaded_track.set_volume(self.volume)
+
+    def _start_music_timer(self, duration=None):
+        """
+        sets a timer for the length of the track.  When the timer ends, silence timer is activated.
```
j-gynn, line 229: `:param:` for `duration`

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        if new_volume > 100:
+            new_volume = 100
+        elif new_volume < 0:
+            new_volume = 0
+
+        self.volume = new_volume / 100
+        game_setting_set("music_volume", new_volume)
+        self.loaded_track.set_volume(self.volume)
+
+    def _start_music_timer(self, duration=None):
```
j-gynn, line 227: I don't think `duration` should have a default value.

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        if self.music_timer and self.music_timer.is_alive():
+            return
+        if not duration:
+            duration = self.loaded_track.get_length()
+        self.music_timer = AudioTimer(duration, self._start_silence_timer)
+        self.music_timer.daemon = True
+        self.music_timer.start()
+
+    def _start_silence_timer(self, duration=randint(30, 300)):
```
j-gynn, line 240: I don't think duration should have a default value

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+            self._stop_timers()
+            self._start_music_timer(self.remaining_time_of_paused_track)
+
+    def fade_out(self, fadeout=2000, delay=randint(30, 300)):
+        """
+        fades the music out and begins the silence timer to count down to next track play
+        :param fadeout: length of fadeout in milliseconds
+        :param delay: this is used to dictate the length of the silence timer that begins when the music starts to fade.
+        If a new music track is going to play immediately after the initial track fades, then you need a small delay
+        between the two, otherwise the fade will get cut off or weird duplication will occur.
```
j-gynn:
```suggestion
        :param delay: Dictates the length of silence between this track and the next one, including fade time.
```
I think it would make more sense to set the `delay` to `max(delay, fadeout)`, so that the delay timer can always account for how long the fade out is. Also, what unit is `delay` in?

scripts/game_structure/audio/music.py
```diff
@@ -0,0 +1,255 @@
...
+        if not duration:
+            duration = self.loaded_track.get_length()
+        self.music_timer = AudioTimer(duration, self._start_silence_timer)
+        self.music_timer.daemon = True
+        self.music_timer.start()
+
+    def _start_silence_timer(self, duration=randint(30, 300)):
+        """
+        Clears old music, then sets a timer for the next track to play.  When the timer ends, new music begins.
+        :param duration: length of silence in seconds, by default this is a random duration between 120 and 300 seconds
```
j-gynn, line 243: PLEASE put this in milliseconds I am BEGGING you

for consistency across all our systems it's so weird for some to be using milliseconds and some to be using seconds, especially when this isn't indicated in the variable name.

scripts/game_structure/audio/sound.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def handle_sound_events(self, event):
+        """
+        assigns universal sound effects to event.type objects
+        SHOULD NOT BE USED FOR INDIVIDUAL UNIQUE BUTTON SOUNDS
+        UIImageButtons have a sound_id parameter for assigning unique sounds to individual buttons
+        :param event: the event that is taking place
+        """
+        # This makes sounds play using UI_BUTTON_PRESSED, instead of UI_BUTTON_START_PRESS
+        try:
```
j-gynn, line 55: Why is this wrapped in a blank try-except block?

scripts/game_structure/audio/sound.py
```diff
@@ -0,0 +1,103 @@
...
+            self.play("button_press", event.ui_element)
+        elif event.type == pygame_gui.UI_BUTTON_ON_HOVERED:
+            if event.ui_element.__class__ not in [CatButton, UISpriteButton]:
+                if self.pressed != event.ui_element:
+                    self.play("button_hover")
+            self.pressed = None
+
+    def play(self, sound, button=None):
+        """plays the given sound, if an ImageButton is passed through then the sound_id of the ImageButton will be
+        used instead"""
```
j-gynn: `:param:` for sound and button please

scripts/game_structure/audio/sound.py
```diff
@@ -0,0 +1,103 @@
...
+
+        try:
+            if pygame.mixer.find_channel():
+                chosen = random.choice(self.sound_dict[sound])
+                chosen.play()
+        except KeyError:
+            logger.exception(f"Could not find sound {sound}")
+
+    def change_volume(self, new_volume):
+        """changes the volume, int given should be between 0 and 100"""
```
j-gynn: `:param:` for new_volume

scripts/game_structure/audio/timer.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def start(self):
+        self.started_at = time.time()
+        super().start()
+
+    def elapsed(self):
+        return time.time() - self.started_at
+
+    def remaining(self):
+        return self.interval - self.elapsed()
```
j-gynn: Any reason these aren't `@property`? Does it match the underlying structure? /genq

scripts/screens/Screens.py
```diff
@@ -177,6 +176,16 @@ def on_use(self):
 
     def screen_switches(self):
         """Runs when this screen is switched to."""
+        if (
+            game.last_screen_forupdate
+            in (GameScreen.START, GameScreen.EVENTS, GameScreen.MAKE_CLAN)
+            or switch_get_value(Switch.cur_screen) == GameScreen.START
+        ):
+            game.audio.check(
```
j-gynn: Why is audio being moved to the `Game` object? /genq

j-gynn (changes requested):
- A lot of inconsistency between `fade`, `fade_out` and `fade_ms`, and possibly others. Personally I'd rather standardise around `fade_ms`, but that's personal preference. Also the use of a very similar word as a Boolean, with no indication that it's different by name.
- Please standardise to milliseconds rather than the mix of millis and seconds. This pops up in a few places so I thought I'd repeat it here in case I missed it in a couple spots.
- I don't understand the motivation behind moving the audio management to living inside `game`. I'm trying very hard to move things *out* of Game(!). Could you talk me through the rationale? I think there's a comment specifically discussing this if you want to use that

I also think it would have been beneficial for some of these to use inheritance. However, I also recognise this has been waiting for a long time and you're probably uninterested in changing how that all works at this point. Therefore, I haven't mentioned it.

scripts/game_structure/audio/ambiance.py
scribblecrumb: I think my reasoning here was just readability in the `ambiance.json`. It's far easier for me to look at the word `lakeside` and identify which of our ambient sounds would fit with that environment than it is for me to look at the word `camp3` and think of what would fit (i'd have to actually go find the camp images, which would take far more time). Though admittedly we don't really mix-and-match our ambiance atm, we have lakeside sounds for the lakeside, tunnel sounds for the tunnels, ect. So perhaps this is irrelevant.

I can see the argument for keeping it consistent with how we refer to camps across the rest of the code, though. But I'd argue that ease-of-use for people attempting to add more ambiance tracks to specific camps isn't a bad idea.  I'm willing to change it though, if you prefer the consistency.

scripts/game_structure/audio/ambiance.py
scribblecrumb: Ugh, I think my brain was fried by the time I got around to thinking abt this. I'll say I *think* it's a non-issue? I never encountered problems with it while testing and haven't received any reports about it from the beta testers.

Looking at the pygame docs, if no channel is found then it just returns None. so I suppose the safest thing here would be to add an `else` catch which starts the silence timer instead? that way it just tries again after a little bit, hopefully finding the channels cleared enough?

scripts/game_structure/audio/music.py
scribblecrumb: Yeah pycharm doesn't give a shit about it. I can always move it out of the default, but it's not as though that would change the actual behavior, so idk that I see the point.

scripts/game_structure/audio/music.py
scribblecrumb: I agree! unit is seconds, I'll specify as such

scripts/game_structure/audio/music.py
scribblecrumb: On the contrary, `duration` is only here if a coder decides it's necessary to dictate a non-default value. By default (when duration is None), we'll find the length of the loaded track and use that for the timer, which is what we want to do the vast majority of the time.

tbf, I guess I don't forsee why someone would want to run this timer longer. So perhaps it's better to just remove this arg altogether?

scripts/game_structure/audio/music.py
scribblecrumb: ugh, unfortunately python timers and pygame audio did not decide to be consistent with each other. timers use seconds, pygame audio uses milliseconds.

I could make all of our variables millisecond based, but then I'd have to manually convert them to seconds whenever I actually start a timer thread. Sounds like a pain in my ass, but I can do it if it's driving you mad.  Compromise could be for me to just rename the `duration` variables to `duration_in_seconds` or smth similar.

scripts/game_structure/audio/music.py
scribblecrumb: Why?

scripts/game_structure/audio/sound.py
scribblecrumb: Good question: I have no idea. Pretty sure this is how we've handled sounds since forever, idk who wrapped this originally or why.  I think I'll just remove the try-except and if something breaks then I'll know what they were trying to except lmao

scripts/game_structure/audio/timer.py
scribblecrumb: It does, though tbf there isn't much underlying structure to begin with. I could make them properties

scripts/screens/Screens.py
scribblecrumb: I think it just felt like a logical place to find it. `game` already holds things like the current screen, keyspressed, screen trackers, ect. Audio felt like it fit there and like, if people were trying to find it, they would head to `game` first; `game.audio` feels very intuitive to me.

I don't think that we should *avoid* adding more things to `game`, it just needs to be things that make sense.  There's shit in there that definitely *doesn't* make sense to store there, like `max_name_length`, but that doesn't mean we shouldn't store anything there at all.

scripts/game_structure/audio/ambiance.py
scribblecrumb: I'm.... not sure what you mean by that?

j-gynn: hm now i have to dive into what exactly happens in a `try-except` block... i think it's fine? if we get weird behaviour later, i reserve the right to say "I told you so" 😉

scripts/game_structure/audio/ambiance.py
j-gynn: ...honestly you're just giving a good argument as to why we should convert everything *else* to use these names... but i digress...

scripts/game_structure/audio/music.py
j-gynn: I don't think it'll do what you think it'll do though - from a brief google, it seems like the random number is chosen when the function is _created_, not when it's called. See: https://stackoverflow.com/questions/62412902/python-function-default-argument-random-value

so if you do want it to pick a random length of time each time, you should move it.

scripts/game_structure/audio/music.py
j-gynn: Principle of Least Astonishment (POLA) would make it easier for people looking in on the code to go "ok everything in sound accepts time in milliseconds". Reduces their cognitive load basically. Up to you. I don't suspect this code will get touched much, but it's nice to keep things standard.

j-gynn (approved):
2 minor nitpicks, up to you to change them or not, but otherwise LGTM.

scripts/game_structure/audio/music.py
scribblecrumb: Well damn, didn't know it did that.

Issue #3896, opened 2025-07-27
Event Frequency Feedback


scribblecrumb:
This is a masterpost for suggestions on the frequency of individual events! If you feel an event happens too often or not often enough, this is where you should let us know.

### **Please use the following form when posting:**

```
**Event:** <!--- include either the full text of the event, a screenshot of the full event, or the event/patrol ID -->
**Specific Feedback:** <!--- Tell us what you think is wrong about this event's frequency! Is it happening repetitively? Does it feel like a niche circumstance that shouldn't happen often? Have you never seen it before in hours of playing and think it should be more common? -->
```

FangWasHere:
**Event:** gen_injury_mangledleg_anymultistatusnotcareful1
**Specific Feedback:** Happening way too often, my Testclan is getting decimated, occurred 4 times over the span of 6 moons

FangWasHere:
**Event:** gen_injury_war_predatorsaved
**Specific Feedback:** Happening way too often, my TestClan is getting decimated, occurred 5 times over the span of 6 moons and on one of those moons it happened 3 times

FangWasHere:
**Event:** multi_injury_multiwaterinthelungs1
**Specific Feedback:** This was adjusted previously and yet cats still yearn for the river depths in alarming frequencies, occurred 6 times over the span of 9 moons

FangWasHere:
**Event:** gen_injury_cliffbreak1
**Specific Feedback:** Happening way too often, practically half my Testclan is incapacitated and/or actively dying from broken bones at any given time, occurred 5 times over the span of 10 moons

maplebreeze33:
**Event:** gen_med_apps_and_poison 
**Specific Feedback:** Happens too often (like nearly half the time maybe more) when you send out a med apprentice with another apprentice- probably the result of there not being many other patrols written under these conditions.

Vickarooni:
**Event:** gen_death_drown_any1
**Specific Feedback:** happening WAY too often (for me it seems to happen every 3-4 moons or so) and it's especially bad because you always lose 2 cats to it, so cats start dropping like flies. i THINK this is the event that has the history text of "this cat drowned alongside a clanmate"

Vickarooni:
**Event:** multi_injury_leafbarebrokenbone1
**Specific Feedback:** happening extremely often to the point my clan is decimated with broken bones. same issue as gen_injury_cliffbreak1, one time i got it to trigger 5 separate times in the same moon

[...]

PR #4160, opened 2025-12-30
[ENHANCEMENT] Patrol Frequency

scribblecrumb:
## About The Pull Request
I've implemented the Frequency and auto-weighting system from short events into patrols.

- `weight` has been replaced by `frequency` in the patrol format (documentation and schemas are updated to account for this)
- This does mean that every patrol file appears altered. I utilized a script to convert `weight` to `frequency`, thus all the patrol files were also reformatted (which is nice! some of the files had pretty mangled indents). However, I don't expect reviewers to read through all these patrols. ^^; Please trust that I've calibrated the new frequencies to fit close to the original intentions for the events.
- Each patrol event and patrol outcome now determines it's own `weight` according to the ways it's been constrained. 
- Added `get_frequency` and `find_new_frequency` helper funcs to `event_filters.py` so that frequency calcs could be shared by patrols and short events in a centralized space.
- Added print lines for frequency/weight of patrol and outcome to the patrol print message
 

This helps unbury highly constrained patrols from the pile and make them more likely to occur when their constraints are being met.  The frequency system is also far easier for writers to understand and gives clear direction instead of just unbridled freedom. Freedom is great, but it becomes a hindrance when working with such a large range of events and attempting to establish consistency between a large team.

## Proof of Testing
image
image
image
image
image
image
image

I've also got the beta testers playing on my branch to give feedback on the frequency of events: [thread](https://discord.com/channels/1003759225522110524/1455331612299628680)
Been playing a bit with it myself and adjusting things as I see fit, I think though that it's hit a decent balance for now!

## Changelog/Credits
- Patrols now utilize the same frequency and weight system as short events
- Documentation updated to reflect this
- Some plains patrols got misplaced into the forest files, they've been put back where they belong
- Hopefully you will see less badgers


scripts/events_module/patrol/patrol_outcome.py
```diff
@@ -43,12 +44,21 @@
...
     """Holds all info on patrol outcomes, and methods to handle that outcome"""
 
+    num_of_traits = len(Personality.trait_ranges["normal_traits"].keys()) + len(
+        Personality.trait_ranges["kit_traits"].keys()
+    )
+    num_of_skills = len(SkillPath)
+
+    num_of_ages = len(CatAge)
+
+    num_of_ranks = CatRank.get_num_of_clan_ranks()
```
j-gynn: since these are constants, shouldn't they be in UPPER_SNAKE_CASE?

Unfortunately we can't have these in `constants.py` I think because of `Personality`, since iirc that pulls from `Cat` because of course it does.

j-gynn (commented):
i hope you don't mind but i have not read a single patrol change <3 only the code. another constants question from me, but once you're happy with the balance from betas, I'm happy to approve.

scripts/events_module/patrol/patrol_outcome.py
scribblecrumb: Ah I guess I wasn't thinking about them as constants, but they definitely are in retrospect lmao I'll change that

scribblecrumb:
Not reading the patrol jsons is exactly what I expected and encourage lmao

Honestly what I've heard from the betas sounds good to me, so I'm happy to get this merged and then fine-tune things as the community provides more feedback via #3896. All in all it already sounds to be a good improvement on the prior system.

j-gynn (approved):
lgtm

PR #4170, opened 2026-01-03
[CLEANUP] Fixing Fox Patrols

tobes03:
## About The Pull Request

Cleaning up the dreaded fox patrols. The main goal of this PR is to reduce the fox patrols down to a more manageable number by cutting patrols that are excessive for one reason or another. In execution, this mostly meant:
- cutting solo patrols (on the rationale that a cat hunting alone would not pursue a fox scent on their own)
- cutting vixen patrols in leaf-fall/desert new-leaf (on the rationale that fox cubs would be adults by then).
- moving seasonal fox/vixen patrols to the respective biome's "any.json", tagged with the appropriate seasons. (in the case of the vixens, I removed a few words of season-specific language)

While doing so, I discovered a huge number of duplicates in the wetlands file and deleted as many as I could find.

There are still a huge number of fox patrols, but there is less reason to delete them other than to simply have fewer fox patrols.

I also ADDED RELATIONSHIP LOGS TO EVERY SINGLE ONE! So they are beautifully uniform across every file, and the relationship logging project does not have to deal with it.

Better for everyone. Especially writers.

## Proof of Testing

image
still possible to generate fox patrol


scribblecrumb (approved):

resources/dicts/patrols/explicit_patrol_art.json
scribblecrumb, line 1: Why were these removed?

scribblecrumb (changes requested):

resources/dicts/patrols/explicit_patrol_art.json
tobes03: I'm trying to rename them, not remove them.  Did I delete them by accident?

scribblecrumb: Oooh no you're right, I'm just reading it wrong ugh

scribblecrumb (approved):

Issue #2816, opened 2024-09-30
[BUG] Mediation screen crash - Dev


mirandarts:
**Type**: (only select one)
- Code - general bugs related to how the game runs and responds to you. 

**Describe the bug**
Whenever I try to open the mediation screen (no matter which cat I select), the game crashes.

**Grade**: (only select one)
- Dire - Still playable, but a major feature is broken

**Reproduce**
Just select whichever mediator cat in the clan and try to go to the mediation screen with the button.

**Patrol ID + Outcome Print (if applicable)**: N/A

**Commit # or Game Version Number**: Dev. Version df923de7

**Additional context**
N/A

**Screenshots**
![Screenshot 2024-09-30 150834](https://github.com/user-attachments/assets/d0701150-a80c-47a3-a095-d55f88be53db)

j-gynn:
(oops)

mirandarts:
the mediation screen can be viewed again but it only shows one page of cats, with no arrows :/ 
![Screenshot 2024-09-30 194819](https://github.com/user-attachments/assets/2cbbc8da-ce0f-4211-9570-2a9aaf2a6ff0)

PR #4172, opened 2026-01-05
[ENHANCEMENT] Event tab pagination

scribblecrumb:
## About The Pull Request
This adds pagination to the Events screen tabs. Honestly it turned out to be relatively simple to set up.

I've added a new switch for saving page positions that essentially matches the switch we already had for saving scroll positions. This ensures that the player can flip between tabs or screens and still return to a tab on the same page they left it.

The page buttons themselves are created and recreated as needed to ensure anchoring stays centered; when possible I've limited this so that we aren't recreating the elements when we don't need to.

For now I have it set up to chunk events into pages of 10 events. This number is saved as a constant in the events script, so pretty easy for us to adjust without fiddling with the code.  I think 10 is pretty reasonable though, it gets through a good number of events without making scrolling through them feel like an ordeal.

The rel tab especially has always struggled when you get to a certain number of cats. It just gets overwhelmed; it lags and the scrolling breaks. Breaking this up into pages eliminates the issue.

## Linked Issues
#2816 shows off the scrolling issues we have with the rel events. While this PR doesn't fix *scrolling* (i think this is something we just. can't really fix. it's more of a pygui limitation), it does address the core of the problem: the tab being overwhelmed by too many events.

## Proof of Testing

https://github.com/user-attachments/assets/75b67548-c134-449c-916e-111560f73681

image

Also stress tested this with a massive Clan from a community member (Vicki) that has around 800 cats (600-ish of which are alive). Rel tab loaded instantly without any problem, whereas it would reportedly crash before.

## Changelog/Credits
- Events on the Event screen now get separated into pages, with 10 events per page.

### DIFF
--- a/scripts/screens/EventsScreen.py
+++ b/scripts/screens/EventsScreen.py
@@ -382,6 +382,7 @@ class EventsScreen(Screens):
     def reset_page_buttons(self, is_page_update=False):
         """
         Resets page button and page counter states
+        :param is_page_update: Set True if page buttons do not need to be recreated.
         """
         if self.page_control and not is_page_update:
             for ele in self.page_control.values():
--- a/scripts/screens/EventsScreen.py
+++ b/scripts/screens/EventsScreen.py
@@ -1,4 +1,5 @@
 from math import ceil
+from opcode import is_pseudo
 from typing import Dict
 
 import i18n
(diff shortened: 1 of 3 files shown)


scripts/screens/EventsScreen.py
```diff
@@ -45,6 +48,11 @@ class EventsScreen(Screens):
     current_display = "all events"
     selected_display = "all events"
 
+    EVENTS_PER_PAGE = 10
```
j-gynn: move this to `game_structure\constants.py` since it's a constant

j-gynn (commented):
one minor quibble that might be up for debate and I haven't tested it in-game but looks good to me otherwise

j-gynn (approved):

PR #4174, opened 2026-01-05
[BUGFIX] Remove random_cat None assignment

### LINKED ISSUE #4171: [BUG] print log and game state that different cats were murderers
### Pre-requisites

### Description

the console says that one cat was the murderer of the victim, but the victim's death history states that a different cat (who had less dislike towards the victim) was the murderer. in addition, the history of that different cat also states that they murdered the victim, while that same different cat's id is also referred to as the murderer in the future_events.json.

### Grade

1. timeskip until a hidden murder (one that initially looks like a normal death event) occurs
2. note who the console states is the murderer
3. check the victim's death history
4. chance of it not lining up

_No response_

65c9d8e8

### Error log

_No response_

### Additional context

outcome print: 
```
Aseptic TARGET CHOSEN Boiserock
KILL KILL KILL
```

### Screenshots

screenshots from the game: 
Image
Image

screenshot from future_events.json: ("5" is Aseptic's ID, "21" is Boiserock's ID, "26" is Beechflow's ID)

Image

### LINKED ISSUE #4173: [BUG] too many failed murders between cats with empty relationships
### Pre-requisites

### Description

see title, in this example i got 3 different failed murder events, and in every single one of them the cats involved had literally 0 relationship points, positive or negative. this isnt just the random murder chance, something appears to be up

### Grade

honestly just moonskip and pay attention to the relationships between the cats involved

_No response_

65c9d8e

### Error log

_No response_

### Additional context

_No response_

### Screenshots

example one:<br>
Image<br>

Image<br>
<br>
example two:<br>

Image<br>

Image<br>
<br>
example three:<br>

Image<br>

Image
<br>
<br>
noting all of this was in the same moon

scribblecrumb:
## About The Pull Request
This was discovered via failed murder fuckery, but has likely been silently affecting many events. When we returned from the `filter_events` func (due to failing to find any events) we were incorrectly assigning `random_cat` as `None`, even when a `random_cat` did already exist.

Easy fix, we just make sure it always returns it's `random_cat` variable, which will be whatever was passed to it in the first place (either `None` or the appropriate cat object). This way we're not losing our `random_cat` to the filter.

## Linked Issues
Fixes #4173 
This was not, in fact, a problem caused by the murder code. The murders just made it more visible since they were clearly losing their initial targets.

## Proof of Testing
image

Failed murder triggered

image

While Brinicle isn't mentioned, the important cat here is Jattematt, as he's the random_cat. As you can see, he's been correctly assigned to the event, which wouldn't have happened before my fix.

image

And another instance

image

## Changelog/Credits
- random_cats no longer get detached from events when the event filter fails to find an event

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -2111,8 +2111,6 @@ def handle_murder(cat):
             )
 
         elif kill_chance <= 20:
-            print(cat.name, "TARGET CHOSEN", Cat.fetch_cat(chosen_target.cat_to).name)
-            print("FAILED")
             create_short_event(
                 event_type="misc",
                 main_cat=cat,
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -2050,8 +2050,7 @@ def handle_murder(cat):
     if random.getrandbits(murder_capable) != 1:
         return
 
-    # If random murder is not triggered, targets can only be those they have some dislike for
-    # If random murder is not triggered, targets can only be those they have extreme negativity for
+    # If random murder is not triggered, targets can only be those they have some mid/extreme neg for
     negative_relation = [
         i
         for i in relationships
@@ -2112,6 +2111,8 @@ def handle_murder(cat):
             )
 
         elif kill_chance <= 20:
+            print(cat.name, "TARGET CHOSEN", Cat.fetch_cat(chosen_target.cat_to).name)
+            print("FAILED")
             create_short_event(
                 event_type="misc",
                 main_cat=cat,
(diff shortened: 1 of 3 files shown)


scripts/events.py
```diff
@@ -2112,6 +2111,8 @@ def handle_murder(cat):
             )
 
         elif kill_chance <= 20:
+            print(cat.name, "TARGET CHOSEN", Cat.fetch_cat(chosen_target.cat_to).name)
+            print("FAILED")
```
archanyhm: Debug prints left in?

archanyhm (changes requested):
Adding a test case for this specific bug might be wise

scribblecrumb:
Added a test to check if an empty event list results in the None reassignment!

archanyhm (approved):

PR #4188, opened 2026-01-07
[CONTENT] The Scottish Patrol

tobes03:
## About The Pull Request

Adds a leader/deputy patrol inspired by Macbeth. The deputy and leader become separated, and a mysterious loner tells the deputy that their destiny is to be leader. It has eight outcomes -- two successes, two fails, two antag successes, two antag fails. One of the successes is locked to certain aggressive personalities and will have a HUGE negative impact on feelings from the deputy toward the leader in order to encourage a murder event.

who doesn't love a bit of the bard

## Proof of Testing

image

image

### DIFF
--- a/resources/lang/en/patrols/new_cat.json
+++ b/resources/lang/en/patrols/new_cat.json
@@ -7407,7 +7407,7 @@
         "deputy": [ 1, 1 ],
         "leader": [ 1, 1]
     },
-    "frequency": 4,
+    "frequency": 1,
     "chance_of_success": 50,
     "intro_text": "p_l and r_c become separated on the way to the border. Out of the mist, a voice calls to r_c by {PRONOUN/r_c/poss} name, only with the -star suffix.",
     "decline_text": "r_c doubles back to find p_l, ignoring the voice.",
--- a/resources/lang/en/patrols/new_cat.json
+++ b/resources/lang/en/patrols/new_cat.json
@@ -7432,6 +7432,7 @@
             [
                 "loner",
                 "female",
+                "age:senior",
                 "meeting",
                 "unknown",
                 "dead"
@@ -7439,6 +7440,7 @@
             [
                 "loner",
                 "female",
+                "age:senior",
                 "meeting",
                 "unknown",
                 "dead"
@@ -7446,6 +7448,7 @@
             [
                 "loner",
                 "female",
+                "age:senior",
                 "meeting",
                 "unknown",
                 "dead"
@@ -7467,14 +7470,16 @@
                 "values": ["like", "trust", "respect"],
                 "amount": -75,
                 "log": {
-                    "cats_from": "from_cat learned from a mysterious loner that {PRONOUN/from_cat/poss} destiny is to replace to_cat."
+                    "cats_from": "from_cat learned from three mysterious loners that {PRONOUN/from_cat/poss} destiny is to replace to_cat."
                 }
             }
             ],
         "new_cat": [
             [
                 "loner",
+                "female",
                 "meeting",
+                "age:senior",
                 "unknown",
                 "dead"
             ],
@@ -7482,12 +7487,14 @@
                 "loner",
                 "female",
                 "meeting",
+                "age:senior",
                 "unknown",
                 "dead"
             ],
             [
                 "loner",
                 "female",
+                "age:senior",
                 "meeting",
                 "unknown",
                 "dead"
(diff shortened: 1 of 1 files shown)


resources/lang/en/patrols/new_cat.json
```diff
@@ -7393,5 +7393,341 @@
...
+    "types": [ "border" ],
+    "tags": [ "new_cat" ],
+    "patrol_art": "gen_found_scent_dangerous",
+    "min_cats": 2,
+    "max_cats": 2,
+    "min_max_status": {
+        "deputy": [ 1, 1 ],
+        "leader": [ 1, 1]
+    },
+    "frequency": 4,
```
scribblecrumb: I think I'd like to see a lower frequency. This is a pretty... unique patrol and it'd be nice to preserve its specialness by keeping it on the rarer side.

resources/lang/en/patrols/new_cat.json
```diff
@@ -7393,5 +7393,341 @@
...
+    "chance_of_success": 50,
+    "intro_text": "p_l and r_c become separated on the way to the border. Out of the mist, a voice calls to r_c by {PRONOUN/r_c/poss} name, only with the -star ending.",
```
scribblecrumb:
```suggestion
    "intro_text": "p_l and r_c become separated on the way to the border. Out of the mist, a voice calls to r_c by {PRONOUN/r_c/poss} name, only with the -star prefix.",
```
I think this feels a little more explicit about what's happening.

resources/lang/en/patrols/new_cat.json
```diff
@@ -7393,5 +7393,341 @@
...
+            "amount": -10,
+            "log": {
+                "cats_from": "from_cat and to_cat felt uneasy with each other after being separated in the mist on patrol."
+            }
+        }
+    ],
+    "art": "cat_running_two"
+        },
+        {
+            "text": "r_c follows the voice and finds three mysterious loners, but before {PRONOUN/r_c/subject} can find out how these loners knows {PRONOUN/r_c/poss} name, p_l reappears and the loners vanish. p_l and r_c return to camp, a tension hanging between them.",
```
scribblecrumb:
```suggestion
            "text": "r_c follows the voice and finds three mysterious loners, but before {PRONOUN/r_c/subject} can find out how these loners know {PRONOUN/r_c/poss} name, p_l reappears and the loners vanish. p_l and r_c return to camp, a tension hanging between them.",
```

resources/lang/en/patrols/new_cat.json
```diff
@@ -7393,5 +7393,341 @@
...
+    "frequency": 4,
+    "dead_cats": ["n_c:0"],
+    "outsider_rep": -1,
+    "relationships": [
+        {
+            "cats_to": ["r_c"],
+            "cats_from": ["p_l"],
+            "mutual": false,
+            "values": ["trust", "comfort"],
+            "amount": -10,
```
scribblecrumb:
```suggestion
            "amount": -20,
```
I feel like this deserves a bit of a bigger reaction; I mean, it's a pretty big incident!

scribblecrumb (changes requested):

resources/lang/en/patrols/new_cat.json
tobes03: Good thinking. I wasn't sure because I feel like it being a leader-deputy patrol will create a certain rareness of its own. How does "2" sound for frequency?

resources/lang/en/patrols/new_cat.json
tobes03: LOL actually so true

resources/lang/en/patrols/new_cat.json
tobes03: Yes! Although I think -star here is a suffix

resources/lang/en/patrols/new_cat.json
scribblecrumb: You're not wrong, but we want to avoid factoring that into frequency decisions since the code already auto-weights patrols based on their constraints. So we don't need to worry about modifying frequency based off constraints ourselves.  If we do, it'll actually throw things off in comparison to other patrols!  Making this a 4 could very easily lead to *most* instances of lead+dep patrols getting this patrol specifically, which could be frustrating (esp in smaller clans, where it's more likely).  I think 2 sounds good to me.

tobes03: THEY DO? WHOOPS. Thank you 😭 in that case I might even make it a 1

scribblecrumb (approved):

PR #4204, opened 2026-01-13
[ENHANCEMENT] Death history changes

scribblecrumb:
## About The Pull Request
This PR changes how we assign and store death history. We no longer create separate leader and regular cat death history, rather all cats receive the same death text.

For leaders, instead of slapping the death text into a pre-built sentence (`"leader lost their first life when they [death_text}"`) we now just list the life numbers of each death text (`"First Life: leader died to a dog."`) Check the Proof of Testing for more examples of this in use.

Regarding save conversion, we're incrementing the save version number to allow us to track which history saves have already been adjusted.  Leader death text is what's being converted here; we're essentially "adding" the start of the sentence onto their old sentence fragments.

All short events and patrols have been adjusted to just use `"death"` rather than `"reg_death"` and `"lead_death"`. Documentation has also been updated.  

This reduces the complexity of our death history system, lowering the mental load on writers and coders.

## Proof of Testing
image
image

(death result message is not something intended to be addressed by this PR)
imageimage

image
image

image
image

image
image
image

## Changelog/Credits
- The way we process death history has changed! Leader death history is now displayed as "[number] Life: [death reason]"
- For those writing events: We no longer utilize the split `lead_death` and `reg_death` system. Check the documentation for the new system. Spoiler: it's simpler and you only have to write one string now!
- Save version incremented to 4 in order to handle death history conversions


j-gynn (commented):
One thing I'm not a huge fan of is when we're losing more than three lives. We could say "remaining lives" maybe? It could just be the capitalisation I don't like. Still looking at the code otherwise, just wanted to point this one out.

scripts/events_module/short/short_event.py
```diff
@@ -700,14 +700,14 @@ def handle_death_history(self):
                 if self.r_c["dies"]:
                     if self.random_cat.status.is_leader:
```
j-gynn: why keep this check?

scripts/events_module/short/short_event.py
```diff
@@ -666,14 +666,14 @@ def handle_death_history(self):
                     # find history
                     if self.main_cat.status.is_leader:
```
j-gynn: why keep this check?

scripts/events_module/short/short_event.py
```diff
@@ -733,14 +733,14 @@ def handle_death_history(self):
                 for cat in self.multi_cat_objects:
                     if cat.status.is_leader:
```
j-gynn: why keep this check?

scripts/events_module/short/short_event.py
```diff
@@ -849,14 +849,14 @@ def handle_injury_history(self, cat, cat_abbr, injury=None):
                     )
                     if cat.status.is_leader:
```
j-gynn: why keep this check?

scripts/screens/ProfileScreen.py
```diff
@@ -1608,64 +1608,47 @@ def get_death_text(self):
...
-                        # added code
-                        if "This cat was" in text:
-                            text = text.replace("This cat was", "{VERB/m_c/were/was}")
-                        else:
-                            text = text[0].lower() + text[1:]
+                        while multi_life_count != 0:
+                            multi_life_count -= 1
+                            temp_index += 1
+                            lives.append(
+                                i18n.t(f"utility.{life_names[temp_index]}").capitalize()
```
j-gynn: personally not a fan of the way this Capitalises Every Death Cardinal. I'd prefer it if it did "First, second, and third lives:"

j-gynn (changes requested):

scribblecrumb:
> One thing I'm not a huge fan of is when we're losing more than three lives. We could say "remaining lives" maybe? It could just be the capitalisation I don't like. Still looking at the code otherwise, just wanted to point this one out.

There's also moments where they lose more than 3 lives, but not all lives. In those cases a "remaining lives" doesn't work.

I'm a fan of keeping it simpler on the code side tbh. The more "if this many, do this thing" we add, the more convoluted that section of the profile code gets and the harder it is to parse. It took me a while to understand wtf we were doing with the old system, which had quite a few of those.

scribblecrumb:
image

updated visual look of leader life history

scripts/events_module/patrol/patrol_outcome.py
```diff
@@ -107,14 +106,7 @@ def __init__(
...
 
-        self.history_reg_death = (
-            history_reg_death if history_reg_death else "m_c died on patrol."
-        )
-        self.history_leader_death = (
-            history_leader_death
-            if history_leader_death is not None
-            else "died on patrol."
-        )
+        self.history_death = history_death if history_death else "died on patrol."
```
j-gynn: shouldn't it be `m_c died on patrol.`?

scripts/events_module/patrol/patrol_outcome.py
```diff
@@ -211,13 +203,8 @@ def generate_from_info(
...
-                        if isinstance(_d.get("history_text"), dict)
-                        else None
-                    ),
-                    history_reg_death=(
-                        _d["history_text"].get("reg_death")
+                    history_death=(
+                        _d["history_text"].get("death")
                         if isinstance(_d.get("history_text"), dict)
                         else None
                     ),
```
j-gynn:
```suggestion
                    history_death=_d.get("history_text", {}).get("death")
```
should do the same thing without making my eyes bulge out

scripts/events_module/patrol/patrol_outcome.py
```diff
@@ -1006,20 +987,14 @@ def __handle_scarring(self, cat: Cat, scar_list: str, patrol: "Patrol"):
     def __handle_condition_history(
         self, cat: Cat, condition: str, patrol: "Patrol", default_overide=False
```
j-gynn: `default_overide` is misspelled. can you correct it whilst you're snuffling around in here, since you did correct the comment?

scripts/screens/ProfileScreen.py
```diff
@@ -1685,31 +1673,18 @@ def get_death_text(self):
...
 
-            if (
-                self.the_cat.status.is_leader
-                or CatRank.LEADER in self.the_cat.status.all_ranks
-                or death_number > 1
-            ):
-                if death_number > 1:
-                    deaths = str("\n" + str(self.the_cat.name) + " ").join(all_deaths)
+            if CatRank.LEADER in self.the_cat.status.all_ranks or number_of_deaths > 1:
+                if number_of_deaths > 1:
```
j-gynn: why is the second check for `number_of_deaths` needed?

j-gynn (changes requested):

j-gynn (approved):

PR #4150, opened 2025-12-27
[BUGFIX] herb loss/gain log

scribblecrumb:
## About The Pull Request
Just needed to make it `self.text`, since it was just `self` it was using the repr value for the event object.

## Linked Issues
Fixes #4116

## Proof of Testing
image

## Changelog/Credits
- Herb loss/gain special events now appear correctly in the med den log.

PR #4207, opened 2026-01-14
[BUGFIX] Re-use of existing outsiders

### LINKED ISSUE #4205: [BUG] Broken Loner War Event
### Pre-requisites

### Description

I'm not entirely sure what's happening here, but I'd figure I'd report it all the same.

A war just started with my clan and another, and an event happened where a loner apparently brought us herbs. But, there's a few issues. The enemy clan isn't properly named in the event, the med cat's log... exists <_<;, The initial cat named turned out to be a rogue, and I got an extra loner. Very strange.

Don't worry, I triple checked my events this time. I do not know where Stone came from.

### Grade

1. Have a war start
2. Get the lonergift_herbs2 event
3. Become confused

_No response_

4a0bfc33

### Error log

_No response_

### Additional context

Screenshot 1: The initial event
Screenshot 2: The med cats log
Screenshot 3: The cat mentioned in the event. Generated as a rogue when she should have been a loner.
Screenshot 4: Extra cat

### Screenshots

Image

Image

Image

Image

scribblecrumb:
## About The Pull Request
So when an event allows the use of multiple existing outsiders, we weren't doing anything to prevent the same outsider from being chosen over and over (resulting in a single cat being used for an event that requires, for example, 3 cats). I've added a preventative to the list comprehension for finding existing cats to use that will stop us from re-using cats.

The creation of text explaining the encountering of a new cat was also a bit funky. Basically we were losing information. I've addressed this with a little re-ordering of operations.

Additionally, `gen_new_cat_war_lonergift_herbs2` is meant to only create and use loners, but included a backstory possibility that is only for rogues, which meant that, despite the `loner` attribute, some of these cats were being created as rogues.

## Linked Issues
fixes #4205 
Note that the med log problem reported in this issue is already resolved in #4150.  I also acknowledge that this event should really be providing involved cat buttons. Luckily this is also already addressed in #4134.

## Proof of Testing
image

Used the debug ensure, thus why this event is repeating.  However this is a good way to test if the event is correctly finding existing cats. Every single iteration of this event has grabbed the same loners! So we're looking good in that respect.  You can also see that every loner is a different cat and that their "welcome" text (`clan has encountered -`) is correct.

image
image
image

The loners in question. All have generated correctly as loners (lowkey really love that one of them is an adol too)

## Changelog/Credits
-  `gen_new_cat_war_lonergift_herbs2` now correctly creates loners instead of rogues
- Events that allow the use of multiple existing outsiders now correctly grab *different* outsiders instead of just repeatedly grabbing the same outsider over and over.
- Events that create or use multiple outsiders now correctly list every outsider encountered within the event text, instead of just the last encountered cat.


j-gynn:
Semi-unrelated hot take, but "The Clan has encountered X" always reads like it's a first-time meeting to me.

scribblecrumb:
> Semi-unrelated hot take, but "The Clan has encountered X" always reads like it's a first-time meeting to me.

hmmm funny cus when I changed it to that way back when it was specifically to replace the default "the clan met x" so that it *wouldn't* read as a first time meeting lmao

here's a few options:
- The Clan came upon x
- The Clan crossed paths with x
- The Clan ran into x

there aren't a lot of options here, the thesaurus isn't helping much

j-gynn:
Could be as simple as saying "The Clan re-encountered X." Could go back to using "met" for a first time encounter too.

If multiple COTC are mentioned, could we list them? So it'd be like "The Clan met X. The Clan re-encountered Q, R, and S."
Instead of repeating the whole string again and again.

scribblecrumb:
idk how easy that'll be to set up. I'd prefer if we could just keep it simple with a single string.  atm we don't go to the trouble of trying to detect if a cat has or hasn't been encountered before.

Could we perhaps make this an enhancement issue? Either way we end up going, it's rather unrelated to this bugfix in particular.

j-gynn:
Fine by me!

j-gynn (approved):

PR #4218, opened 2026-01-16
[CLEANUP] Dog Patrol Condensing

tobes03:
## About The Pull Request

Just a few patrols condensed. This isn't going to save all that much space because there are not many overlapping outcomes, but it does make it a little easier to read.

More straightforward for writers.

## Proof of Testing

image
dawg patrol still generating

### DIFF
--- a/resources/lang/en/patrols/desert/border/any.json
+++ b/resources/lang/en/patrols/desert/border/any.json
@@ -194,11 +194,11 @@
         "chance_of_success": 40,
         "success_outcomes": [
             {
-                "text": "The patrol shadows the dog, sending a message back to warn camp. It's an exhausting and stressful day, but by the time night falls, they've confirmed that the dog is leaving the local area, and the patrol has assured c_n's safety.",
-                "exp": 20,
                 "min_max_status": {
                     "normal adult": [2, 3]
                 },
+                "text": "The patrol shadows the dog, sending a message back to warn camp. It's an exhausting and stressful day, but by the time night falls, they've confirmed that the dog is leaving the local area, and the patrol has assured c_n's safety.",
+                "exp": 20,
                 "relationships": [
                     {
                         "cats_to": [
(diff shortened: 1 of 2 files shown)


scribblecrumb (changes requested):
Something I'm noticing is that allowing apprentices might bite us in the butt here.

All the outcome min_max_statuses are for normal adults, with apps are not part of. So let's say we have a patrol of just 2 apprentices. This is allowed per the patrol-wide min_max_status, but *all* the outcomes require some number of adults.

I would suggest either adding outcomes that have *no* min_max_status requirements or adding some that explicitly allow up to 2 apprentices.  Alternatively, just disallow apps entirely on these patrols.

tobes03:
> Something I'm noticing is that allowing apprentices might bite us in the butt here.
> 
> All the outcome min_max_statuses are for normal adults, with apps are not part of. So let's say we have a patrol of just 2 apprentices. This is allowed per the patrol-wide min_max_status, but _all_ the outcomes require some number of adults.
> 
> I would suggest either adding outcomes that have _no_ min_max_status requirements or adding some that explicitly allow up to 2 apprentices. Alternatively, just disallow apps entirely on these patrols.

But since the Normal Adult is minimum 2, shouldn't it be impossible for the patrol to generate with just two apps? I'm also not against just disallowing apps on this patrol lol but I thought I safeguarded against app-only shenanigans

scribblecrumb:
> > Something I'm noticing is that allowing apprentices might bite us in the butt here.
> > All the outcome min_max_statuses are for normal adults, with apps are not part of. So let's say we have a patrol of just 2 apprentices. This is allowed per the patrol-wide min_max_status, but _all_ the outcomes require some number of adults.
> > I would suggest either adding outcomes that have _no_ min_max_status requirements or adding some that explicitly allow up to 2 apprentices. Alternatively, just disallow apps entirely on these patrols.
> 
> But since the Normal Adult is minimum 2, shouldn't it be impossible for the patrol to generate with just two apps? I'm also not against just disallowing apps on this patrol lol but I thought I safeguarded against app-only shenanigans

Oh shit no you're right. god it is *so* hard to keep this stuff straight sometimes lmao

scribblecrumb (approved):

PR #4226, opened 2026-01-19
[DOCUMENTATION] Remove direct references to team members

ChibiGalaxies:
## About The Pull Request

Updated to remove direct references to current and ex-team members in the writing page to maintain timelessness of the document. 

Includes the removal of an ask to dm an ex-team member with their username or to ping them when having grammar trouble and removes team member names in examples. All replaced with generic equivalents.

Doesn't remove any credits from the pages, but a discussion might be needed by the team as to whether said credits should be kept, and if so, what format they should be in. This changes across the wiki so having a consensus on this would be ideal. 

About documentation, rather than the game itself. Keeps documentation timeless. Ideally, documentation shouldn't need to be updated if team members depart for whatever reason. Therefore direct references to them should possibly be avoided. 

Mentioned noticing this in the moderator chat of the discord and said I'd fix it.


docs/dev/writing/index.md
```diff
@@ -210,7 +210,7 @@ Event and flavor text should always use Americanized spelling.
 - HTML tags for bolding and italicizing are usable. These tags are:
     - `<i>` and `</i>` for italicizing
     - `<b>` and `</b>` for bolding
-- Keeping a change log: as a writer, people are going to be super excited to experience what you write! Make sure to keep a rough record of what that might be, i.e Tiri has no idea how many fox patrols she has added total, but does know how many species are added are in what biomes, so that when she made the PR to merge them into the game she could mention their existence in the changelog for that PR.
+- Keeping a change log: as a writer, people are going to be super excited to experience what you write! Make sure to keep a rough record of what that might be, i.e A writer has no idea how many fox patrols they have added total, but does know how many species are added are in what biomes, so that when they made the PR to merge them into the game they could mention their existence in the changelog for that PR.
```
j-gynn: This is an anecdote from a specific writer's experience, so perhaps it should be noted that "a given writer" so that it's clear it's not just a generic mass noun "writer"

docs/dev/writing/index.md
```diff
@@ -199,7 +199,7 @@ Some animals have multiple names that can refer to them, we want to keep our wri
...
         - Basically, anytime there's a possessive, you want an apostrophe in there somewhere. If the word ends in S, plural or singular, the apostrophe goes at the end. If the singular form of the word ends in S and it's a plural possessive, add "es" at the end, then apostrophe.  The exception to this is “it’s’ and “its”, where “its” is the possessive form and has no apostrophe.
 - All ClanGen game text should follow normal grammar rules for capitalizing the first letter of a sentence, and trying to avoid spelling or grammar typos. We all make typos, don't worry! But this is one of the ways beta testing your new content can help you!
-- If you are struggling with remembering or understanding a grammar rule, even one that isn't mentioned in the above list, feel free to contact zabe#1117 on Discord for one on one help. Either DM or ping in the writing contributor thread, please!
+- If you are struggling with remembering or understanding a grammar rule, even one that isn't mentioned in the above list, feel free to message in the writing contributor thread or in other offical channels of the discord server.
```
j-gynn:
```suggestion
- If you are struggling with remembering or understanding a grammar rule, even one that isn't mentioned in the above list, feel free to message in the writing contributor thread or in other official channels of the Discord server.
```

docs/dev/writing/index.md
j-gynn:
```suggestion
- Keeping a change log: as a writer, people are going to be super excited to experience what you write! Make sure to keep a rough record of what that might be. For example, a writer may have no idea how many fox patrols they are adding total. However, they do know how many species they have added and in what biomes. So, when they make the PR to merge them into the game, they can mention their existence in the changelog for that PR.
```

j-gynn (changes requested):

docs/dev/writing/index.md
ChibiGalaxies: That is true. I was editing that as if it was strictly an example of the point made, assuming that was the original reason it was used, rather than it being being anecdotal. If it would read better being specified more directly as an anecdote, that change can be made, but I'm not certain it's neccessary?

j-gynn (approved):

Issue #4220, opened 2026-01-17
[BUG] cat thinking about their rank as a separate cat


cawcaw420908:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

the only medicine cat in my clan refers to the medicine cat as a separate cat from himself. 
i also had this happen with a leader once many many moons after she became leader so that instant was caused by a different bug than the one in #3978.

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. have a medicine cat with a connection to StarClan
2. moonskip
3. check thoughts
4. chance of error

### Commit # or Game Version Number

52444d36

### Screenshots

<img width="745" height="302" alt="Image" src="https://github.com/user-attachments/assets/5d168abf-05ba-4c60-8e59-767bf648cd08" />

<img width="680" height="308" alt="Image" src="https://github.com/user-attachments/assets/98a43b49-0ed4-435c-9b32-f469af666ba3" />

(image also posted in the comments of #3978 but bringing it here because it wasn't caused by the thought rolled before promotion thing as it happened many many moons after she became leader)

j-gynn:
Closed by #4231

PR #4231, opened 2026-01-21
[BUGFIX] cat thinking about their rank as a separate cat 

tobes03:
## About The Pull Request

Adding status constraints to a lot of thoughts to ensure that cats cannot have thoughts about their own ranks.

## Linked Issues

#4220 

## Proof of Testing

I'm suffering trying to generate this thought but trust me bro. I'll keep working on generating it


resources/lang/en/thoughts/alive/general.json
```diff
@@ -1241,7 +1260,8 @@
             "warrior",
             "mediator",
             "deputy",
-            "leader"
+            "leader",
+            "elder"
```
scribblecrumb: med apps could be here too

resources/lang/en/thoughts/alive/general.json
```diff
@@ -2339,6 +2369,15 @@
...
         ],
+        "main_status_constraint": [
+            "apprentice",
+            "warrior",
+            "deputy",
+            "leader",
+            "elder",
+            "mediator apprentice",
+            "mediator"
+        ],
```
scribblecrumb, line 2383: I think med apps could be allowed here too, it'd be kinda cute lmao

scribblecrumb (changes requested):

resources/lang/en/thoughts/alive/general.json
scribblecrumb: You added the med apps to a different block than this one lmao

tobes03: 😭 damn... but it made sense there too. Anyway. I have definitely found the correct one and added it this time

scribblecrumb (approved):

PR #4234, opened 2026-01-22
[REL LOGS] General Death Events

tobes03:
## About The Pull Request

Mostly murders and lives being saved. The murders previously had a lot of logs where the murderer would have a sharp negative reaction to their victim, which doesn't really make logical sense. Those interactions were reversed to account for the rare instances where m_c is a leader who only loses one life. (The rest of the time, the victim's emotions don't matter because they're dead)

## Proof of Testing

Pending


scribblecrumb:
I'm all for keeping reactions to account for leaders, but could you double check that the events you did that for are available for leaders? A lot of murder events bar leaders from being the victim.

tobes03:
> I'm all for keeping reactions to account for leaders, but could you double check that the events you did that for are available for leaders? A lot of murder events bar leaders from being the victim.

Whoops, you're totally right. I could've saved myself a bit of work if I'd noticed that sooner lol. I deleted the useless blocks

scribblecrumb (approved):

PR #4245, opened 2026-01-23
Kill utility.py

j-gynn:
## About The Pull Request

GOOD NEWS! She's dead! Utility Py is dead! (Bam.)

I've been wanting to kill this thing for a long time, and this evening I just decided to take it into the back alley.

## Proof of Testing

Just an assortment of things to prove things are still thinging.
image
image
image


scripts/clan_package/get_clan_cats.py
```diff
@@ -129,7 +129,7 @@ def get_cats_same_age(Cat, cat, age_range=10):
     return cats
 
 
-def get_free_possible_mates(cat):
+def get_free_possible_mates(cat, existing_romance=False):
```
scribblecrumb: could we rename to `is_existing_romance` to keep better bool naming conventions? arch has been on me for this sort of thing lol so might as well keep it up

scripts/events_module/consequences.py
```diff
@@ -0,0 +1,976 @@
...
+        # create relationships
+        new_cat.create_relationships_new_cat()
+        # Note - we always update inheritance after the cats are generated, to
+        # allow us to add parents.
+        # new_cat.create_inheritance_new_cat()
+
+    return created_cats
+
+
+def gather_cat_objects(
```
scribblecrumb, line 714: not sure i agree with this func being here, but i acknowledge that it's being used by the funcs in this script.  idk, just doesn't feel like it fits under the `consequence` category? food for thought, i don't have a solution in mind here

scribblecrumb (changes requested):

scripts/events_module/consequences.py
j-gynn: I'm not sure where else it would live honestly, but I do see your point! By "consequences" I really mean "stuff that happens at the end of a patrol/event", which is what this is used for. If you can think of a better name for the file that isn't already in use, lmk!

scripts/clan_package/get_clan_cats.py
j-gynn: Sooo instead of doing that, I took another look. This func is only used in one place, where we run it twice in a row. Once with the argument, once without. So I just modified the function to return two lists, murked the argument and made a slight modification to the place that calls it. It's more efficient and honestly more clear imo.

scribblecrumb (approved):
Resolve the conflicts and you're good to go

archanyhm:
Haven’t finished my review yet, please hold

scripts/cat/sprites/load_sprites.py
Koriiy, line 1: requesting that maybe this is moved to its own file instead of being added to sprites? because it's not really involved with sprite loading, just sprite drawing. 
it's also a frequently modified part of the now dead utility, and a new file may be useful in the case of moving platform drawing to a location that makes more sense too. 

aware that sprites will have to be imported into it in that case, but i think that should be okay (also hope this is the correct way to do this since i'm not really reviewing!)

scribblecrumb (approved):

PR #4276, opened 2026-01-29
[ENHANCEMENT] Exclusionary values

scribblecrumb:
## About The Pull Request
This allows quite a few event parameters to utilize exclusionary values. When a value has `-` at the beginning, it will be considered exclusionary. For example, an event with `"trait": ["-calm"]` will disallow any cat with the calm trait.

This has been implemented into location, season, trait, skill, status, age, and backstory event filters. I have plans to eventually extend the functionality to relationship filtering, but that's going to be more complex (and utility.py is waiting to die in a PR rn) so I'm leaving it alone for now. I don't think it's as immediately applicable either.

I've done some work implementing exclusionary values into some of our jsons, but I haven't gone through every single one. I was mostly just looking to demonstrate it's use and test it on a larger scale, it can be something we slowly spread through the various jsons as we work on other things.  Some of my changes here won't seem equivalent, but I was changing things within context. If I felt like the prior constraints were unreasonable, then I gave them new constraints that better fit.

Documentation has been updated to indicate which parameters can utilize exclusionary values.

I also ended up integrating backstory category detection into the event filters. This wasn't there previously, even though our documentation implied it was and the new_cat attributes allowed their use. 

It helps smooth the writing process and makes it far far easier to read events at a glance. It's quicker to parse an event with `"-kitten"` in it's statuses than it is to parse one listing every single status except `kitten`. It's also better future-proofing. If we add a new rank, for example, we won't have to do as much work integrating it into constrained events.

Integrating backstory categories allows for better future scale-ability. 

## Proof of Testing
I'll step through these in debug to demonstrate how they work. Not gonna run through every single parameter bc that'll be A Lot of screenshots, but all of these function with extremely similar code and share a detection func, so generally if one works then all work.

Testing statuses
image
image

Cat is a leader

image

Passed them through status check and onto trait check, as the cat is not a kitten.

image

This event doesn't allow kittens *or* leaders.

image

Returns false and event is rejected, as the cat is a leader.

Testing backstories
image

Backstory category is exclusionary

image

Cat has a qualifying backstory

image

Returns false and event is rejected.

## Changelog/Credits
- Exclusionary values are now allowed in some event parameters. Documentation has been updated accordingly and labels eligible parameters. This also kills the old `not_skill` and `not_trait` parameters utilized by short events.
- Some events have been converted to use exclusionary values, but not all. This can be an ongoing background project for us.
- Backstory parameters can now utilize backstory categories as constraints, instead of just individual backstories.


scripts/models/common/skill.py
```diff
@@ -4,4 +4,4 @@
 
 
 class Skill(RootModel):
-    root: Annotated[str, StringConstraints(pattern=r"^[A-Z]+,\s*[0-4]$")]
+    root: Annotated[str, StringConstraints(pattern=r"^[A-Z]+|-[A-Z]+,\s*[0-4]$")]
```
archanyhm: I think what you really mean is `^-?[A-Z]+` instead of `^[A-Z]+|-[A-Z]+`

scribblecrumb: Is that how it works? I was getting so confusing lmao

archanyhm: > Is that how it works? I was getting so confusing lmao

Assuming that you intended to say "either starting with [A-Z] and repeat the pattern one or more times OR starting with - and then [A-Z] and repeat the pattern one or more times", yeah. There's a _tiny_ "bug" in there `^[A-Z]+|-[A-Z]+` where the second part `-[A-Z]` lacks the `^` marker, so it could either be `^-?[A-Z]+` or `^([A-Z]+|-[A-Z]+)`. RegEx is confusing as hell.

What always helps me is using https://regex101.com/r/uYaHD0/1 since it explains the rules and highlights the individual capture groups.

scripts/models/thought/thought_schema_item.py
```diff
@@ -6,6 +6,7 @@
 from pydantic_core import MISSING
 
 from scripts.cat.enums import CatAge
```
j-gynn: is there a reason this is still here? /genq

scripts/events_module/event_filters.py
```diff
@@ -47,14 +48,21 @@ def find_new_frequency(used_frequencies: set) -> int:
...
     """
-    checks if the clan is within the given locations
+    Checks if the clan is within the allowed locations.
     """
     if "any" in locations:
         return True
     if not game.clan:
         return False
 
+    exclusionary = __check_for_exclusionary_value(locations)
```
j-gynn: Does this need to be a mangled function? can it be private instead (just one underscore)

scripts/events_module/event_filters.py
```diff
@@ -66,26 +74,46 @@ def event_for_location(locations: list) -> bool:
         if game.clan.override_biome:
             if req_biome == game.clan.override_biome:
                 if "any" in req_camps or game.clan.camp_bg in req_camps:
-                    return True
+                    return False if exclusionary else True
```
j-gynn:
```suggestion
                    return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -66,26 +74,46 @@ def event_for_location(locations: list) -> bool:
...
+
         elif req_biome == game.clan.biome.lower():
             if "any" in req_camps or game.clan.camp_bg in req_camps:
-                return True
-    return False
+                return False if exclusionary else True
```
j-gynn:
```suggestion
                return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -66,26 +74,46 @@ def event_for_location(locations: list) -> bool:
...
+        # "any" will never be exclusionary, so we check for it first
+        return True
+
+    exclusionary = __check_for_exclusionary_value(seasons)
+    # check if these are exclusionary values
+    if exclusionary:
+        seasons = [x.replace("-", "") for x in seasons]
 
+    if game.clan.current_season.lower() in seasons:
+        return False if exclusionary else True
```
j-gynn:
```suggestion
        return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -66,26 +74,46 @@ def event_for_location(locations: list) -> bool:
...
 
-def event_for_tags(tags: list, cat, other_cat=None, mentor_tags_fulfilled=None) -> bool:
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -331,6 +357,7 @@ def event_for_cat(
             return False
 
     # checking relationships
+    # TODO: relationship exclusions. wait on this until utility PR is in
```
j-gynn: Utility PR is in - make those changes please!

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
 
-    return cat.age.value in ages
+    # check for exclusionary values
+    exclusionary = __check_for_exclusionary_value(ages)
+
+    if exclusionary:
+        ages = [x.replace("-", "") for x in ages]
+
+    if cat.age.value in ages:
+        return False if exclusionary else True
```
j-gynn:
```suggestion
        return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
+
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
+
+    if exclusionary:
+        statuses = [x.replace("-", "") for x in statuses]
+
+    if cat.status.rank in statuses:
+        return False if exclusionary else True
+
     if "lost" in statuses and cat.status.is_lost():
-        return True
+        return False if exclusionary else True
```
j-gynn:
```suggestion
        return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
 
-    return False
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
+        traits = [x.replace("-", "") for x in traits]
 
-def _check_cat_not_trait(cat, traits: list) -> bool:
-    """
-    checks if cat has the excluded traits
-    """
-    if not traits:
-        return True
+    if cat.personality.trait in traits:
+        return False if exclusionary else True
```
j-gynn:
```suggestion
        return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -371,52 +398,70 @@ def _check_cat_age(cat, ages: list) -> bool:
...
 
-    return not cat.personality.trait in traits
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -425,45 +470,44 @@ def _check_cat_skills(cat, skills: list) -> bool:
             continue
 
         if cat.skills.meets_skill_requirement(skill_info[0], int(skill_info[1])):
-            return True
+            return False if exclusionary else True
```
j-gynn:
```suggestion
            return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -425,45 +470,44 @@ def _check_cat_skills(cat, skills: list) -> bool:
...
 
-    return False
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -425,45 +470,44 @@ def _check_cat_skills(cat, skills: list) -> bool:
...
+    for story in backstories:
+        if story in BACKSTORIES["backstory_categories"].keys():
+            allowed_stories.extend(BACKSTORIES["backstory_categories"][story])
+        else:
+            allowed_stories.append(story)
 
-    if cat.backstory in backstories:
-        return True
+    if cat.backstory in allowed_stories:
+        return False if exclusionary else True
```
j-gynn:
```suggestion
        return not exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -425,45 +470,44 @@ def _check_cat_skills(cat, skills: list) -> bool:
...
 
-    return False
+    return True if exclusionary else False
```
j-gynn:
```suggestion
    return exclusionary
```

scripts/events_module/event_filters.py
```diff
@@ -530,6 +572,7 @@ def cat_for_event(
             return None
 
     # rel status check
+    # TODO: set up exclusionary values
```
j-gynn: Please (to)do

scripts/events_module/event_filters.py
```diff
@@ -635,64 +693,70 @@ def _get_cats_with_skill(cat_list: list, skills: tuple) -> list:
...
-    checks cat_list against required backstories and returns qualifying cats
+    Checks the given list for an exclusionary value and returns True or False
     """
-    if not backstories:
-        return cat_list
+    for value in possible_values:
+        if value.find("-") == 0:
+            return True
+        else:
+            return False
```
j-gynn: this means you only check the first value. intended?

if so:

```suggestion
    for value in possible_values:
        return value.find("-") == 0
```

if not, i think this will work?

```suggestion
    return any(value.find("-") == 0 for value in possible_values)
```

j-gynn (changes requested):

j-gynn (changes requested):
I also think this might benefit from `exclusionary` being renamed to `is_exclusionary` in all instances

scripts/events_module/event_filters.py
scribblecrumb: Yeah, it's intentional in that we assume all values will either be exclusionary or not.  So we can rely on the first value to indicate what *all* the values are.

j-gynn (approved):

PR #4289, opened 2026-01-30
Transgender Kittens

MacNCheeseNMac:
## About The Pull Request

Kittens three moons or older can now transition/come out. They will use the adolescent modifier.

Many real transgender people transition pre-puberty, so there's no reason that transition should be limited to adolescents/adults in ClanGen.  

## Proof of Testing

image

## Changelog/Credits

- Kittens 3 moons or older can now come out as transgender


MacNCheeseNMac:
Did not mean to close this 🤦 

will remake in a bit (busy)

PR #4297, opened 2026-02-01
Refactor: rename clan attributes name->uuid and displayname->name

### LINKED ISSUE #4293: [ENHANCE] Rename `name` and `displayname` clan attributes
### Description

These `clan` attributes are unnecessarily confusing.  `name` should be renamed to `UUID` and `displayname` to `name`. This would clear up *alot* of the bugs we see due to usage confusion between these two.

Camden-Kirkpatrick:
## About The Pull Request

I changed the clan attribute "name" to "uuid", and the clan attribute "displayname" to "name".

## Linked Issues

Fixes: 4293
## Proof of Testing

image

## Changelog/Credits

Changed the clan attribute "name" to "uuid", and the clan attribute "displayname" to "name".

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2222,7 +2222,7 @@ class Cat:
         elif len(switch_get_value(Switch.clan_list)) > 0:
             clanname = switch_get_value(Switch.clan_list)[0]
         elif game.clan is not None:
-            clanname = game.clan.uuid
+            clanname = game.clan.save_id
 
         condition_directory = get_save_dir() + "/" + clanname + "/conditions"
         condition_file_path = condition_directory + "/" + self.ID + "_conditions.json"
@@ -2988,7 +2988,7 @@ class Cat:
             clan = (
                 switch_get_value(Switch.clan_list)[0]
                 if game.clan is None
-                else game.clan.uuid
+                else game.clan.save_id
             )
 
             with open(
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2222,7 +2222,7 @@ class Cat:
         elif len(switch_get_value(Switch.clan_list)) > 0:
             clanname = switch_get_value(Switch.clan_list)[0]
         elif game.clan is not None:
-            clanname = game.clan.name
+            clanname = game.clan.uuid
 
         condition_directory = get_save_dir() + "/" + clanname + "/conditions"
         condition_file_path = condition_directory + "/" + self.ID + "_conditions.json"
@@ -2988,7 +2988,7 @@ class Cat:
             clan = (
                 switch_get_value(Switch.clan_list)[0]
                 if game.clan is None
-                else game.clan.name
+                else game.clan.uuid
             )
 
             with open(
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -61,7 +61,7 @@ class Clan:
 
     def __init__(
         self,
-        uuid="",
+        save_id="",
         leader=None,
         deputy=None,
         medicine_cat=None,
@@ -74,18 +74,15 @@ class Clan:
         self_run_init_functions=True,
         name="",
     ):
-        if uuid == "":
+        if save_id == "":
             return
 
         if starting_members is None:
             starting_members = []
 
-        # name is the unique id of the clan. i'm sorry if this is confusing...
-        # TODO: change to better name like clan_id
-        self.name = uuid
-        # displayname is the name you should use whenever displaying the clan name in UI
+        self.save_id = save_id
         if not name:
-            self.name = uuid
+            self.name = save_id
         else:
             self.name = name
         self.leader = leader
@@ -574,7 +571,7 @@ class Clan:
             elif general[8] == "None":
                 general[8] = 50
             game.clan = Clan(
-                uuid=general[0],
+                save_id=general[0],
                 leader=Cat.all_cats[leader_info[0]],
                 deputy=Cat.all_cats.get(deputy_info[0], None),
                 medicine_cat=Cat.all_cats.get(med_cat_info[0], None),
@@ -593,7 +590,7 @@ class Clan:
             elif general[7] == "None":
                 general[7] = "classic"
             game.clan = Clan(
-                uuid=general[0],
+                save_id=general[0],
                 leader=Cat.all_cats[leader_info[0]],
                 deputy=Cat.all_cats.get(deputy_info[0], None),
                 medicine_cat=Cat.all_cats.get(med_cat_info[0], None),
@@ -609,7 +606,7 @@ class Clan:
             elif general[3] == "None":
                 general[3] = "camp1"
             game.clan = Clan(
-                uuid=general[0],
+                save_id=general[0],
                 leader=Cat.all_cats[leader_info[0]],
                 deputy=Cat.all_cats.get(deputy_info[0], None),
                 medicine_cat=Cat.all_cats.get(med_cat_info[0], None),
@@ -620,7 +617,7 @@ class Clan:
             game.clan.post_initialization_functions()
         elif len(general) == 3:
             game.clan = Clan(
-                uuid=general[0],
+                save_id=general[0],
                 leader=Cat.all_cats[leader_info[0]],
                 deputy=Cat.all_cats.get(deputy_info[0], None),
                 medicine_cat=Cat.all_cats.get(med_cat_info[0], None),
@@ -736,7 +733,7 @@ class Clan:
             displayname = clan_data["clanname"]
 
         game.clan = Clan(
-            uuid=clan_data["clanname"],
+            save_id=clan_data["clanname"],
             name=displayname,
             leader=leader,
             deputy=deputy,
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -74,6 +74,10 @@ class Clan:
         self_run_init_functions=True,
         name="",
     ):
+        """
+        :param save_id: The save file name for the Clan, this should not be used for player-facing text beyond the save file screen
+        :param name: The display name for the Clan, this is what should appear while the playing the game.
+        """
         if save_id == "":
             return
 
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -61,7 +61,7 @@ class Clan:
 
     def __init__(
         self,
-        name="",
+        uuid="",
         leader=None,
         deputy=None,
         medicine_cat=None,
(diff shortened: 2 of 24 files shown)


scripts/clan.py
```diff
@@ -72,22 +72,22 @@ def __init__(
...
-        self.name = name
+        self.name = uuid
         # displayname is the name you should use whenever displaying the clan name in UI
-        if not displayname:
-            self.displayname = name
+        if not name:
+            self.name = uuid
         else:
-            self.displayname = displayname
+            self.name = name
```
scribblecrumb: Go ahead and remove the comments in this section, as they're no longer applicable

scripts/clan.py
```diff
@@ -72,22 +72,22 @@ def __init__(
...
+        if uuid == "":
             return
 
         if starting_members is None:
             starting_members = []
 
         # name is the unique id of the clan. i'm sorry if this is confusing...
         # TODO: change to better name like clan_id
-        self.name = name
+        self.name = uuid
```
scribblecrumb:
```suggestion
        self.uuid = uuid
```

scripts/screens/MakeClanScreen.py
```diff
@@ -2144,8 +2144,8 @@ def save_clan(self):
             clan_name = self.clan_name
 
         game.clan = Clan(
-            name=clan_name,
-            displayname=displayname,
+            uuid=clan_name,
+            name=displayname,
```
scribblecrumb, line 2148: For the sake of clarity here, can we fix these local attributes as well? This local `displayname`, for one, isn't really necessary. We could just be using `self.clan_name` in it's place.  `self.clan_name` can stay, but the local `clan_name` could be renamed to `UUID` so that it's clear that it's different from `self.clan_name`

scribblecrumb (changes requested):

scribblecrumb:
Also, just a general github tip! If you reference the issue number in your description with the hashtag in front of it (like this: #4293) then it'll autolink it! And with the keywork "fixes" that you've already used, it'll auto-close the issue once this PR is approved and merged.

archanyhm:
Just a quick interjection --- @scribblecrumb is there any reason to actually have the field be called `uuid` instead of `id`? Whether it's a `UUID` or any other format is pretty much irrelevant in most cases, as it's just an implementation detail. No?

scribblecrumb:
> Just a quick interjection --- @scribblecrumb is there any reason to actually have the field be called `uuid` instead of `id`? Whether it's a `UUID` or any other format is pretty much irrelevant in most cases, as it's just an implementation detail. No?

Well, keep in mind that we also have the group IDs that differentiate between different groups within a save file (player clan vs. other clan). Even though we don't actually save a player clan group ID and instead *assume* it will always be group 1, I think it would still be good to differentiate the "clan save ID" from the "group ID" in some respect. Calling it `UUID` helps with that; and I don't see us moving to a different ID format in the future?

j-gynn:
Also, `id` is a built-in function in Python, so it's best not to overlap with that.

archanyhm:
@scribblecrumb 
> > Just a quick interjection --- @scribblecrumb is there any reason to actually have the field be called `uuid` instead of `id`? Whether it's a `UUID` or any other format is pretty much irrelevant in most cases, as it's just an implementation detail. No?
> 
> Well, keep in mind that we also have the group IDs that differentiate between different groups within a save file (player clan vs. other clan). Even though we don't actually save a player clan group ID and instead _assume_ it will always be group 1, I think it would still be good to differentiate the "clan save ID" from the "group ID" in some respect. Calling it `UUID` helps with that; and I don't see us moving to a different ID format in the future?

That kind of highlights the issue with the naming convention; "UUID" doesn't imply anything about the purpose or role of this specific field, just that it is an UUID. Which is an implementation detail that shouldn't leak through the abstraction. Calling it `clan_id` or anything more descriptive is a better move imo.


@j-gynn 
> Also, `id` is a built-in function in Python, so it's best not to overlap with that.

This doesn't apply to property names, does it?

scribblecrumb:
> @scribblecrumb
> 
> > > Just a quick interjection --- @scribblecrumb is there any reason to actually have the field be called `uuid` instead of `id`? Whether it's a `UUID` or any other format is pretty much irrelevant in most cases, as it's just an implementation detail. No?
> > 
> > 
> > Well, keep in mind that we also have the group IDs that differentiate between different groups within a save file (player clan vs. other clan). Even though we don't actually save a player clan group ID and instead _assume_ it will always be group 1, I think it would still be good to differentiate the "clan save ID" from the "group ID" in some respect. Calling it `UUID` helps with that; and I don't see us moving to a different ID format in the future?
> 
> That kind of highlights the issue with the naming convention; "UUID" doesn't imply anything about the purpose or role of this specific field, just that it is an UUID. Which is an implementation detail that shouldn't leak through the abstraction. Calling it `clan_id` or anything more descriptive is a better move imo.
> 

Fair enough, what about `save_id`? I think that's most indicative of it's usage and differentiates enough from the intra-save group ids.

j-gynn:
> > Also, `id` is a built-in function in Python, so it's best not to overlap with that.
>
> @archanyhm 
> This doesn't apply to property names, does it?

No, but with our habit of using the local variable with the exact same name as the property, it'd still be overriding it at some point.

Camden-Kirkpatrick:
> Also, just a general github tip! If you reference the issue number in your description with the hashtag in front of it (like this: #4293) then it'll autolink it! And with the keywork "fixes" that you've already used, it'll auto-close the issue once this PR is approved and merged.

Yeah, I meant to put a "#", but I mistyped.

Camden-Kirkpatrick:
@scribblecrumb Should I change the name for "uuid" to something else, or just leave it? After that, I can apply the other changes that you marked.

scribblecrumb:
> @scribblecrumb Should I change the name for "uuid" to something else, or just leave it? After that, I can apply the other changes that you marked.

Go ahead and change it to `save_id`!

Camden-Kirkpatrick:
@scribblecrumb Alright, it should be good now.

scripts/cat/cats.py
```diff
@@ -2222,7 +2222,7 @@ def save_condition(self):
         elif len(switch_get_value(Switch.clan_list)) > 0:
             clanname = switch_get_value(Switch.clan_list)[0]
         elif game.clan is not None:
-            clanname = game.clan.name
+            clanname = game.clan.uuid
```
scribblecrumb: rename

scripts/cat/cats.py
```diff
@@ -2988,7 +2988,7 @@ def load_faded_cat(cat: str):
             clan = (
                 switch_get_value(Switch.clan_list)[0]
                 if game.clan is None
-                else game.clan.name
+                else game.clan.uuid
```
scribblecrumb: rename

scripts/clan.py
```diff
@@ -72,26 +72,19 @@ def __init__(
         starting_members=None,
         starting_season="Newleaf",
         self_run_init_functions=True,
-        displayname="",
+        name="",
     ):
-        """
-        :param name: The save file name for the Clan, this should not be used for player-facing text beyond the save file screen
-        :param displayname: The display name for the Clan, this is what should appear while the playing the game.
-        """
```
scribblecrumb, line 80: Instead of removing this param documentation, can we update it to the new names?

scripts/screens/MakeClanScreen.py
scribblecrumb: > For the sake of clarity here, can we fix these local attributes as well? This local displayname, for one, isn't really necessary. We could just be using self.clan_name in it's place. self.clan_name can stay, but the local clan_name could be renamed to UUID so that it's clear that it's different from self.clan_name

Still need to address this! But of course, rename `clan_name` to `save_id` instead

scribblecrumb (changes requested):

Camden-Kirkpatrick:
Why are you still bringing up "uuid"? I thought you told me to change it to "save_id"?

Camden-Kirkpatrick:
I'm going to let someone else finish this.

scribblecrumb:
> Why are you still bringing up "uuid"? I thought you told me to change it to "save_id"?

I was quoting a change I requested in my first review that you hadn't addressed! Yes, I still want it to be `save_id`. The point was just that I wanted those local attributes to match the `clan` attributes they were being assigned to, for the sake of readability and clarity.

Camden-Kirkpatrick:
Is this correct for MakeClanScreen.py? I'm still confused:
 ```py
def save_clan(self):
        game.mediated.clear()
        game.patrolled.clear()
        save_load.faded_ids.clear()
        Cat.outside_cats.clear()
        Patrol.used_patrols.clear()
        convert_camp = {1: "camp1", 2: "camp2", 3: "camp3", 4: "camp4"}
        displayname = self.clan_name
        if self._clan_name_exists(self.clan_name):
            clan_name = self._generate_unique_clan_name(self.clan_name)
        else:
            clan_name = self.clan_name

        game.clan = Clan(
            save_id=clan_name,
            name=displayname,
            leader=self.leader,
            deputy=self.deputy,
            medicine_cat=self.med_cat,
            biome=self.biome_selected,
            camp_bg=convert_camp[self.selected_camp_tab],
            symbol=self.symbol_selected,
            game_mode=self.game_mode,
            starting_members=self.members,
            starting_season=self.selected_season,
        )
        game.clan.create_clan()
        game.cur_events_list.clear()
        game.herb_events_list.clear()
        game.clan.herb_supply.start_storage(len(self.members))
        game.clan.save_herb_supply(game.clan)
        Cat.grief_strings.clear()
        Cat.sort_cats()
```

**Comment by Camden-Kirkpatrick** [comment]
@scribblecrumb Let me know if the above is correct or not.

**Comment by Camden-Kirkpatrick** [comment]
@scribblecrumb Lmk

**Comment by archanyhm** [comment]
@Camden-Kirkpatrick Please avoid excessive bumping of PRs and tagging of maintainers. Scribble will get to it when there's time for it.

File: scripts/screens/MakeClanScreen.py [hunk 7]
```diff
@@ -2144,8 +2144,8 @@ def save_clan(self):
             clan_name = self.clan_name
```
scribblecrumb, line 2144: What I'm trying to suggest is that you rename these local attributes: `displayname` and `clan_name` to match the new naming we see in `Clan`.  So `displayname` becomes `name` and `clan_name` becomes `save_id`.  This helps with readability for whoever has to edit this file after you.

scribblecrumb:
closing as stale

PR #4317, opened 2026-02-02
[ENHANCEMENT] Dens and supplies dropdown hover

scribblecrumb:
## About The Pull Request
This adds an option to our dropdown ui_element that allows the dropdown to open and close based on hover status. 

This is a *tad* finicky, but I think it's implemented decently. I had to create a func to manually check the hover status, as some quirks of py-gui's built in hover attribute meant that the element as a whole was always considered unhovered. 

This was a requested QoL feature based on the new UI! It helps speed up navigation and removes some tiring clicks.

## Proof of Testing

https://github.com/user-attachments/assets/ccae1ffb-3688-40f0-90f4-94ecc2c998a0

## Changelog/Credits
- Dens and Supplies buttons now open and close their dropdowns based on hover status

### DIFF
--- a/scripts/game_structure/ui_elements.py
+++ b/scripts/game_structure/ui_elements.py
@@ -2481,8 +2481,6 @@ class UIDropDown(UIDropDownContainer):
         if self.parent_reflect_selection and new_list:
             self.parent_button.set_text(new_list[0])
 
-
-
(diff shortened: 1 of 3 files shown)


j-gynn (approved):
i almost don't want to know how it works because it's so magic.

PR #4318, opened 2026-02-02
[BUGFIX] Rel log pronouns

scribblecrumb:
## About The Pull Request
This fixes the pesky from_cat and to_cat pronoun defaulting. What was happening was that the first round of text adjust was getting confused at the presence of pronoun tags that didn't come with a matching replace_dict entry.

I've added a catch to the pronoun replacement func that returns if the pronoun tag is for a cat that isn't included in the replacement dict.  This does mean that if we have any spots in our text where a pronoun is tagged in an unhandled way, then it won't smoothly default to pronouns. However, I think that if that's happening, I'd rather it be obvious so that we can fix it.

## Proof of Testing
image

## Changelog/Credits
- Rel logs now include the correct pronouns for mentioned cats

### DIFF
--- a/scripts/events_module/text_adjust.py
+++ b/scripts/events_module/text_adjust.py
@@ -49,6 +49,12 @@ def pronoun_repl(m, cat_pronouns_dict, raise_exception=False):
     inner_details = m.group(1).split("/")
     out = None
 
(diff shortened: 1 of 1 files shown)


j-gynn (approved):
This will possibly mess with the default adjective gender for localisation, but we'll burn that bridge when a localiser gets there.

PR #4321, opened 2026-02-02
[ENHANCEMENT] Update profile den buttons to only match rank

scribblecrumb:
## About The Pull Request
- Removed the ability for 'medicine cat den' buttons to show up on the profiles of sick/injured cats
- Added 'leaders den' buttons to leader profiles
- Added 'warriors den' buttons to deputy and warrior profiles

We were receiving feedback from the community that profile > den buttons always switching to the med den when the cat had a condition was prohibitive to gameplay. I don't particularly think it was a helpful feature in the first place, so this makes it so that profile den buttons simply always match the rank and do not change.  This also adds den buttons that were previously missing, like the leaders receiving leader den buttons or warriors receiving warrior den buttons.  This increases our consistency with this feature.

## Proof of Testing
image
image
image
image
image

## Changelog/Credits
- Removed the ability for 'medicine cat den' buttons to show up on the profiles of sick/injured cats
- Added 'leaders den' buttons to leader profiles
- Added 'warriors den' buttons to deputy and warrior profiles

### DIFF
--- a/scripts/screens/ProfileScreen.py
+++ b/scripts/screens/ProfileScreen.py
@@ -190,13 +190,14 @@ class ProfileScreen(Screens):
                 and event.ui_element == self.profile_elements["leader_ceremony"]
             ):
                 self.change_screen(GameScreen.CEREMONY)
-            elif event.ui_element == self.profile_elements["med_den"]:
+            elif event.ui_element == self.profile_elements.get("med_den"):
                 self.change_screen(GameScreen.MED_DEN)
-            elif (
-                "mediation" in self.profile_elements
-                and event.ui_element == self.profile_elements["mediation"]
-            ):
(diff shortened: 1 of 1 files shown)


j-gynn (changes requested):
Instead of having separate calls, couldn't all the calls be consolidated into something like `workplace` and then the only difference has to be when they're instantiated? Add a new variable like `workplace_screen` so that we can just change screen to that variable

otherwise i like it, just tryna reduce repetition

scribblecrumb:
I'm not sure I'm understanding what you mean?  If I am picking up what you mean, I'm not sure that's actually viable since each button requires different sizes (and thus also positioning).

j-gynn (approved):
I've considered more and it's probably easier to keep it this way.

Issue #4303, opened 2026-02-01
[UI] Toggle "Ignore Fullscreen Scaling Rules" Broken


Honeycase:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

Despite the toggle being disabled, the full-screen box displays as large as possible by default, as if the toggle is enabled. Toggling on and off does not appear to change anything. 

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. run clangen
2. enable full screen

### Commit # or Game Version Number

d39c4a24

### Is this bug only happening in fullscreen?

Yes

### If yes, what is your display size in pixels?

1920x1080

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/3fe43878-2271-46e9-a35c-8935c5f9780e" />

For reference, this is the screen on windowed
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a8888d5b-2d26-44e3-aca6-d7560bf1d441" />

j-gynn:
Fixed by #4332, closing

PR #4332, opened 2026-02-04
[BUGFIX] Fix minor fullscreen issues

Chinch-Bug:
## About The Pull Request
The fullscreen scaling setting is called "fullscreen scaling" in the settings file, but the actual get_setting call was looking for "fullscreen_scaling". Also, after more recent edits to how screens are tracked, the F11 keybind for toggling fullscreen would cause crashes

## Linked Issues
#4303 

## Proof of Testing
Ignore scaling rules turned off:
image

Ignore scaling rules turned on:
image

### DIFF
--- a/main.py
+++ b/main.py
@@ -236,7 +236,9 @@ while 1:
                 # debugmode.toggle_console()
             elif event.key == pygame.K_F11:
                 scripts.game_structure.screen_settings.toggle_fullscreen(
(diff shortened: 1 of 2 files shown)


j-gynn (changes requested):
Please run black using `uv run black .` and then I can merge this in.

j-gynn (approved):
Oop already done, thanks!

Issue #4075, opened 2025-10-05
[UI] acquired appearance changes initially not registering on sprite


cawcaw420908:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

when a cat gets an appearance change (gaining a scar or accessory), it doesn't show up on their sprite in camp or when you click on it. however it does show up when you click into the close-up view of the cat. this is fixed after saving and reloading the game. this is NOT fixed by a moonskip.

### Grade

Important - Has a large effect on the gameplay but is not urgent

### Steps to reproduce the behavior

1. have a cat gain a scar or accessory
2. see error

### Commit # or Game Version Number

7de0f419

### Additional context

the same problem also showed up when a different cat gained a different accessory, and when this same cat got a scar a few moons back (before the most recent commit got merged)

### Screenshots

accessory gaining event 
<img width="519" height="108" alt="Image" src="https://github.com/user-attachments/assets/f7a90fac-2f01-4a40-adc7-b635fea8bd06" />

not showing up on the cat info page
<img width="706" height="269" alt="Image" src="https://github.com/user-attachments/assets/fdd3c8e8-f183-44bd-bb9f-758d57b1e4a6" />

showing up on the cat close up
<img width="779" height="519" alt="Image" src="https://github.com/user-attachments/assets/d22e7056-a7b8-4a81-a60a-8cf2dfe9e443" />

not showing up in camp
<img width="60" height="55" alt="Image" src="https://github.com/user-attachments/assets/1c970011-d065-4e78-8048-94962bc90582" />

j-gynn:
I think I know why this is happening, the cache isn't being invalidated again. Grr.

j-gynn:
I've been told this also happens for scars, so I'll be fixing that too.

j-gynn:
This bug only happens if a cat already has one accessory and then receives another.

Hold onto this evidence of accessories working whilst I fix scars (this was awful to get and I just don't wanna have to get it twice)

<img width="648" height="130" alt="Image" src="https://github.com/user-attachments/assets/611056d4-ca8f-4f08-8b41-4feddc6eca40" />

<img width="523" height="267" alt="Image" src="https://github.com/user-attachments/assets/a690d2a7-19dc-44e4-8f40-fb9c5c76834d" />

PR #4335, opened 2026-02-05
[ENHANCEMENT] Reestablish functionality for 'scars' block in events by adding potential_scars to Injury

### LINKED ISSUE #3971: [ENHANCE] Fixing "scars" portion of injury blocks
### Description

Injury blocks in the short and patrol event format have a "scars" list, previously used to determine what scar a cat on classic gets in place of an injury. However, classic now has access to injuries, so this "scar" section does functionally nothing. 

I had *thought* it was also used to determine what scars could be available when an injury healed, however that doesn't seem to be the case.  We either never did that, or it was removed at some point.  I think it *is* good functionality to have and would be a good thing to add.

I think we could just save "potential_scars" inside the cat's injury conditions and then check for it when scars are determined.

afellowcorn:
## About The Pull Request
The 'scars' block of events and patrols is now saved as a potential_scars list in the Injury instance, allowing writers to override the standard scar pool for that injury. This also allows to scar for injuries that would not normally permit it.

Since the condition system was added to Classic mode, the 'scars' block has no actual use.
This was supposedly intended behaviour that was never added.

## Linked Issues
Closes #3971 

## Proof of Testing
Patrol injury:

Zrzut ekranu 2026-02-05 155150
Zrzut ekranu 2026-02-05 155318
Zrzut ekranu 2026-02-05 155255

Event injury:

Zrzut ekranu 2026-02-05 160715
Zrzut ekranu 2026-02-05 160646
Zrzut ekranu 2026-02-05 160705

_**Note that while testing a given scar may not appear on cat sprite until reload due to #4075.**_

### DIFF
--- a/docs/dev/writing/patrols.md
+++ b/docs/dev/writing/patrols.md
@@ -726,7 +726,7 @@ What each parameter does, and what the options are for outcomes.
 ***
 
 #### injury: List[Dict[str, various]]
->Optional. Indicates which cats get injured, and how. In classic mode, there are no conditions, so you can include a "scars" line to scar the cat instead. You can include as many of the following blocks (in a list) as you want. 
+>Optional. Indicates which cats get injured, and how. You can include as many of the following blocks (in a list) as you want. 
 >
 >```json
 >{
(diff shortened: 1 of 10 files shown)


j-gynn (approved):
LGTM!

PR #4337, opened 2026-02-05
[BUGFIX] skill test failure

scribblecrumb:
## About The Pull Request
So the reason this test would fail once in a blue moon is bc sometimes the main_cat would randomly be assigned the HUNTER skill as their secondary, leading to inconsistent results. I've prevented this by assigning a specific secondary skill.

## Proof of Testing
image

## Changelog/Credits
- `test_main_cat_skill_one` should no longer fail out of the blue


j-gynn (approved):
anything to stop the stupid test bugs. **anything.**

PR #4348, opened 2026-02-06
[ENHANCEMENT] Relationship debug command and debug console QOL

### LINKED ISSUE #4345: [ENHANCE] Easier way to set relationship values for testing purposes
### Description

When testing things that rely on the cats to have a certain relationship value, it would be handy to have an easy way to set the ideal relationship value to multiple cats to increase the chances of said things.

Treefire33:
## About The Pull Request
Adds the "relationships" command to the debug console. Current it only has one sub-command which is setting a specific relationship value (as requested by #4345). 
I also went through and did some quality-of-life stuff for the debug console, added a button that runs the currently typed command and added a way to recall the previous command by pressing the up arrow key.

I'm not too well-versed in the relationship code but a quick read and I gathered that this *should* be enough to satisfy #4345, though I could have also missed something.

The relationship command adds a way to modify relationship values in game, which removes the need to either wait for the values to occur naturally or to have to edit the values manually.

## Linked Issues
fixes: #4345 

## Proof of Testing
https://github.com/user-attachments/assets/4d17f90e-52ca-4211-8325-d95e7f6998e7

### DIFF
--- a/scripts/debug_commands/cat_relationship.py
+++ b/scripts/debug_commands/cat_relationship.py
@@ -1,0 +1,101 @@
+from typing import List
+
+from scripts.cat.cats import Cat
+from scripts.cat_relations.relationship import Relationship, RelType
+from scripts.debug_commands.command import Command
+from scripts.debug_commands.utils import add_output_line_to_log
+
+def set_cat_relationship_to_cat(cat_a: Cat, rel_type: str, cat_b: Cat, rel_value: int):
+    """
+    Sets the specified relationship type (rel_type) 
+    of cat_a towards cat_b to a specified value (rel_value)
+    """
+    from_cat_relationship: Relationship = cat_a.relationships.get(cat_b.ID)
+    if not from_cat_relationship:
+        return
+
+    if rel_type == RelType.ROMANCE:
+        from_cat_relationship.romance = rel_value
+    elif rel_type == RelType.COMFORT:
+        from_cat_relationship.comfort = rel_value
+    elif rel_type == RelType.LIKE:
+        from_cat_relationship.like = rel_value
+    elif rel_type == RelType.RESPECT:
+        from_cat_relationship.respect = rel_value
+    elif rel_type == RelType.TRUST:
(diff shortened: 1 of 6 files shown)


j-gynn:
Two things:

1. Since this is fixing #4345, please make sure to assign yourself to the issue before you start working on it with `/assign-me` so that others know not to work on it.

2. The formatter test is failing; please run `uv run black .` in your console and push the changes. 

Once you've done both, please let me know and I'll take a look at this. Thank you for contributing!

scripts/debug_commands/cat_relationship.py
```diff
@@ -0,0 +1,103 @@
...
+    if rel_type == RelType.ROMANCE:
+        from_cat_relationship.romance = rel_value
+    elif rel_type == RelType.COMFORT:
+        from_cat_relationship.comfort = rel_value
+    elif rel_type == RelType.LIKE:
+        from_cat_relationship.like = rel_value
+    elif rel_type == RelType.RESPECT:
+        from_cat_relationship.respect = rel_value
+    elif rel_type == RelType.TRUST:
+        from_cat_relationship.trust = rel_value
```
scribblecrumb: could shorten this with `set_attr`! smth like `set_attr(from_cat_relationship, rel_type, rel_value)` then there'd be no need for big if/elif statement

scripts/debug_console.py
```diff
@@ -149,6 +167,9 @@ def push_multiline(self, lines: str):
         for line in lines.split("\n"):
             self.push_line(line)
 
+    def on_close_window_button_pressed(self):
+        self.hide()
```
scribblecrumb, line 171: just double checking, is there a reason to hide instead of killing?

scribblecrumb (changes requested):

scripts/debug_console.py
Treefire33: The debug mode class changes the console's visibility when toggled rather than killing and recreating it when toggled.

scribblecrumb (approved):

PR #4354, opened 2026-02-06
ci: add comment on issue on formatting fail

archanyhm:
Instead of just failing the check, we're now informing the user of the reason for the failure and instructing them to run black to format their local changes.


archanyhm:
Drafting since I'm donig some more research on the privileges the job token gets

archanyhm:
Should be good now(?)

j-gynn (approved):

PR #4356, opened 2026-02-06
[BUGFIX] clancat status crash

scribblecrumb:
## About The Pull Request
Fixes clancat status crash by:
- Ensuring there's a default ID for the `is_lost()` func
- Matching generated litter's initial social status to their parent's current social status. this prevents them from being considered "former clancats", which is what was truly causing this bug

I think once #4295 is in, this could also be helped by making former clancat new-cats "leave" their old clan instead of being "lost" from their old clan.

Also fixes a bug that would generate extra group IDs for other clans.

## Linked Issues
this was discussed heavily in the discord

## Proof of Testing
image

Generated a shitton of abandoned litters, none of the kits are considered lost. Also debugged with a break in `become_lost()` to see if any kittens were making it there, and none were (despite their parent making it to that func and becoming lost)

## Changelog/Credits
- Abandoned kittens will no longer be considered "lost" when generated with a "former clancat" blood parent, despite being part of the player Clan (this would eventually crash the game if one of those abandoned kittens grew up and had their *own* kittens)
- Game no longer generates extra other Clan IDs upon clan creation


j-gynn:
Would you prefer #4295 to be merged first and then get those changes in that you mentioned in here?

scribblecrumb:
I think that'd be nice! It would be one less thing for me to remember to do later

j-gynn:
That one's merged in, hmu when you're ready for review

scribblecrumb:
@j-gynn ready!

j-gynn (approved):

PR #4379, opened 2026-02-07
Added debug event functionality, added a few new_cat debug events

j-gynn:
## About The Pull Request

Adds functionality for debug events. These allow devs to write simple, short events with very specific constraints (or lack thereof) to test functionality more easily. These events can never generate in-game as they are stored in a different file and have a frequency of 0. Debug events must have the word "debug" in their event_id somewhere.

To generate a debug event, you must specify it in `debug_ensure_event_id`.

At the moment, I only added a small `_debug.json` file for new_cat because that is what sparked the idea, but it will work for any other ShortEvents in the same manner.

Makes dev lives easier, especially with some difficult/rare situations (generating OtherClan cats, anyone?)

## Proof of Testing
Image

### DIFF
--- a/resources/lang/en/events/new_cat/_debug.json
+++ b/resources/lang/en/events/new_cat/_debug.json
@@ -1,0 +1,238 @@
+[
+    {
+        "event_id": "debug_newcat_loner_oldname",
+        "location": [
+            "any"
+        ],
+        "season": [
+            "any"
+        ],
+        "frequency": 0,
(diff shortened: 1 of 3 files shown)


scribblecrumb (approved):

scribblecrumb:
Not gonna make u do it for this PR, but a doc page instructing devs on how to utilize, find, and create new debug events would be really handy!

PR #4401, opened 2026-02-11
[BALANCING] Herb improvements

### LINKED ISSUE #4158: [BUG] cat receives herbs for perm condition after dying
### Pre-requisites

### Description

a cat continues to receive herbs for a permanent condition despite dying of old age the moon before. her permanent condition also wasn't deleted from her conditions tab either, even though that usually happens when cats with permcons die. 
this doesn't occur every time a cat with a permanent condition dies, as this didn't appear to happen with two of my other cats who had permanent conditions (who did not die of old age).
edit: one more moonskip still doesn't get rid of the issue

### Grade

1. have an elder with a permanent condition
2. kill that cat
3. moonskip 
4. see error

_No response_

9a58557d

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image
Image

scribblecrumb:
## About The Pull Request
This is focused on herb balancing and on implementing the herb rarities that have been sitting in the wings for a while.

- All herbs now have variable rarities within each biome/season combo
- All herbs now have tailored expiration times. This was focused on making the very commonly found + used herbs expire faster and the very rare + less used herbs expire slower.
- Herb gathering focus now provides a static buff to quantity, instead of basing the quantity buff off the number of warriors. This is to prevent the crazy insane numbers that felt rly unrealistic (like... literal 100s of herbs being gathered in a moon with an average sized clan).  The "benefit" of the gathering focus is now more about it providing an "extra" gathering round to each med cat, with the quantity effectively doubled.
- Rarity of the herb now plays into the quantity gathered. "Rare" herbs will be gathered in lower quantities, while "common" herbs will be gathered in larger quantities.
- Herbs are only added to the "in_need" list when they have low or empty supplies.  I suspect the absence of this qualifier is what was leading to wildly disproportional herb supplies ( 300 cobwebs and 1 tansy, for example). 
- We *might* have been applying herbs to any and all cats, rather than just the clan cats. Probably wasn't affecting anything, since outsiders and dead cats don't get conditions anyways, but this is a good thing to fix regardless. (actually this *was* breaking stuff lol, so this fixed #4158)

We love balancing.

Rarities have been in the works for a while, just no one wanted to sit down and figure out what all the numbers should be. So ur welcome, I finally did it.

## Proof of Testing
I did a fair amount of testing in my own Clan, seeing how well meddies kept up with demand. It seemed like they held it together fairly well, even without me doing consistent gathering patrols. Med den was generally staying at "adequate" supply, with it dipping down to "low" during leafbare, which I think is good.  Gathering focus didn't feel too overpowered, just a helpful boost, esp in newleaf when the Clan is trying to recover the herb stores in the wake of leafbare.  I noticed the range of herb amounts staying relatively proportional, though you still had some outliers, of course.  But it's nowhere near some of the wack ones I've seen previously with differences of 200-400.

image

## Changelog/Credits
- All herbs now have variable rarities within each biome/season combo
- All herbs now have tailored expiration times. This was focused on making the very commonly found + used herbs expire faster and the very rare + less used herbs expire slower.
- Herb gathering focus now provides a static buff to quantity, instead of basing the quantity buff off the number of warriors. This is to prevent the crazy insane numbers that felt rly unrealistic (like... literal 100s of herbs being gathered in a moon with an average sized clan).  The "benefit" of the gathering focus is now more about it providing an "extra" gathering round to each med cat, with the quantity effectively doubled.
- Rarity of the herb now plays into the quantity gathered. "Rare" herbs will be gathered in lower quantities, while "common" herbs will be gathered in larger quantities.
- Herbs are only added to the "in_need" list when they have low or empty supplies.  I suspect the absence of this qualifier is what was leading to wildly disproportional herb supplies ( 300 cobwebs and 1 tansy, for example).

### DIFF
--- a/scripts/clan_resources/herb/herb_supply.py
+++ b/scripts/clan_resources/herb/herb_supply.py
@@ -383,7 +383,6 @@ class HerbSupply:
 
         # get herbs found
         herb_list = []
-        list_of_herb_strs = []
         for med in med_cats:
             if assistants:
                 list_of_herb_strs, found_herbs = game.clan.herb_supply.get_found_herbs(
@@ -508,10 +507,10 @@ class HerbSupply:
                     quantity_modifier -= 0.5
                 elif rarity in (1, 2):
                     quantity_modifier += 1
-                found_herbs[herb] = int(
+                found_herbs[herb] = max(1, int(
                     choices(population=[2, 3, 4], weights=weight, k=1)[0]
                     * quantity_modifier
-                )
+                ))
                 amount_of_herbs -= 1
 
         return self.handle_found_herbs_outcomes(found_herbs)
--- a/scripts/clan_resources/herb/herb_supply.py
+++ b/scripts/clan_resources/herb/herb_supply.py
@@ -507,10 +507,13 @@ class HerbSupply:
                     quantity_modifier -= 0.5
                 elif rarity in (1, 2):
                     quantity_modifier += 1
-                found_herbs[herb] = max(1, int(
-                    choices(population=[2, 3, 4], weights=weight, k=1)[0]
-                    * quantity_modifier
-                ))
+                found_herbs[herb] = max(
+                    1,
+                    int(
+                        choices(population=[2, 3, 4], weights=weight, k=1)[0]
+                        * quantity_modifier
+                    ),
+                )
                 amount_of_herbs -= 1
 
         return self.handle_found_herbs_outcomes(found_herbs)
--- a/scripts/clan_resources/herb/herb_supply.py
+++ b/scripts/clan_resources/herb/herb_supply.py
@@ -113,7 +113,7 @@ class HerbSupply:
     @property
     def low_qualifier(self) -> int:
         """
-        returns the lowest qualifier for a low supply (supply must be higher than this qualifier and lower than or equal to the adequate qualifier)
+        returns the lowest qualifier for a low supply
         """
         return 0
 
@@ -504,7 +504,7 @@ class HerbSupply:
             # chance to find an herb is based on its rarity
             if randint(1, rarity) == 1:
                 if rarity in (5, 6):
-                    quantity_modifier -= 0.5
+                    quantity_modifier = quantity_modifier / 2
                 elif rarity in (1, 2):
                     quantity_modifier += 1
                 found_herbs[herb] = max(
--- a/scripts/clan_resources/herb/herb_supply.py
+++ b/scripts/clan_resources/herb/herb_supply.py
@@ -501,7 +501,7 @@ class HerbSupply:
             if not rarity:
                 continue
 
-            # chance to find an herb is based on it's rarity
+            # chance to find an herb is based on its rarity
             if randint(1, rarity) == 1:
                 if rarity in (5, 6):
                     quantity_modifier -= 0.5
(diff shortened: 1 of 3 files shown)


scripts/clan_resources/herb/herb_supply.py
```diff
@@ -111,7 +113,7 @@ def sorted_by_need(self) -> list:
     @property
     def low_qualifier(self) -> int:
         """
-        returns the lowest qualifier for a low supply
+        returns the lowest qualifier for a low supply (supply must be higher than this qualifier and lower than or equal to the adequate qualifier)
```
j-gynn: what an awful way of saying this /hj

if i understand correctly, what this is trying to say is "the lower bound for a low_qualifier is this variable, and the upper bound is adequate_qualifier"?

scripts/clan_resources/herb/herb_supply.py
```diff
@@ -154,7 +156,7 @@ def set_required_herb_count(self, clan_size):
         """
         takes given clan_size and multiplies it by the required_herbs_per_cat from constants.CONFIG
         """
-        self.required_herb_count = (
+        self.required_herb_count = int(
```
j-gynn, line 159: parsing this as an int blindly removes anything after the decimal (aka floor). intended?

scripts/clan_resources/herb/herb_supply.py
```diff
@@ -491,36 +492,33 @@ def get_found_herbs(
...
                 game.clan.biome
                 if not game.clan.override_biome
                 else game.clan.override_biome,
                 game.clan.current_season,
-            ):
+            )
+            if not rarity:
                 continue
 
             # chance to find an herb is based on it's rarity
```
j-gynn:
```suggestion
            # chance to find an herb is based on its rarity
```

scripts/clan_resources/herb/herb_supply.py
```diff
@@ -491,36 +492,33 @@ def get_found_herbs(
...
-            if (
-                randint(
+            if randint(1, rarity) == 1:
+                if rarity in (5, 6):
+                    quantity_modifier -= 0.5
```
j-gynn: so if it's rare and we pass a rare roll we find *less*?

this feels deeply unintuitive if that *is* what this is doing, and even worse if it's not.

j-gynn (changes requested):

scripts/clan_resources/herb/herb_supply.py
scribblecrumb: i was mostly just trying to clarify that a supply of *0* isn't considered `LOW`, bc my own brain kept getting caught by that lmao

scripts/clan_resources/herb/herb_supply.py
scribblecrumb: I'm fine with it tbh, it loses a tad bit of nuance, but I don't think a difference of 1 is going to throw anything wildly out of wack

scripts/clan_resources/herb/herb_supply.py
scribblecrumb: What I was encountering is that it felt kinda weird to have cats find like... 14 lungwort at a time, as an example. Lungwort is supposed to be rare! Cats probably shouldn't be gathering massive chunks of it at a time.  In the same vein, moss is extremely plentiful and also *very* widely used, so it feels alright (and balanced) for cats to gather large amounts.

So yes, this makes it so that rare herbs will be gathered at lower quantities.  But, tbf, rare herbs also aren't used very often and I've given them the most generous expirations.

j-gynn (approved):

PR #4407, opened 2026-02-12
[TESTS] Event filters

scribblecrumb:
## About The Pull Request
Adds some tests for event filters. Doesn't cover all of them yet cus I got tired of tests partway through lmao, but I intend to keep chipping away at them until we've covered what we can.

Covered things pretty thoroughly imo, including exclusionary values, mixed values, ect. Tried to give them all some explanatory failure messages as well.

## Why This Is Good For ClanGen
We love tests, tests are the best

## Proof of Testing
image

## Changelog/Credits
- Adds new event_filter tests covering:
  - location
  - season
  - leader life tags
  - game mode tags
  - cat age
  - cat status

PR #4437, opened 2026-02-15
[TESTS] Adds partial event_filtering tests

### LINKED ISSUE #3149: [CODE] Remove bouncy
**Description**
Apparently bouncy was accidentally removed during the personality overhaul. Let’s add it back?

### LINKED ISSUE #4431: [BUG] Incorrect reassignment of "current_rep" constraint
### Pre-requisites

### Description

War specific new cat events are not triggering at all, even when forcing new cat events through game config. Tested on most recent dev version. i know they exist in the files, as i've searched through them myself, and i know i've encountered theses events in the past, but at some point they seem to have stopped working. attached is screenshot of the console when testing this

### Grade

1. turn on force new_cat events in game config
2. set clan to be at war or get into war naturally
3. moonskip and see no new cats from the war events :(

_No response_

not sure, downloaded from https://github.com/ClanGenOfficial/clangen/pull/4430 but displays "unknown" in bottom corner despite not being a modified version 

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

j-gynn:
## About The Pull Request

**Blocked by #4407**

Adds about 2000 tests for event_filtering to ensure everything works. Added this file into the automatic test runs so it's part of our Actions testing retinue. It's a lot of new tests, yes, but most of those are in ensuring the game doesn't explode when cats switch ranks. It only takes about a second to run.

Also fixes a smattering of bugs found along the way of writing these tests.

### Bugs found & squished
- Empty `clan_relations` fields are now no longer considered an immediate fail
- For some reason, the game would explode if I tried to change outsider ranks without an associated thoughts file, so there's those too now.
- The DISCREDITS RelTier can now correctly trigger (as before it was mistagged)
- `strangers` relationship tag now works as intended
- invalid ages, skills, traits, and backstories now generate an exception if they're malformed
- Killed `bouncy` as it was causing tests to fail (since it's impossible to have a kit with this trait it's considered invalid). Related:  #3149 but quite the opposite. sorry not sorry, they bounced off a cliff 🤷

More tests ensures ClanGen's codebase is healthy and uncovers bugs faster

## Proof of Testing
image


scribblecrumb:
I know we talked about it in discord, but I'll say it here too just so we don't forget: I'd like to see the testing of cat-based constraints run through `event_for_cat()` instead of the individual private funcs. This way we'll also test whether the control function is passing all it's information correctly.

scribblecrumb (approved):

PR #4448, opened 2026-02-16
[CLEANUP] Moving element classes

scribblecrumb:
## About The Pull Request
Starting the ordeal of breaking up the `ui_elements.py` file into individual files held within `ui/elements`. For now I've only moved button classes and `TextBoxTweaked` (bc it was causing a circular import if i didn't). I'm tryna do this is smaller chunks so it doesn't feel so overwhelming.  I changed absolutely nothing about the code, this was purely moving them.

Makes it far easier to navigate these classes when they aren't in one monster file.

## Proof of Testing

https://github.com/user-attachments/assets/c11bddac-993c-44ea-bd59-e2a25ce5af1c

## Changelog/Credits
- Moves custom button classes and UITextBoxTweaked class into individual files


scripts/screens/ChangeGenderScreen.py
```diff
@@ -11,11 +11,9 @@
...
 from scripts.game_structure import game
 from scripts.game_structure.localization import load_lang_resource
-from scripts.game_structure.ui_elements import (
-    UIImageButton,
-    CatButton,
-    UISurfaceImageButton,
-)
+from ..ui.elements.cat_button import CatButton
+from ..ui.elements.image_button import UIImageButton
+from ..ui.elements.surface_image_button import UISurfaceImageButton
```
j-gynn, line 16: relative imports? i think it's an automatic pycharm thing but i wish it wouldn't

scripts/screens/ChooseAdoptiveParentScreen.py
```diff
@@ -7,11 +7,9 @@
...
 from scripts.game_structure import image_cache
 from scripts.game_structure.propagating_thread import PropagatingThread
-from scripts.game_structure.ui_elements import (
-    UIImageButton,
-    UISpriteButton,
-    UISurfaceImageButton,
-)
+from ..ui.elements.sprite_button import UISpriteButton
+from ..ui.elements.image_button import UIImageButton
+from ..ui.elements.surface_image_button import UISurfaceImageButton
```
j-gynn, line 12: same as above

j-gynn (approved):
just a question about relative imports, but since we don't have a style guide i'm supposing it's fine

PR #4454, opened 2026-02-19
[DEV CONTENT] CatFactory

j-gynn:
## About The Pull Request

Okay... buckle up. This is cleaning up part of how you create and access Cat objects, hopefully a two-part feature. This one focuses on the use of the Factory pattern to build our Cats. This solves Problem 2 outlined on Discord and copied here below for convenience/transparency.

[Context copied from Discord](https://discord.com/channels/1003759225522110524/1473387504999731412/1473387525039980544)

> ## Problem 2: The Enormous __init__
> Cat.__init__ is enormous. It's a big spaghetti of faded and unfaded cats, new cat generation and existing cat loading, random calls galore... it's a mess. Due to the aforementioned dependency issues, we can't call Cat() in lots of places, meaning we have functions like create_new_cat that sort of awkwardly shoehorn the process into working.
> 
> ### Solution: Industrialisation (CatFactory)
> Make Cat only have to think about the data it stores again, rather than having to care about whether the cat is faded and handle it differently, or whether it's a new cat and therefore needs randomization, or if it's from save data and therefore should just kind of run.
> 
> Given a dictionary data with the appropriate data:
> 
> CatFactory.create_cat(CatData.FADED, data)  # calls FadedCatFactory
> CatFactory.create_cat(CatData.NEW, data)  # calls NewCatFactory
> CatFactory.create_cat(CatData.LOAD, data)  # calls ExistingCatFactory
> 
> 
> The CatFactory class handles calling subfactories for each of the relevant CatFactorys.

There are four CatFactory variants. These are all built on an abstract base class to ensure they have the right components (and one subclasses another for reasons that will be clear shortly). These are accompanied by a new enum that describes the type of Cat being created (new, from JSON, etc.).

- CatType.NEW: this means a new cat is being made. Think kits, encountered loner, etc. Any time the game wants to randomly assign variables, it's a CatType.NEW.
- CatType.LOAD_JSON and CatType.LOAD_CSV: kinda obvious here? At the moment these both use LoadCatFactory, but I wrote it this way so that we could one day put the CSV code into its own factory. Alas, I do not understand the CSV code so I am leaving it for now or this PR would never be made.
- CatType.TEST: the associated Factory extends the CatType.NEW factory, changing the functions to guarantee the same output for all relevant variables (a newborn female clancat). It doesn't change up name or pelt generation because those are fun to leave alone and/or terrifying to mess with.
- CatType.FADED: ahh faded my beloathed. A more elegant solution will be possible once the storage of cats is divorced from the Cat class, but for now we're stuck with this one being implemented but unusable.

This can be expanded in future with more CatTypes for more specific situations if desired. I wanted to keep the initial as barebones as possible to avoid the lines changed being utterly stratospheric.

(from Discord above)
> ### Rationale
> Faded cats have a very different data structure (much smaller). Setting them up with their own factory means it'll be easier to fully divorce them from Cat in future if need be, while also removing their logic from the Cat class as much as possible.
> 
> New cats do a lot of randomisation. If this is contained within NewCatFactory, then we can control where random elements are called, and even replace the randomization function with predetermined outcomes (very useful for tests and what brought this back into my head from ages ago).
> 
> Existing cats have very limited data protection on their creation due to having to account for new cat generation. Data could be corrupted or missing and the system just merrily chugs along without a care in the world. Giving them their own ExistingCatFactory means we can print better warnings about missing data, because we know it's not meant to be missing.
> 

## Proof of Testing

Testing thread in Discord (link soon)

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -29,6 +29,7 @@ from scripts.cat.factories.typed_dicts import (
     CatTogglesDict,
     InheritanceDict,
     AfterlifeAffinityDict,
+    GenderDict,
 )
 from scripts.cat.history import History
 from scripts.cat.names import Name
@@ -148,12 +149,8 @@ class Cat:
         birth_cooldown: int,
         specsuffix_hidden=False,  # to delete once Name is decoupled from Cat
         *,
-        status_dict: StatusDict = None,
         example=False,
         faded=False,
-        skill_dict=None,
-        loading_cat=False,  # Set to true if you are loading a cat at start-up.
-        disable_random=False,
         **kwargs,
     ):
         """Initialise the cat.
@@ -181,7 +178,7 @@ class Cat:
 
         # This must be at the top. It's a smaller list of things to init, which is only for faded cats
         if faded:
-            # self.init_faded(ID, status_dict, prefix, suffix, moons, **kwargs)
+            self.init_faded(ID, moons, status, inheritance)
             return
 
         self.generate_events = GenerateEvents()
@@ -264,9 +261,6 @@ class Cat:
         self._sprite_working: bool = self.not_working()
         """used to store whether we should be displaying sick sprite or not"""
 
-        # SAVE CAT INTO ALL_CATS DICTIONARY IN CATS-CLASS
-        self.all_cats[self.ID] = self
-
         if self.ID is not None and self.ID != "0":
             Cat.insert_cat(self)
 
@@ -274,30 +268,28 @@ class Cat:
     def age(self):
         return CatAge.get_from_moons(self.moons)
 
-    def init_faded(self, ID, status, prefix, suffix, moons, **kwargs):
-        """Perform faded-specific initialization
-
+    def init_faded(
+        self, ID: str, moons: int, status: Status, inheritance: InheritanceDict
+    ):
+        """
+        Perform faded-specific initialization
         :param ID: Cat ID
-        :param status: Cat status
-        :param prefix: Cat's prefix
-        :param suffix: Cat's suffix
-        :param moons: Age in moons
-        :param kwargs:
-
-        :return: None
+        :param moons: age in moons
+        :param status: last known status
+        :param inheritance: family data
+        :return: bool
         """
         self.ID = ID
-        self.parent1 = None
-        self.parent2 = None
-        self.adoptive_parents = []
-        self.mate = []
-        self.status = Status(**status) if status else Status()
+        self.parent1 = inheritance["parent1"]
+        self.parent2 = inheritance["parent2"]
+        self.adoptive_parents = inheritance["adoptive_parents"]
+        self.faded_offspring = inheritance["faded_offspring"]
+        self.mate = inheritance["mate"]
+        self.status = status
         self._pronouns = {}  # Needs to be set as a dict
         self.moons = moons
         self.inheritance = None  # This should never be used, but just for safety
-        self.name = Name(prefix=prefix, suffix=suffix, cat=self)
-
-        self.set_faded()  # Sets the faded sprite and faded tag (self.faded = True)
+        # name is assigned by FadedCatFactory
         return True
 
     def __repr__(self):
@@ -2841,34 +2833,49 @@ class Cat:
             return False
 
         if isinstance(cat_info["status"], str):
-            status_dict = {"rank": cat_info["status"]}
+            status = Status(rank=cat_info["status"])
+            # they are definitely dead
+            status.send_to_afterlife(
+                CatGroup.DARK_FOREST_ID
+                if cat_info.get("df", False)
+                else CatGroup.STARCLAN_ID
+            )
         else:
-            status_dict = cat_info["status"]
+            status = Status(**cat_info["status"])
 
         cat_ob = Cat(
             ID=cat_info["ID"],
-            prefix=cat_info["name_prefix"],
-            suffix=cat_info["name_suffix"],
-            status_dict=status_dict,
+            gender_dict=GenderDict(sex=None, genderalign=None),
+            pelt=None,
             moons=cat_info["moons"],
+            status=status,
+            backstory="",
+            catskills=None,
+            personality=None,
+            mentorship={},
+            inheritance=InheritanceDict(
+                parent1=cat_info["parent1"],
+                parent2=cat_info["parent2"],
+                adoptive_parents=cat_info["adoptive_parents"],
+                mate=[],
+                previous_mates=[],
+                faded_offspring=cat_info["faded_offspring"],
+            ),
+            affinity={},
+            toggles={},
+            experience=0,
+            birth_cooldown=0,
+            specsuffix_hidden=False,
             faded=True,
         )
-        if cat_info["parent1"]:
-            cat_ob.parent1 = cat_info["parent1"]
-        if cat_info["parent2"]:
-            cat_ob.parent2 = cat_info["parent2"]
-        cat_ob.faded_offspring = cat_info["faded_offspring"]
-        cat_ob.adoptive_parents = (
-            cat_info["adoptive_parents"] if "adoptive_parents" in cat_info else []
+
+        cat_ob.name = Name(
+            prefix=cat_info["name_prefix"],
+            suffix=cat_info["name_suffix"],
+            specsuffix_hidden=False,
+            load_existing_name=True,
+            cat=cat_ob,
         )
-        cat_ob.faded = True
-
-        if cat_info.get("df"):
-            cat_ob.status.send_to_afterlife(target_ID=CatGroup.DARK_FOREST_ID)
-        elif isinstance(cat_info["status"], str):
-            cat_ob.status.send_to_afterlife(target_ID=CatGroup.STARCLAN_ID)
-
-        cat_ob.dead_for = cat_info["dead_for"] if "dead_for" in cat_info else 1
 
         return cat_ob
 
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -183,12 +183,16 @@ class Cat:
         self._moons = None
 
         # Public attributes
+        self.ID = ID
+
         if isinstance(gender, dict):
             self.gender = gender["sex"]
             self.genderalign = gender["genderalign"]
         else:
             self.gender = gender
             self.genderalign = gender
+
+        self.moons = moons
         self.status: Status = Status(**status_dict) if status_dict else Status()
         self.backstory = backstory
         self.age: Optional[CatAge] = None
@@ -242,8 +246,6 @@ class Cat:
         self.specsuffix_hidden = specsuffix_hidden
         self.inheritance = None
 
-        self.ID = ID
-
         # backstory
         if self.backstory is None:
             self.backstory = "clanborn"
@@ -252,8 +254,12 @@ class Cat:
 
         # These things should only run when generating a new cat, rather than loading one in.
         if not loading_cat:
-            self.personality = init_params["personality"]
-            self.experience = init_params["experience"]
+            if init_params is None:
+                init_params = {}
+            self.personality = init_params.get(
+                "personality", Personality(lawful=6, social=6, aggress=6, stable=6)
+            )
+            self.experience = init_params.get("experience", 0)
 
         if "biome" in kwargs:
             biome = kwargs["biome"]
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -24,6 +24,12 @@ from scripts.cat.enums import (
     CatCompatibility,
     CatThought,
 )
+from scripts.cat.factories.typed_dicts import (
+    MentorshipDict,
+    CatTogglesDict,
+    InheritanceDict,
+    AfterlifeAffinityDict,
+)
 from scripts.cat.history import History
 from scripts.cat.names import Name
 from scripts.cat.pelts import Pelt
@@ -126,24 +132,27 @@ class Cat:
 
     def __init__(
         self,
-        prefix=None,
-        gender=None,
+        ID: str,
+        gender_dict: dict,
+        pelt: Pelt,
+        moons: int,
+        status: Status,
+        backstory: str,
+        catskills: CatSkills,
+        personality: Personality,
+        mentorship: MentorshipDict,
+        inheritance: InheritanceDict,
+        affinity: AfterlifeAffinityDict,
+        toggles: CatTogglesDict,
+        experience: int,
+        birth_cooldown: int,
+        specsuffix_hidden=False,  # to delete once Name is decoupled from Cat
+        *,
         status_dict: StatusDict = None,
-        backstory="clanborn",
-        parent1=None,
-        parent2=None,
-        adoptive_parents=None,
-        suffix=None,
-        specsuffix_hidden=False,
-        ID=None,
-        moons=None,
         example=False,
         faded=False,
         skill_dict=None,
-        pelt: Pelt = None,
-        init_params: dict = None,
         loading_cat=False,  # Set to true if you are loading a cat at start-up.
-        *,
         disable_random=False,
         **kwargs,
     ):
@@ -172,7 +181,7 @@ class Cat:
 
         # This must be at the top. It's a smaller list of things to init, which is only for faded cats
         if faded:
-            self.init_faded(ID, status_dict, prefix, suffix, moons, **kwargs)
+            # self.init_faded(ID, status_dict, prefix, suffix, moons, **kwargs)
             return
 
         self.generate_events = GenerateEvents()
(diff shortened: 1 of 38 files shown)


j-gynn:
Marking as a draft for now whilst the beta testers digest the bugs

j-gynn:
Marking as ready for review as the betas haven't been able to find any more major issues!

scripts/cat/factories/create_cat.py
```diff
@@ -0,0 +1,98 @@
+from __future__ import annotations
+
+from random import randint, sample, choice
+
+from scripts.cat.cats import Cat
```
scribblecrumb: unused import?

scripts/cat/factories/create_cat.py
```diff
@@ -0,0 +1,98 @@
...
+from scripts.cat.enums import CatRank, CatAge
+from scripts.cat.factories.cat_factory import CatFactory
+from scripts.cat.factories.enums import CatType
+from scripts.cat.pelts import Pelt
+from scripts.game_structure import game
+
+
+def create_cat(rank, moons=None, biome=None, cat_type: CatType = CatType.TEST):
```
scribblecrumb, line 12: docstring pls?

Also, I'm a tad confused about this file's use vs. the various factories. Is this just a kind of "misc" creation to cover the two types of example cats we make? It seems like this func sees a lot of use in tests, despite us having a test cat factory

scripts/cat/factories/create_cat.py
```diff
@@ -0,0 +1,98 @@
...
+    ]
+
+    new_cat.pelt.scars = tuple(
+        scar for scar in new_cat.pelt.scars if scar not in not_allowed_scars
+    )
+
+    return new_cat
+
+
+def create_example_cats():
```
scribblecrumb, line 45: docstring pls?

scripts/cat/factories/typed_dicts.py
```diff
@@ -0,0 +1,42 @@
+from typing import TypedDict, Optional, List, Dict
+
+
```
scribblecrumb, line 5: Would it be reasonable to move the existing StatusDict here as well?

scripts/cat/factories/typed_dicts.py
```diff
@@ -0,0 +1,42 @@
...
+    previous_mates: List[str]
+
+
+class InheritanceDict(TypedDict):
+    parent1: Optional[str]
+    parent2: Optional[str]
+    adoptive_parents: List[str]
+    faded_offspring: List[str]
+    mate: List[str]
+    previous_mates: List[str]
```
scribblecrumb, line 31: Why do we include mate/previous_mates in two different dicts? It doesn't look like `MateshipDict` sees any use anyways

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+        self,
+        status_dict: Optional[Union[Dict, str]],
+        moons: int,
+        old_bools: List[Optional[bool]],
+    ) -> Status:
+        """
+        Check & convert status to new Status
+        :param status_dict: Possible status_dict
+        :param moons: age in moons
+        :param old_bools: old-style status bools in a tuple
```
scribblecrumb: "in a tuple"? But the param is typed as a list?

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+        if "pattern" in kwargs:
+            kwargs["tortie_marking"] = kwargs["pattern"]
+            del kwargs["pattern"]
+
+        # just to be sure that scars exists as a list
+        kwargs["scars"] = kwargs.get("scars", [])
+
+        for specialty in ("specialty", "specialty2"):
+            if old_scars := kwargs.get(specialty):
+                kwargs["scars"] = [*kwargs["scars"], old_scars]
```
scribblecrumb: Shouldn't this be a tuple? It seems like Pelt does some conversion, but does that need to happen in Pelt or can we just do it now

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,287 @@
...
+        return self.rng.randint(Cat.age_moons[age][0], Cat.age_moons[age][1])
+
+    def _determine_age_moons_and_status(
+        self, moons, status_dict
+    ) -> Tuple[CatAge, int, Status]:
+        """
+
+        :param moons:
+        :param status_dict:
+        :return: moons and status_dict
```
scribblecrumb: incomplete docstring? or did u mean to leave it like this?

scripts/cat/cats.py
```diff
@@ -126,53 +132,54 @@ class Cat:
...
+        :param catskills: CatSkills object
+        :param personality: Personality object
+        :param mentorship: MentorshipDict containing mentor data and apprentice data, including former for both
+        :param inheritance: Inheritance object
+        :param affinity: AffinityDict containing starclan & dark forest affinity values
+        :param toggles: Dict of cat-related behavior toggles
+        :param experience: Cat's experience value
+        :param birth_cooldown: How many moons that must pass before this cat can give birth again
+        :param specsuffix_hidden: Whether to show or hide the "special suffix" for a cat's name
+        :param example: I actually don't know what this does
```
scribblecrumb: I believe this was previously used to mark a cat as being an 'example' cat via the make clan screen? might be obsolete at this point

scripts/cat/cats.py
```diff
@@ -181,175 +188,74 @@ def __init__(
...
+        self.former_apprentices = mentorship["former_apprentices"]
+
+        # inheritance
+        self.parent1 = inheritance["parent1"]
+        self.parent2 = inheritance["parent2"]
+        self.adoptive_parents = inheritance["adoptive_parents"]
+        self.faded_offspring = inheritance["faded_offspring"]
+        """Stores of a list of faded offspring, for relation tracking purposes"""
         self.mate = []
         self.previous_mates = []
```
scribblecrumb: shouldn't these be assigned here as well? i.e `= inheritance["mate"]`

scripts/cat/cats.py
```diff
@@ -3028,34 +2839,49 @@ def load_faded_cat(cat: str):
             return False
 
         if isinstance(cat_info["status"], str):
-            status_dict = {"rank": cat_info["status"]}
+            status = Status(rank=cat_info["status"])
+            # they are definitely dead
+            status.send_to_afterlife(
+                CatGroup.DARK_FOREST_ID
+                if cat_info.get("df", False)
+                else CatGroup.STARCLAN_ID
```
scribblecrumb, line 2847: UR cats can fade too, do we need to account for that here?

scribblecrumb (changes requested):

scripts/cat/cats.py
j-gynn: This is what the original code did... and I don't think we have UR faded artwork ?

scripts/cat/factories/create_cat.py
j-gynn: I didn't write this, I have no idea what it does. I just had to move it out of Cat haha

scripts/cat/factories/create_cat.py
j-gynn: This is a pre-existing function and I don't honestly understand where it's used and why. I can go through and replace every call with the CatFactory if you prefer?

scripts/cat/factories/new_cat_factory.py
j-gynn: whoops.

scripts/cat/factories/typed_dicts.py
j-gynn: `MateshipDict` was my old solution until I decided it didn't need to exist. Just forgot to delete it. Removing it now~

scripts/cat/cats.py
j-gynn: ah good catch! thank you!!

scripts/cat/factories/create_cat.py
j-gynn: It does one additional thing that's not currently in TestCatFactory, actually - it removes scars that cause limb loss. So there's that. At present, TestCatFactory doesn't touch the randomization of Pelt or Name because we just don't test for them yet.

scripts/cat/cats.py
scribblecrumb: image

:P we do

i'm sure it's a v rare case tho, which is why it's never really caused a problem before

scripts/cat/factories/create_cat.py
scribblecrumb: Ah, it creates the cats for the MakeClan screen!

scripts/cat/factories/create_cat.py
scribblecrumb: Ah okay, didn't realize you hadn't altered it at all. 

I think it causes some confusion to have this as well as CatFactory (with it's own `create_cat` funcs).  Like I looked at this next to all the CatFactory scripts and went "Huh?"

iirc this is just used in the creation of MakeClanScreen cat "examples". That's why it removes scars that are related to permanent conditions, bc someone (read: me) didn't feel like going to the struggle of detecting the presence of those scars and adding the related perm condition to the cat.

Proposed solution:
Rename this file to `initial_cat_generation` or something else indicating it's use in creating that initial group of kitties for players to pick from (or maybe `generate_example_cat` so that it also covers the func that's used in creating the example cats in the EventEdit screen?)

scribblecrumb: Oh, and I would prefer to see the tests actually use TestCatFactory where possible. I don't believe we really have any tests where rando pelts will be a problem.

scripts/cat/factories/base_factory.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class BaseCatFactory(ABC):
+    @abstractmethod
+    def __init__(self, rng):
+        pass
+
+    @abstractmethod
+    def create_cat(self, **kwargs) -> Cat:
+        pass
```
archanyhm, line 17: Is there any reason to choose **kwargs here?

scripts/cat/factories/base_factory.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+
+from scripts.cat.cats import Cat
+
+
+class BaseCatFactory(ABC):
+    @abstractmethod
+    def __init__(self, rng):
```
archanyhm: Type hints, please

scripts/cat/factories/cat_factory.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class CatFactory:
+    rng = random.Random()
+
+    __factories = {
+        CatType.NEW: NewCatFactory,
+        CatType.LOAD_JSON: LoadCatFactory,
+        CatType.LOAD_CSV: LoadCatFactory,
+        CatType.FADED: LoadCatFactory,
```
archanyhm, line 18: Why do these three map to the same factory? what's the point of differentiating between them at that point?

scripts/cat/factories/base_factory.py
archanyhm: Small addendum: I know that the point is to provide one unified interface for calling `create_cat` on, but that's not really serving any purpose. The interface being unified makes sense if --- and only if --- they are mostly appearing the same to the caller. In this case, if we expect vastly different parameters based on which type of factory we intend to call, the unified interface does nothing but to occlude the actual signature that the underlying methods understand. Imo if we have different sets of "constructors" for cats, we should use separate methods instead that have a clearly defined interface.

scripts/cat/factories/create_cat.py
archanyhm: > This is a pre-existing function and I don't honestly understand where it's used and why. I can go through and replace every call with the CatFactory if you prefer?

In cases like this, I think it's best to add a comment in the code indicating that this was mostly moved from a previous location, and that no/little changes have been applied since it makes it easier to put into context.

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+from scripts.cat.history import History
+from scripts.cat.names import Name
+from scripts.cat.pelts import Pelt
+from scripts.cat.personality import Personality
+from scripts.cat.skills import CatSkills
+from scripts.cat.status import Status
+
+with open(f"resources/dicts/conversion_dict.json", "r", encoding="utf-8") as read_file:
+    CONVERT = ujson.loads(read_file.read())
+
```
archanyhm, line 14: This should not happen in the top level scope of the file, since this introduces side effects when importing the module.

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+
+class LoadCatFactory(BaseCatFactory):
+    cat_id = None
+
+    def __init__(self, rng):
+        self.rng = rng  # turns out we do need rng.
```
archanyhm: Pointless comment; explain "why" instead

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+        status_dict: Optional[Union[Dict, str]],
+        moons: int,
+        old_bools: List[Optional[bool]],
+    ) -> Status:
+        """
+        Check & convert status to new Status
+        :param status_dict: Possible status_dict
+        :param moons: age in moons
+        :param old_bools: old-style status bools in a list
+        :return: valid status
```
archanyhm: Extremely opaque in meaning. Convert status to new Status? What's the difference between them? Why does it matter? Those things should be outlined in the docstring.

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+        if eye_color2 == "BLUE2":
+            eye_color2 = "COBALT"
+
+        return eye_color, eye_color2
+
+    def _build_pelt(self, kwargs):
+        """
+        Handles some check & convert functionality for pelts
+        :param kwargs: Everything we've ever passed into the factory
+        :return: A dict of the keys needed to build the pelt
```
archanyhm: I don't think this docstring provides any value. "some check & convert functionality" is very vague. "Everything we've ever passed into the factory" and "a dict of the keys needed to build the pelt" all don't really provide much additional meaning. This is probably an indication that the kwargs-based approach occludes actual semantics.

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+                trait=trait,
+                kit_trait=is_kit_trait,
+                lawful=facets[0],
+                social=facets[1],
+                aggress=facets[2],
+                stable=facets[3],
+            )
+        else:
+            print(f"WARNING: no facets found for cat ID: {self.cat_id}")
+            return Personality(trait=trait, kit_trait=is_kit_trait)
```
archanyhm:
```suggestion
        if facets is None:
            print(f"WARNING: no facets found for cat ID: {self.cat_id}")
            return Personality(trait=trait, kit_trait=is_kit_trait)

       
        facets = [int(i) for i in facets.split(",")]
        return Personality(
            trait=trait,
            kit_trait=is_kit_trait,
            lawful=facets[0],
            social=facets[1],
            aggress=facets[2],
            stable=facets[3],
        )
```

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+    ) -> Tuple[CatSkills, str]:
+        """
+        Handle conversion of some *very old* skills & backstories
+        :param skill_dict: modern skill dict
+        :param skill: skill string
+        :param backstory: backstory string
+        :param rank: needed to generate new skills
+        :param age: needed to generate new skills
+        :return:
+        """
```
archanyhm: `def _convert_skill` -> `Handle conversion of some *very old* skills` **& backstories**

`def _convert_skills_and_backstories`?

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,341 @@
...
+                else:
+                    backstory = "clanborn"
+            return CatSkills.get_skills_from_old(skill, rank, age), backstory
+        else:
+            raise Exception(f"No skill data provided for cat ID: {self.cat_id}")
+
+    def _convert_history(self, died_by, scar_events, cat) -> History:
+        """
+        Unfortunately, this has to be handled *after* the creation of the cat
+        because of the horrible nested cat. fixme.
```
archanyhm: Not sure why this comment is here. I'd put it at the callsite instead and give it a proper description.

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+            load_existing_name=True,
+            cat=cat,
+        )
+
+        Cat.all_cats[cat.ID] = cat
+
+        return cat
+
+    def _random_age(self):
+        return self.rng.choice([*CatAge])
```
archanyhm, line 120:
```suggestion
    def _get_random_age(self):
        return self.rng.choice([*CatAge])
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def _random_age_from_rank(self, rank):
```
archanyhm, line 122:
```suggestion
    def _get_random_age_by_rank(self, rank):
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        :param rank: Provided cat's rank
+        :return: Age the cat should be
```
archanyhm, line 125: "Age the cat should be"?

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        return self.rng.choice(
+            [
+                CatAge.YOUNG_ADULT,
+                CatAge.ADULT,
+                CatAge.ADULT,
+                CatAge.SENIOR_ADULT,
+            ]
+        )
+
+    def _random_status_from_age(self, age):
```
archanyhm, line 148:
```suggestion
    def _get_random_status_by_age(self, age):
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        # it's a bit silly that we do this, then undo it,  and finally redo in Cat() but i don't want this refactor getting huge
```
archanyhm, line 149: prefer extending the "why" instead of this tangent

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        status.generate_new_status(age, disable_random=type(self.rng) != BASE_RNG)
+
+        return status
+
+    def _random_moons(self, age: CatAge) -> int:
+        """
+        Generate random moons appropriate for the given age
+        :param age: CatAge
+        :return: Appropriate moons
+        """
```
archanyhm, line 160:
```suggestion
    def _get_random_moons_for_age(self, age: CatAge) -> int:
        """
        Generate random moons appropriate for the given age
        :param age: CatAge
        :return: Appropriate moons
        """
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+            moons = self._random_moons(age)
+        else:
+            status = None
+
+        if not isinstance(moons, int) or not status or not age:
+            raise Exception("Something went wrong generating age, moons or status_dict")
+
+        return age, moons, status
+
+    def _random_gender_and_genderalign(self, age) -> dict:
```
archanyhm, line 205:
```suggestion
    def _get_random_gender_and_genderalign(self, age) -> dict:
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+            gender["genderalign"] = "nonbinary"
+        elif trans_chance == 1:
+            gender["genderalign"] = (
+                "trans male" if gender["sex"] == "female" else "trans female"
+            )
+
+        return gender
+
+    @staticmethod
+    def _random_pelt(gender, parents, age):
```
archanyhm, line 227:
```suggestion
    def _get_random_pelt(gender, parents, age):
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        return Pelt.generate_new_pelt(
+            gender,
+            tuple(Cat.fetch_cat(i) for i in parents if i),
+            age,
+        )
+
+    def _random_personality(self, age: CatAge):
```
archanyhm, line 234:
```suggestion
    def _get_random_personality(self, age: CatAge):
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+        if type(self.rng) != BASE_RNG:
+            return Personality(
+                lawful=8, social=8, aggress=8, stable=8, kit_trait=age.is_baby()
+            )
+        return Personality(kit_trait=age.is_baby())
+
+    def _random_experience(self, age, moons: int) -> int:
```
archanyhm, line 241:
```suggestion
    def _get_random_experience(self, age, moons: int) -> int:
```

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,288 @@
...
+            )
+        elif age == CatAge.SENIOR:
+            return self.rng.randint(
+                Cat.experience_levels_range["expert"][0],
+                Cat.experience_levels_range["master"][1],
+            )
+        else:
+            return 0
+
+    def _random_skills_dict(self, rank, age):
```
archanyhm, line 273:
```suggestion
    def _get_random_skills_dict(self, rank, age):
```

scripts/cat/factories/test_cat_factory.py
```diff
@@ -0,0 +1,50 @@
+from scripts.cat.enums import CatAge
+from scripts.cat.factories.new_cat_factory import NewCatFactory
+from scripts.cat.personality import Personality
+from scripts.cat.skills import CatSkills, SkillPath
+from scripts.cat.status import Status
+
+# this is a patchwork fix for now
```
archanyhm, line 7: why?

scripts/cat/factories/test_cat_factory.py
archanyhm, line 1: Generally all the renaming suggestions from the NewFatFactory apply to this as well

archanyhm (changes requested):

scripts/cat/factories/load_cat_factory.py
j-gynn: ah, much more elegant! I just pulled this whole thing from the original code that handled it with only the barest modicum of thought to make sure it'd still work 😅

scripts/cat/factories/new_cat_factory.py
j-gynn: Yeah that's a bad description, isn't it? "Random age appropriate for the cat's rank" maybe?

scripts/cat/factories/test_cat_factory.py
j-gynn: Honestly? Because I dipped my toes into the random.Random module and discovered I really, **really** do not know where to start in getting a sensible deterministic output. This fix is just ignoring the neat setup I did for `self.rng` and instead explicitly returning expected values, knowing this is a test cat

archanyhm: Okay, my bad, haha --- what I meant was "this comment should explain the why", like "why is it a patchwork fix? why is it necessary that way?"

scripts/cat/factories/new_cat_factory.py
archanyhm: Yes, that sounds good!

scripts/cat/factories/load_cat_factory.py
j-gynn: Where would you suggest putting it? In the class itself?

archanyhm: Yep

j-gynn:
I feel like I've implemented this wrong (idk, maybe it's the 59-parameter function) but I've created a rudimentary mapper. Upside: all our parameters are now explicit. Downside, you can't play any old saves.

j-gynn:
Closing this PR to target a separate clangen branch as these changes are getting a bit too big for just one PR.

Issue #3983, opened 2025-08-16
[BUG] cat younger than 150 moons dies of old age


Vickarooni:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

for context; my leader is 146 moons old, and is mates with a cat who is 170. my guess is that i got the event where two elderly mates die together, and the game didn't check for if both cats are older than 150, just one, so my leader lost all his lives before he hit the intended age where old age deaths start to happen

### Grade

Regular - Should be handled soon, but doesn't necessarily break the game or inhibit the player

### Steps to reproduce the behavior

1. have two elderly mates (with just one being older than 150 moons)
2. moonskip until you get the event "gen_death_age_any5"
3. see error

### Commit # or Game Version Number

38f74dd

### Additional context

unsure if the cat dying before he's supposed to has anything to do with him being leader, so i'm also mentioning it to be sure. also, the history text has an extra period as an added little issue
if i find any other problems with cats dying of old age when younger than 150 moons, i will update this issue accordingly

### Screenshots

<img width="507" height="144" alt="Image" src="https://github.com/user-attachments/assets/0d33feb8-7cf9-4864-9ac3-64557765f0ca" />

both mates die together

<img width="688" height="276" alt="Image" src="https://github.com/user-attachments/assets/8a773d22-9954-4ee6-94ad-ba7b06e7f008" />

airblaze is way older than 150 moons, so the old age death is correct...

<img width="761" height="540" alt="Image" src="https://github.com/user-attachments/assets/3a0884e7-ec65-4b87-a2a0-7a1066ec13f1" />

...but johnnystar hasn't hit that threshold yet, so he shouldn't be able to die from this event

j-gynn:
Still exists. This appears to be a limitation rather than a bug per-se.

Our code checks that both are *seniors*, which has a threshold of 120 moons and up. Intended behaviour, but technically unintended consequence.

Currently, only m_c is checked for the `old_age` threshold - as of 8ab8271694d0032fd0c547d933ba8a0bf29b2154, this is on `line 395` in the function `filter_events` in `short_event_generation.py`. We'd need a check for `r_c`'s `old_age` threshold where `dies` is True. We could also do a new tag, but I think the check will cover most everything else.

PR #4462, opened 2026-02-20
fix: cat younger than 150 moons dies of old age

tibisabau:
## About The Pull Request

Fixed a bug where cats younger than 150 moons could die of old age when paired with an older mate in death events. The issue occurred because the code only checked if the main cat (m_c) had reached the old age death threshold (150 moons), but did not verify that the relationship cat (r_c) also met this requirement when both cats died together in old age events.

Added a check in the `filter_events` function in `short_event_generation.py` to ensure that when a relationship cat (r_c) dies in an old age event, they must also be at least 150 moons old. This prevents situations where a younger leader or cat loses all their lives prematurely due to their older mate's death.

## Linked Issues

#3983

## Proof of Testing

- All existing tests passed: 20 tests in `test_handle_short_events.py` and `test_event_filters.py` passed successfully
- Pylint validation passed with 0 errors on the modified file
- The fix ensures both cats in the "gen_death_age_any5" event (and similar old age death events) meet the 150 moon threshold before the event can trigger

## Changelog/Credits

**The code in this pull request was generated by GitHub Copilot with the Claude Sonnet 4.5 model.**

### DIFF
--- a/scripts/events_module/short/short_event_generation.py
+++ b/scripts/events_module/short/short_event_generation.py
@@ -398,6 +398,15 @@ def filter_events(
             < constants.CONFIG["death_related"]["old_age_death_start"]
         ):
             continue
+        # check for old age for r_c when they die
+        if (
+            random_cat
+            and "old_age" in event.sub_type
+            and event.r_c.get("dies", False)
+            and random_cat.moons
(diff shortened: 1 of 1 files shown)


archanyhm:
Closing since we are currently not accepting primarily AI generated contributions.

PR #4470, opened 2026-02-22
[DEV/savefile-rework] CatFactory

j-gynn:
## About The Pull Request

Okay... buckle up. This is cleaning up part of how you create and access Cat objects, hopefully a two-part feature. This one focuses on the use of the Factory pattern to build our Cats. This solves Problem 2 outlined on Discord and copied here below for convenience/transparency.

[Context copied from Discord](https://discord.com/channels/1003759225522110524/1473387504999731412/1473387525039980544)

> ## Problem 2: The Enormous __init__
> Cat.__init__ is enormous. It's a big spaghetti of faded and unfaded cats, new cat generation and existing cat loading, random calls galore... it's a mess. Due to the aforementioned dependency issues, we can't call Cat() in lots of places, meaning we have functions like create_new_cat that sort of awkwardly shoehorn the process into working.
> 
> ### Solution: Industrialisation (CatFactory)
> Make Cat only have to think about the data it stores again, rather than having to care about whether the cat is faded and handle it differently, or whether it's a new cat and therefore needs randomization, or if it's from save data and therefore should just kind of run.
> 
> Given a dictionary data with the appropriate data:
> 
> CatFactory.create_cat(CatData.FADED, data)  # calls FadedCatFactory
> CatFactory.create_cat(CatData.NEW, data)  # calls NewCatFactory
> CatFactory.create_cat(CatData.LOAD, data)  # calls ExistingCatFactory
> 
> 
> The CatFactory class handles calling subfactories for each of the relevant CatFactorys.

There are four CatFactory variants. These are all built on an abstract base class to ensure they have the right components (and one subclasses another for reasons that will be clear shortly). These are accompanied by a new enum that describes the type of Cat being created (new, from JSON, etc.).

- CatType.NEW: this means a new cat is being made. Think kits, encountered loner, etc. Any time the game wants to randomly assign variables, it's a CatType.NEW.
- CatType.LOAD_JSON and CatType.LOAD_CSV: kinda obvious here? At the moment these both use LoadCatFactory, but I wrote it this way so that we could one day put the CSV code into its own factory. Alas, I do not understand the CSV code so I am leaving it for now or this PR would never be made.
- CatType.TEST: the associated Factory extends the CatType.NEW factory, changing the functions to guarantee the same output for all relevant variables (a newborn female clancat). It doesn't change up name or pelt generation because those are fun to leave alone and/or terrifying to mess with.
- CatType.FADED: ahh faded my beloathed. A more elegant solution will be possible once the storage of cats is divorced from the Cat class, but for now we're stuck with this one being implemented but unusable.

This can be expanded in future with more CatTypes for more specific situations if desired. I wanted to keep the initial as barebones as possible to avoid the lines changed being utterly stratospheric.

(from Discord above)
> ### Rationale
> Faded cats have a very different data structure (much smaller). Setting them up with their own factory means it'll be easier to fully divorce them from Cat in future if need be, while also removing their logic from the Cat class as much as possible.
> 
> New cats do a lot of randomisation. If this is contained within NewCatFactory, then we can control where random elements are called, and even replace the randomization function with predetermined outcomes (very useful for tests and what brought this back into my head from ages ago).
> 
> Existing cats have very limited data protection on their creation due to having to account for new cat generation. Data could be corrupted or missing and the system just merrily chugs along without a care in the world. Giving them their own ExistingCatFactory means we can print better warnings about missing data, because we know it's not meant to be missing.
> 

## Linked Issues

Related: #4454

## Proof of Testing

Testing thread in Discord (link soon)

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -29,6 +29,7 @@ from scripts.cat.factories.typed_dicts import (
     CatTogglesDict,
     InheritanceDict,
     AfterlifeAffinityDict,
+    GenderDict,
 )
 from scripts.cat.history import History
 from scripts.cat.names import Name
@@ -148,12 +149,8 @@ class Cat:
         birth_cooldown: int,
         specsuffix_hidden=False,  # to delete once Name is decoupled from Cat
         *,
-        status_dict: StatusDict = None,
         example=False,
         faded=False,
-        skill_dict=None,
-        loading_cat=False,  # Set to true if you are loading a cat at start-up.
-        disable_random=False,
         **kwargs,
     ):
         """Initialise the cat.
@@ -181,7 +178,7 @@ class Cat:
 
         # This must be at the top. It's a smaller list of things to init, which is only for faded cats
         if faded:
-            # self.init_faded(ID, status_dict, prefix, suffix, moons, **kwargs)
+            self.init_faded(ID, moons, status, inheritance)
             return
 
         self.generate_events = GenerateEvents()
@@ -264,9 +261,6 @@ class Cat:
         self._sprite_working: bool = self.not_working()
         """used to store whether we should be displaying sick sprite or not"""
 
-        # SAVE CAT INTO ALL_CATS DICTIONARY IN CATS-CLASS
-        self.all_cats[self.ID] = self
-
         if self.ID is not None and self.ID != "0":
             Cat.insert_cat(self)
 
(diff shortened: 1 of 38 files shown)


scribblecrumb:
Do u still want this or can we close it now that the other CatFactory PR got merged?

j-gynn:
Closing it!

PR #4471, opened 2026-02-22
[BRANCHWORK] CatFactory

j-gynn:
## About The Pull Request

Adds the CatFactory and its changes to the savefile_rework branch.

This will allow us to build a better foundation based on some of its changes.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -29,6 +29,7 @@ from scripts.cat.factories.typed_dicts import (
     CatTogglesDict,
     InheritanceDict,
     AfterlifeAffinityDict,
+    GenderDict,
 )
 from scripts.cat.history import History
 from scripts.cat.names import Name
@@ -148,12 +149,8 @@ class Cat:
         birth_cooldown: int,
         specsuffix_hidden=False,  # to delete once Name is decoupled from Cat
         *,
-        status_dict: StatusDict = None,
         example=False,
         faded=False,
-        skill_dict=None,
-        loading_cat=False,  # Set to true if you are loading a cat at start-up.
-        disable_random=False,
         **kwargs,
     ):
         """Initialise the cat.
@@ -181,7 +178,7 @@ class Cat:
 
         # This must be at the top. It's a smaller list of things to init, which is only for faded cats
         if faded:
-            # self.init_faded(ID, status_dict, prefix, suffix, moons, **kwargs)
+            self.init_faded(ID, moons, status, inheritance)
             return
 
         self.generate_events = GenerateEvents()
@@ -264,9 +261,6 @@ class Cat:
         self._sprite_working: bool = self.not_working()
         """used to store whether we should be displaying sick sprite or not"""
 
-        # SAVE CAT INTO ALL_CATS DICTIONARY IN CATS-CLASS
-        self.all_cats[self.ID] = self
-
         if self.ID is not None and self.ID != "0":
             Cat.insert_cat(self)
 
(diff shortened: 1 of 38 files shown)


scripts/cat/factories/create_example_cat.py
scribblecrumb, line 1: rename this file to `generate_example_cat` for clarity as suggested in the prior PR?

scripts/cat/factories/create_cat.py
```diff
@@ -0,0 +1,70 @@
...
+                ]
+            )
+            game.choose_cats[cat_index] = CatFactory.create_cat(
+                cat_type=CatType.NEW, rank=random_rank, no_disabling_scars=True
+            )
+
+
+def create_option_preview_cat(scar: str = None, acc: str = None):
+    """
+    Creates a cat with the specified scar.
```
scribblecrumb:
```suggestion
    Creates a cat with the specified scar or accessory.
```
i know you didn't write this, we just might as well fix it now tho

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,350 @@
...
+        mate = kwargs.get("mate")
+        mate = mate if isinstance(mate, list) else [mate]
+        inheritance = InheritanceDict(
+            parent1=kwargs["parent1"],
+            parent2=kwargs["parent2"],
+            adoptive_parents=kwargs.get("adoptive_parents", []),
+            faded_offspring=kwargs.get("faded_offsprings", []),
+            mate=mate,
+            previous_mates=kwargs.get("previous_mates", []),
+        )
```
scribblecrumb: is there a reason you wanna define mate ahead of time on line 57 vs just moving everything you're doing there to line 63?

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,350 @@
...
+        mentorship = MentorshipDict(
+            mentor=kwargs["mentor"],
+            former_mentor=kwargs.get("former_mentor", []),
+            patrol_with_mentor=kwargs.get("patrol_with_mentor", 0),
+            apprentice=kwargs["current_apprentice"],
+            former_apprentices=kwargs["former_apprentices"],
+        )
+
+        toggles = CatTogglesDict(
+            no_kits=kwargs["no_kits"],
```
scribblecrumb: should this also be a `.get` or is it exempt for some reason

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,350 @@
...
+                kwargs.get("dead"),
+                kwargs.get("df"),
+                kwargs.get("driven_out"),
+                kwargs.get("exiled"),
+                kwargs.get("outside"),
+            ],
+        )
+
+        backstory = self._convert_backstory(kwargs.get("backstory"))
+        cat_skill, backstory = self._convert_skill_and_backstory(
```
scribblecrumb: if we've renamed other `catskills` to just `skills`, then let's do it here too

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,350 @@
...
+        return pelt
+
+    @staticmethod
+    def _convert_backstory(backstory) -> str:
+        """
+        Convert an old-style backstory to the new version
+        :param backstory:
+        :return: the new-style backstory
+        """
+        # if the key isn't found, return it as the value (no need to convert
```
scribblecrumb:
```suggestion
        # if the key isn't found, return it as the value (no need to convert)
```

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,350 @@
...
+        return BACKSTORIES["conversion"].get(backstory, backstory)
+
+    def _build_personality(self, facets, trait, is_kit_trait):
```
scribblecrumb: docstring?

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,311 @@
...
+from scripts.game_structure import game, constants
+
+BASE_RNG = random.Random
+
+
+class NewCatFactory(BaseCatFactory):
+    def __init__(self, rng: random.Random):
+        self.rng = rng if rng else BASE_RNG()
+
+    def create_cat(self, **overrides):
```
scribblecrumb, line 28: docstring?

scripts/cat/cats.py
```diff
@@ -126,53 +132,54 @@ class Cat:
...
+        personality: Personality,
+        mentorship: MentorshipDict,
+        inheritance: InheritanceDict,
+        affinity: AfterlifeAffinityDict,
+        toggles: CatTogglesDict,
+        experience: int,
+        birth_cooldown: int,
+        specsuffix_hidden=False,  # to delete once Name is decoupled from Cat
+        *,
         example=False,
```
scribblecrumb, line 151: is `example` still used or can we just remove it?

scribblecrumb (dismissed):

scripts/cat/factories/load_cat_factory.py
j-gynn: There's no way that I know of to do an elegant inline check here... `kwargs.get("mate", []) if isinstance(kwargs.get("mate"), list) else kwargs.get("mate", []),` is what it'd work out being. Though there *is* a bug here, so I'll fix that.

scripts/cat/factories/load_cat_factory.py
j-gynn: It didn't have one on the original, whereas the others all did. I agree, it's weird. I'll add it.

scripts/cat/cats.py
j-gynn: It's still used by the example cats function, unfortunately.

scripts/cat/factories/load_cat_factory.py
scribblecrumb: naw i see the logic in having `mate = kwargs.get("mate")`, cus doing the kwargs.get every time is a pain in the ass, but couldn't line 63 just be `mate=mate if isinstance(mate, list) else [mate]`?

j-gynn: Good point. I'll change it.

scribblecrumb:
lgtm but i'll let arch have a chance to look at it too

archanyhm (commented):
lgtm for now

scribblecrumb (approved):

PR #4486, opened 2026-02-28
Localization names

j-gynn:
### Blocked by (#4485) - do not merge until that's in.

This does technically make it easier to add name packs, I guess? If you don't want to do the undocumented, secret TXT file method.

- Finally allows for localized names, assuming they follow a prefix-suffix format.
- Adds a player-level override on localized names, for if you want to continue using English names despite the presence of custom names.
- Adds a language-level config option to disable normal prefixes from appearing in the Clan name pool. This is probably most relevant for languages with animacy concerns.
- Fixes a bug that would cause the game to crash if an error message containing non-English characters was printed.

## Linked Issues

Blocked by #4485

## Changelog/Credits

- Allows for localized names, assuming they follow a prefix-suffix format.
- Fixes a bug that would cause the game to crash if an error message containing non-English characters was printed.


scribblecrumb:
conflicts on this <3

scribblecrumb (approved):

PR #4526, opened 2026-03-09
[CLEANUP] Rel event review

scribblecrumb:
## About The Pull Request
This reviews some of the extra relationship reactions attached to relationship events and adjusts them to better fit the event text.  The focus here was on narrowing reactions that felt too broad (such as a wholesale decrease across a bunch of rel types) and evaluating affected types to see if a different one would be more appropriate.

I also cleaned up a couple things I noticed along the way, like replacing long inclusionary lists with short exclusionary ones. I didn't go wild with this, there's only a couple instances of it.

This is a planned improvement alongside the rel log stuff.  The goal is to re-evaluate these reactionary values to see if they still fit well after the big relationship rework we had.  

## Proof of Testing
image

We're still working!

## Changelog/Credits
- Adjusts reactionary relationship changes attached to various relationship events
- Fixes a couple typos within relationship events
- Cleans up some relationship event params

### DIFF
--- a/resources/lang/en/events/relationship_events/group_interactions/4/negative.json
+++ b/resources/lang/en/events/relationship_events/group_interactions/4/negative.json
@@ -15,28 +15,28 @@
 				"comfort": "increase"
 			},
 			"m_c_to_r_c2": {
-				"like": "decrease"
+				"respect": "decrease"
 			},
 			"r_c1_to_r_c2": {
-				"like": "decrease"
+				"respect": "decrease"
 			},
 			"r_c2_to_m_c":{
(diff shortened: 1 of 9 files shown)


resources/lang/en/events/relationship_events/group_interactions/4/negative.json
```diff
@@ -15,28 +15,28 @@
 				"comfort": "increase"
```
tobes03, line 15: Should there be an equivalent r_c1 to m_c comfort increase on this one?

tobes03 (approved):

PR #4530, opened 2026-03-09
[BUGFIX] Invited Cats Joining the Clan As Mediators When Settings Don't Allow mediators

### LINKED ISSUE #4524: [BUG] Invited cats sometimes join the clan as mediators even though the settings don't allow it
### Pre-requisites

### Description

Basically what I said in the title, found this on accident in gameplay when I had the setting unchecked and my invited cats have a 1/3rd chance of becoming mediators though they technically shouldn't. 

Since it looks like an easy fix, I'll try to see if I can make a bugfix for it tomorrow.

### Grade

1. Have the "become mediator" setting off
2. Have an outside cat and invite them into the clan
3. If that cat doesn't become mediator, keep reloading the game they do join as one

_No response_

7e2a22b

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

The mediator setting isn't checked

Image
Image

I invited this cat (she gave up her kit to the clan , so it shouldn't be saying she rejoined with her kitten, it's a separate bug) and she decided to join the clan as a mediator even though the setting wasn't checked.

Arcticnightsky:
## About The Pull Request
Sometimes when invited cats join the clan, and the mediator setting isn't checked, they still have a 1/3 chance of choosing the mediator rank because the code never checked that. This fixes that so if mediators aren't allowed in the game, the cat chooses to become a warrior instead. Good for those who hadn't read that far into the series so they don't have to play with a rank they aren't yet familiar with.

## Linked Issues
Fixes #4524

## Proof of Testing
After inviting around 10 different cats into my clan with the mediator setting off, they all chose to become either warriors or medicine cats, no mediators at all.

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -526,6 +526,9 @@ def handle_lead_den_event():
                         # if cat is an apprentice, make sure they get a mentor!
                         if invited_cat.status.rank == CatRank.APPRENTICE:
                             invited_cat.update_mentor()
+                        # if the cat chose to become a mediator but the settings don't allow it, make them a warrior or medicine cat instead
+                        if invited_cat.status.rank == CatRank.MEDIATOR and not get_clan_setting("become_mediator"):
+                            invited_cat.status._change_rank(random.choice([CatRank.WARRIOR, CatRank.MEDICINE_CAT]))
 
                     invited_cat.create_relationships_new_cat()
 
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -527,8 +527,13 @@ def handle_lead_den_event():
                         if invited_cat.status.rank == CatRank.APPRENTICE:
                             invited_cat.update_mentor()
                         # if the cat chose to become a mediator but the settings don't allow it, make them a warrior or medicine cat instead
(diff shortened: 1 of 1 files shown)


scripts/events.py
```diff
@@ -526,6 +526,14 @@ def handle_lead_den_event():
...
                         if invited_cat.status.rank == CatRank.APPRENTICE:
                             invited_cat.update_mentor()
+                        # if the cat chose to become a mediator but the settings don't allow it, make them a warrior or medicine cat instead
+                        if (
+                            invited_cat.status.rank == CatRank.MEDIATOR
+                            and not get_clan_setting("become_mediator")
+                        ):
+                            invited_cat.status._change_rank(
+                                random.choice([CatRank.WARRIOR, CatRank.MEDICINE_CAT])
+                            )
```
scribblecrumb: Let's default to just a warrior instead of warrior or med cat, I don't want this to skew us toward more medicine cats than we want.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #4532, opened 2026-03-10
[Platform] iPad support

jbcpollak:
## About The Pull Request

My kids wanted to run clangen on their iPads, so after about 5 full time days of work, here it is.

This PR makes a series of mostly minor infrastructure changes to clangen to make it work on iPads/iOS devices. It would be great if some of those changes could be adopted, if for no other reason than to make maintaining a separate iOS branch easier. I'd be happy to break changes up into separate PRs if that would make things easier.

The changes are:

* Python 3.13 
* Pygame must be >= 2.5.6
* Fallback to Python internal json when native-ujson is not available
* disable the auto-updater on iOS
   * required by the platform, and also works around issues with native versions of the cryptography library being a pain
* force full screen on iOS and disable quit button
* main game loop is now async - this is a requirement for iOS, works fine locally on desktop too
   * this also allows the game to suspend when its in the background
* introduced a PlatformManager protocol and created one specifically for iOS vs Desktop
   * addresses logging and user_data_dir() differences

This stores data in ICloud, so it should theoretically sync across devices or if you need to wipe your iDevice as well. Untested on iPhones.

Also, I used Gemini to help diagnose and recommend each of the above changes, I reviewed the code changes myself and they look good.

I'll be honest and say since there is probably limited chance of a fan-game making it onto the App Store, I'm not sure this can every be broadly distributed, however I did the work and wanted to share - if anyone out there wants to compile and run it yourself locally you can.

I think some of the above changes are good modernizations to the codebase anyway, if someone is able to take responsibility for bundling and deploying to the App Store and get it approved, more people could play.

## Linked Issues

You will need to use build_app.py [from this PR](https://github.com/seekerluke/pygame-ios-templates/pull/4) to get it to compile.

## Proof of Testing

PXL_20260310_050620105


main.py
```diff
@@ -1,12 +1,21 @@
+# === This is require for iOS specific file handling. DO NOT MOVE THIS! ===
+# This is required to allow the game to load files from the app bundle on iOS, 
+# which has an unstable relative path. By adding the app bundle to the start of sys.path, 
+# we can ensure that imports will work correctly.
+import sys, os
+if getattr(sys, "platform", "") == "ios":
+    sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+
```
jbcpollak: I don't like this, but there is a path-search problem in iOS's bundled python without it. If there is interest in actually merging this PR I will work to remove this or move it up to the bundling framework so it doesn't pollute the game's code.

scripts/game_structure/screen_settings.py
```diff
@@ -330,13 +330,14 @@ def load_manager(res: Tuple[int, int], screen_offset: Tuple[int, int], scale: fl
         font_name="clangen", regular_path="resources/fonts/clangen.ttf"
     )
 
+    screen_scale_path = os.path.join(get_cache_dir(), "screen_scale.json")
     generate_screen_scale(
         "resources/theme/master_screen_scale.json",
-        "resources/theme/generated/screen_scale.json",
+        screen_scale_path,
```
jbcpollak, line 336: this is probably technically a bug - as is the game is writing the custom scaling theme back to the application directory, but it should be writing it to the user directory.

resources/dicts/events/reformat_events.py
```diff
@@ -1,4 +1,4 @@
-import json as ujson
+from utils.json_compat import json
```
jbcpollak, line 1: I found a few places were Python json was being imported to look like ujson, I'm not sure if that was intentional but if so a comment might be a good idea.

I figured normalizing it to all be the same was the best case

PR #4553, opened 2026-03-15
[CONTENT] New Loving Content

tobes03:
## About The Pull Request

This is a fairly modest PR aimed at adding content to "loving" in line with its new trait dictionary definition. This has included moving some of its thoughts to compassionate (or deleting them outright if they are basically identical to already-existing compassionate thoughts).

Some of the content hints at unhealthy relationship dynamics, which is intentional. I didn't want to go too far with this, but part of the discussion around "loving" is that it's not a purely positive trait -- it can be jealous, possessive, obsessive, etc. I wanted to add that dimension to its writing.

Reduces overlap between loving and compassionate. Makes "loving" in-game content more accurate to its intended definition.

## Proof of Testing

image


resources/lang/en/events/lead_ceremony_df.json
```diff
@@ -1031,7 +1031,7 @@
       "rank": ["warrior", "elder", "medicine cat", "mediator", "apprentice", "deputy", "medicine cat apprentice", "mediator apprentice"],
       "life_giving": [
         {
-            "text": "r_c arrives in a plume of gray smoke, prowling up to m_c with a snarl rumbling in {PRONOUN/r_c/poss} chest. {PRONOUN/r_c/subject/CAP} {VERB/r_c/warn/warns} m_c that {PRONOUN/m_c/poss} soft-hearted nature will be a weakness if {PRONOUN/m_c/subject} {VERB/m_c/want/wants} to hold onto power. With a last contemptuous look, r_c gives a life for [virtue].",
```
scribblecrumb, line 1034: could we preserve this OG text under a different trait? compassionate maybe?

scribblecrumb (changes requested):

resources/lang/en/events/lead_ceremony_df.json
tobes03: yes! done

scribblecrumb:
Btw, tests are failing because you lower-cased the season names! while i don't particularly like the capitalization either, i don't feel like fixing up the schema test atm. So you should revert those changes.

tobes03:
> Btw, tests are failing because you lower-cased the season names! while i don't particularly like the capitalization either, i don't feel like fixing up the schema test atm. So you should revert those changes.

oh shoot, i just thought it was a mistake. well, all fixed now

scribblecrumb (approved):

PR #4584, opened 2026-03-20
[BUGFIX] intensity tagging

scribblecrumb:
## About The Pull Request
Intensities just got mistagged in the rel review.

## Proof of Testing
image

timeskipped and got no warning messages about missing interactions

## Changelog/Credits
- medium intensity relationship events will now generate correctly


j-gynn (approved):

Vickarooni:
commenting here just to bring up that i think this bug is happening again, i'm getting the same error spam but it's complaining about mostly like events instead of respect (though i got respect once)
image

PR #4602, opened 2026-03-23
[BUGFIX] Unconstrain some kit thoughts

scribblecrumb:
## About The Pull Request
Separates the first group of `KITTEN` thoughts so that we have an entirely unconstrained group and then a group that implies the presence of other cats.  Previously, all thoughts either required a second cat, or were constrained to certain injuries, seasons, or traits.  While a very rare circumstance, this would break things if a kitten was the only cat in the Clan.  It's also one of the reasons tests are currently failing (though not the main one, so tests will still fail)

## Proof of Testing
image

successfully found thoughts for the kitten!

## Changelog/Credits
- Separates a group of kitten thoughts into two: an entirely unconstrained group and a group that implies the presence of other cats.  This means that if a kitten is the only cat in the Clan, they'll now receive a valid thought instead of erroring.


resources/lang/en/thoughts/while_alive/kitten.json
```diff
@@ -20,39 +17,47 @@
             "Is trying to growl menacingly",
             "Is adamantly refusing to take {PRONOUN/m_c/poss} nap",
```
tobes03: move "Is adamantly refusing to take {PRONOUN/m_c/poss} nap", too?

resources/lang/en/thoughts/while_alive/kitten.json
```diff
@@ -3,14 +3,11 @@
         "id": "gen_kit",
         "thoughts": [
             "Plays moss ball by {PRONOUN/m_c/self}",
-            "Wonders who {PRONOUN/m_c/poss} mentor will be",
             "Wants to take a nap",
             "Tries to sneak out of camp",
```
tobes03: Move "Tries to sneak out of camp", too? I feel like the implication that they could be caught by someone warrants it.

tobes03 (changes requested):

tobes03 (approved):

PR #4627, opened 2026-03-25
[BUGFIX] Fix tick fever localization inconsistency.

### LINKED ISSUE #4376: [BUG] inconsistency with tick fever/illness, always uses default strings
### Pre-requisites

### Description

while i was perusing around the files because i was trying to report that tick fever was too common, i discovered that in the gain_illness_strings.json, tick fever is instead referred to as "tick _illness_" even though i have never seen this term being used in-game. checking injuries_risk_string.json, it is correctly referred to as tick fever, so there appears to be a kind of inconsistency with this condition's name. could be related, but it also appears to be going through a similar issue as the illness strings, where it is not using any of them proper and is instead using default text ("[cat]" has gotten tick fever")

### Grade

1. moonskip until cats get tick fever
2. you will always get the default text

(this is also mostly a problem with the text strings rather than something that can be reproduced in game. after all, the strings are not displaying correctly & the condition seems to use two different names?)

_No response_

e648ef3

### Error log

_No response_

### Additional context

_No response_

### Screenshots

unsure what to put here, so i guess i'll leave screenshots of the strings that are not getting used as they do in fact exist in the game files:

Image
<br>
^ strings for "tick illness" a condition that is never seen in-game with this name. i was told it's supposed to be the same thing as tick fever, but for some reason it's called differently and these strings are never used. these are in gain_illness_strings.json.
<br>
<br>

Image
<br>
^ these are for the used "tick fever", albeit these strings aren't used either because they're always replaced with default text rather than the more detailed events. these are in injuries_risk_strings.json.

amilleah:
## About The Pull Request

Changes "tick illness" to "tick fever" in `condition_got_strings/gain_illness_strings.json`

## Linked Issues

Fixes #4376 

## Proof of Testing

Created tick bites conditions.json for all cats in a clan
Increased `expanded_illness_chance` to 10/25 from 10/250
Reload game and moonskip until Cat's condition progresses to illness. 

Find attached changed "tick fever" strings appearing in-game.

There's an unrelated issue of r_c appearing as a string. Can modify current PR or open a separate PR if not currently addressed or planned. 

tick bites progressed to fever

tick fever condition

### DIFF
--- a/resources/lang/en/conditions/condition_got_strings/gain_illness_strings.json
+++ b/resources/lang/en/conditions/condition_got_strings/gain_illness_strings.json
@@ -41,9 +41,9 @@
     "starving": [
         "m_c is starving."
     ],
-    "tick illness": [
-        "r_c worries over m_c, looking for any tick-heads an apprentice might have left, but {PRONOUN/r_c/subject} can only find a distinctive rash. m_c knows what it means- tick illness.",
-        "m_c's heart sinks as {PRONOUN/m_c/poss} fur starts falling out. The apprentices must have missed a tick- r_c confirms it's tick illness.",
-        "m_c complains about a new rash, going silent when r_c finds a tick-head. {PRONOUN/m_c/subject/CAP} has tick illness."
+    "tick fever": [
+        "r_c worries over m_c, looking for any tick-heads an apprentice might have left, but {PRONOUN/r_c/subject} can only find a distinctive rash. m_c knows what it means- tick fever.",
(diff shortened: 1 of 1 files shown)


scribblecrumb (approved):

scribblecrumb:
I coulda sworn I fixed that r_c-in-string problem a while ago, but maybe it was just a similar adjacent bug. Feel free to fix that in a different PR!

PR #4628, opened 2026-03-25
[BUGFIX] Add old_age_death event check for multiple cats

### LINKED ISSUE #3983: [BUG] cat younger than 150 moons dies of old age

amilleah:
## About The Pull Request

Checks if any cat is younger than the `old_age_death_start` during `short_event_generation`.

Prevents cats from dying of old age in paired events (e.g. `gen_death_age_any5`) where only one cat meets the age threshold.

## Linked Issues

Fixes #3983

## Proof of Testing

1. Make a cat couple elders (one < 150 moons, one > 150 moons)
3. Force bug event with `game_config.toml`:
```toml
[event_generation]
  debug_ensure_event_id = "gen_death_age_any5"
```
4. Moonskip; the event is not selected:

Where Cliffspots is 170 moons and Sablestar is 140 moons.

### DIFF
--- a/scripts/events_module/short/short_event_generation.py
+++ b/scripts/events_module/short/short_event_generation.py
@@ -396,12 +396,18 @@ def filter_events(
                     continue
 
         # check for old age
-        if (
-            "old_age" in event.sub_type
-            and main_cat.moons
-            < constants.CONFIG["death_related"]["old_age_death_start"]
-        ):
-            continue
+        if "old_age" in event.sub_type:
+            if (
+                main_cat.moons
+                < constants.CONFIG["death_related"]["old_age_death_start"]
+            ):
+                continue
+            if (
+                random_cat
+                and random_cat.moons
+                < constants.CONFIG["death_related"]["old_age_death_start"]
+            ):
+                continue
         # remove some non-old age events to encourage elders to die of old age more often
         if (
             "old_age" not in event.sub_type
@@ -600,6 +606,15 @@ def filter_events(
             failed_ids.append(chosen_event.event_id)
             final_events.remove(chosen_event)
             chosen_event = None
+        elif (
+            "old_age" in chosen_event.sub_type
+            and chosen_cat.moons
+            < constants.CONFIG["death_related"]["old_age_death_start"]
+        ):
+            failed_ids.append(chosen_event.event_id)
+            final_events.remove(chosen_event)
+            chosen_event = None
+            chosen_cat = None
         else:
             break
 


scribblecrumb (changes requested):
I don't think this is quite how we want to handle it.  This fix means that the event would still be triggered, the mentioned cats just wouldn't die if they weren't old enough. Players would still see the death event, though, leading to confusion.

Where this *should* be fixed is within the event filtering itself, in `short_event_generation`.  The problem is with the second involved cat in particular, as the filtering for the main cat already works.  It's a bit of a weird one, since we can't really include it in the `cat_for_event` filtering func.  You'd need to account for passed random_cats in the section around line 400 of `short_event_generation` as well as down by line 588 where we find a random_cat if one wasn't passed from the start.  Let me know if you need further direction 👍

amilleah:
oh okay perfect! that makes much more sense, I'll amend the commit and keep this open for review~ thanks!!

scripts/events_module/short/short_event_generation.py
```diff
@@ -553,6 +553,12 @@ def filter_events(
         else:
             chosen_event = random.choice(final_events)
 
+    # check if random cat meets old_age threshold
+    if chosen_event and "old_age" in chosen_event.sub_type:
+        if chosen_cat.moons < constants.CONFIG["death_related"]["old_age_death_start"]:
+            chosen_event = None
+            chosen_cat = None
+
```
scribblecrumb: Can we move this initial check up either to where we check the main_cat's old age or to where we check the possible event against the random_cat?  If a random_cat was determined from the very start, then we don't want to decouple that random_cat from the event that gets chosen. Rather, we want to find an event that fits with both the main_cat and random_cat, or find no event at all.

For further context: atm short events either have a random_cat determined before an event is found, or they find an event and look for a random_cat that will fit with it.  The majority of the time, we utilize the latter method of finding an event first.  However, we still have to account for a random_cat being predetermined, mostly in the cases of FutureEvents such as murder reveals.  The main onus of a FutureEvent is that they create an event that involves specific, predetermined cats.  If a predetermined cat somehow disqualifies all possible events, then the FutureEvent is *supposed* to fail and never be triggered (for example, a specific random_cat is supposed to reveal a murder. however they die before they can! instead of just finding a different random_cat to use, the murder should instead never be revealed.)

This all means that we have to avoid ever decoupling a predetermined random_cat from the event.

scribblecrumb (changes requested):

scripts/events_module/short/short_event_generation.py
amilleah: Ok! When I force the event in game_config.toml:
```toml
[event_generation]
  debug_ensure_event_id = "gen_death_age_any5"
```
The only output I get is for the younger cat:
```text
WARNING: no death: [] events found for Sablestar
```

It looks like the event is correctly filtered out, but I'm unsure if this output is intended (previously would fail + show the warning: no death events for both cats).

scribblecrumb: Ah, maybe I wasn't clear enough before. I was saying that this section should be moved up to around line 400 where we check old age for the main_cat.  So rather than checking random_cat *here*, we need to check it alongside main_cat earlier.

scribblecrumb (approved):

PR #4629, opened 2026-03-26
Fix for new ghosts not having death history

### LINKED ISSUE #4201: [BUG] UR cats created as ghosts don't have death history
### Pre-requisites

### Description

outsider cats who were created as alive and died later have the death history "this cat died while roaming around", but outsider cats who were created dead (e.g. cats created from the injured cat thunderpath failure outcome, dead parents created when generating adopted kittens) do not. not sure if intended. 

### Grade

1. timeskip until a same-sex couple adopts a litter of kits (just one way, other way is sending cats on patrols until you get one that creates ghost cats) 
2. go into the kits' profile
3. go into the dead parents' profile
4. check history
5. see error

_No response_

87567839

### Error log

_No response_

### Additional context

_No response_

### Screenshots

(vervain was a loner created through a moon event and died later - his history is displaying correctly)
Image

(periwinkle was a cat created to be a dead parent to an adopted litter of kits. he has no death history)
Image

Rowannothere:
## About The Pull Request

Bug fix for new ghosts not having death history. It might be a bit hacky, I feel like 2 lines of code is just too easy?? But it works and doesn't break anything so?? I also noticed that newly generated dead cats from other Clans don't generate death history, either, and didn't fix that yet, though. I just wanted to make sure this code was ok first LOL

## Linked Issues

Fixes: #4201 

## Proof of Testing
Screenshot (2489)
Screenshot (2490)
Screenshot (2491)
Screenshot (2492)

### DIFF
--- a/scripts/events_module/consequences.py
+++ b/scripts/events_module/consequences.py
@@ -738,6 +738,9 @@ def create_new_cat(
         # KILL >:D only if we're sposed to tho
         if not alive:
             new_cat.die()
+            if new_cat.status.group_ID == "3":
+                death_history = i18n.t("events.death.outsider_deaths.history.loner")         
+                new_cat.history.add_death(death_text=death_history)
 
         # newbie thought
         new_cat.get_new_thought(thought)
--- a/scripts/events_module/consequences.py
+++ b/scripts/events_module/consequences.py
@@ -18,8 +18,7 @@ from scripts.cat_relations.enums import RelType
 from scripts.clan_package.settings import get_clan_setting
 from scripts.game_structure import game, constants
 from scripts.cat.constants import BACKSTORIES, PERMANENT
-from scripts.events_module.text_adjust import process_text, adjust_list_text
-
+from scripts.events_module.text_adjust import process_text, adjust_list_text, history_text_adjust
 
 def create_new_cat_block(
     Cat: Optional["Cat"],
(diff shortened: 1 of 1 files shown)


scribblecrumb (changes requested):
I think it might be better to base the history message off of the cat's rank.  You use the `loner` history here, but the dead cat might be a rogue or kittypet.  While I believe they have essentially "duplicate" histories, that might change in the future (and might also differ for various localizations).

scripts/events_module/consequences.py
```diff
@@ -738,6 +737,22 @@ def create_new_cat(
...
+            elif new_cat.status.rank == CatRank.KITTYPET:
+                death_history = i18n.t("events.death.outsider_deaths.history.kittypet")
+                new_cat.history.add_death(death_text=history_text_adjust(death_history, other_clan_name=None,clan=game.clan))
+            elif new_cat.status.is_other_clancat:
+                for item in game.clan.all_other_clans:
+                    for entry in new_cat.status.group_history:
+                        if entry.get("group") == item.group_ID:
+                            other_clan_name = f"{item.name}Clan"
+                death_history = i18n.t("events.death.outsider_deaths.history.other_clan")
+                new_cat.history.add_death(death_text=history_text_adjust(death_history, other_clan_name, game.clan))
```
scribblecrumb, line 755:
```suggestion
            if new_cat.status.is_other_clancat:
                for item in game.clan.all_other_clans:
                    for entry in new_cat.status.group_history:
                        if entry.get("group") == item.group_ID:
                            other_clan_name = f"{item.name}Clan"
                death_history = i18n.t("events.death.outsider_deaths.history.other_clan")
                new_cat.history.add_death(death_text=history_text_adjust(death_history, other_clan_name, game.clan))
            else:
                death_history = i18n.t(f"events.death.outsider_deaths.history.{new_cat.status.rank}")
                new_cat.history.add_death(death_text=history_text_adjust(death_history, other_clan_name=None, clan=game.clan))
```
I think this would still work, but with less repetition (but test it lol, typing code in github comments isn't the easiest thing.)

Since the ranks are string enums, we can treat them like strings! So besides other_clan cats, we can use that rank string to pull the correct history text.  Since other_clan is the exception here, we just handle them more manually first, then once they're taken care of we *should* only have valid outsider ranks that we know will have lang strings matching their rank.

scribblecrumb (changes requested):

scribblecrumb:
Hello! Just giving this a poke in case you didn't notice the review comment.

Ryos00:
May be reopened once changes are made according to the former review.

PR #4650, opened 2026-03-28
[BUGFIX] Tests

scribblecrumb:
## About The Pull Request
This is aimed at fixing or temp patching various test problems causing tests to fail atm
- Just put an early return on the status_history test to effectively skip it until jgynn finds time to fix it.  The game code isn't really the problem here, rather the test implementation itself. So I don't see a problem with just skipping it for now.  I don't want gynn to feel like he needs to rush to get it fixed, so I think this is a fine enough patch for the issue as it stands.
- pylint was crying about an import for some reason. an adjustment to the import line seems to have fixed it?
- A crash was occurring, courtesy of the `dies` key not always being present. fixed via `.get()`
- idk when it happened, but `test_interactions` was royally messed up.  just passing the entirely wrong args into a not really appropriate func!  I adjusted the func used and now it works as intended.
- The skill test in that file was also just using the entirely wrong skill tags. Again, no clue when this happened? Easy fix though.
-  There was an intermittent issue where guides could generate as newborns with a mismatched rank. This was due to the `disable_random` variable sometimes being left as `True` past when it should be. I implemented a "reset" of sorts to ensure it goes back to `False` when we want it to.
- Every so often the `mentor_app` test would fail. I think that might've been due to randomness, so I added `disable_random` to those cat creations to hopefully fix it.

failing tests bad

## Proof of Testing
If tests pass ;)

## Changelog/Credits
- Temp patch on the `status_history` test to disable it until more extensive fixes can be made
- Fixed pylint import crying
- Fixed crash related to the `dies` key not always being present
- Fixed various problems within the `test_interactions` script
- Fixed intermittent freshkill test problem where guides would generate as newborns, but without a newborn rank
- Removed some randomness from `mentor_app` test to hopefully prevent intermittent failures.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -171,6 +171,8 @@ class Cat:
 
         if disable_random is not None:
             Cat.disable_random = disable_random
+        else:
+            Cat.disable_random = False
 
         if (
             faded
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -169,10 +169,7 @@ class Cat:
 
         self._history = None
 
-        if disable_random is not None:
-            Cat.disable_random = disable_random
-        else:
-            Cat.disable_random = False
+Cat.disable_random = bool(disable_random)
 
         if (
             faded
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -169,7 +169,7 @@ class Cat:
 
         self._history = None
 
-Cat.disable_random = bool(disable_random)
+        Cat.disable_random = bool(disable_random)
 
         if (
             faded
--- a/tests/test_event_filters.py
+++ b/tests/test_event_filters.py
@@ -491,8 +491,10 @@ class TestInterpersonalRelationshipConstraints(unittest.TestCase):
             )
 
     def test_mentor_app(self):
-        app = Cat(moons=8)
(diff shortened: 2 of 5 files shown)


scripts/cat/cats.py
```diff
@@ -171,6 +171,8 @@ def __init__(
 
         if disable_random is not None:
             Cat.disable_random = disable_random
+        else:
+            Cat.disable_random = False
```
j-gynn: there's actually a reason the `disable_random` was persistent, but i understand that it's buggy. the reason I didn't set it up like this was so that tests could just globally say "everything in here is disabled" and then use a post-test teardown to revert it. if you wanna do this then we might as well remove the variable... but if not, ig use this
```suggestion
Cat.disable_random = bool(disable_random)
```

tests/test_event_filters.py
```diff
@@ -1241,6 +1243,7 @@ def test_statuses_lost(self):
             self.assertFalse(event_for_cat(cat=cat, cat_info={"status": [f"-lost"]}))
 
     def test_status_history(self):
+        return  # temp patch until the test can be fixed proper
```
j-gynn, line 1246: is there a github issue associated with this? I'll try and remember to get on it but i'm concerned i might forget lol

j-gynn (changes requested):
one thing and an inquiry

j-gynn (approved):

PR #4651, opened 2026-03-28
[BUGFIXES] Accurate Death Text & Scar Clean-up for Misc Patrol Files

tobes03:
## About The Pull Request

The other patrol files. As with the others, mostly a process of removing the scar lines. I edited a few patrol outcomes because they were missing pronoun tags

## Proof of Testing

image

### DIFF
--- a/resources/lang/en/patrols/disaster.json
+++ b/resources/lang/en/patrols/disaster.json
@@ -917,15 +917,12 @@
                         ],
                         "injuries": [
                             "battle_injury"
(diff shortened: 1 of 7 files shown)


tobes03:
image

just pushed another commit to fix a pronoun error I saw on Discord. I think the problem was that there was no new cat block so the pronoun wasn't being generated correctly.

scribblecrumb (approved):

PR #4682, opened 2026-04-01
[BUGFIX] Fix dead_for much longer than 'dead_moons' bug

### LINKED ISSUE #4475: [UI] Cats in the dark forest say they've been dead for much longer then they have been
### Pre-requisites

### Description

Some of the cats in the dark forest have insane length of time that they've been dead. Past the amount of moons I've even played. This doesn't seem to be an issue with any Starclan cats?

### Grade

1. Have a cat in the dark forest
2. time skip around a bit and check how long they've been dead for

(does not happen with every single dark forest cat)

_No response_

43a3dd25

No

_No response_

_No response_

### Additional context

_No response_

### Screenshots

Image
Image
Image
Image

### LINKED ISSUE #4684: [ENHANCE] Remove `dead_moons`
### Description

Presumably this attribute can be removed from the cat objects.  `dead_for` is already being found via the status `moons_as`, so it's entirely redundant save information.

amilleah:
## About The Pull Request

Variable `dead_for` was set to `cat["dead_moons"]` which would then be used to update the `group_history[-1]` entry as the sum of all `"dead_moons"` entries, inflating the value with every save+reload of the game.

## Linked Issues

Fixes #4475
Fixes #4684 

## Proof of Testing

Steps to replicate the bug:
1. Convert a StarClan cat to join the Dark Forest (ideally the clan's `guide`/ a cat that has been dead for a while)
```json
"group_history": [
            {
                "group": CatGroup.PLAYER_CLAN_ID,
                "rank": CatRank.WARRIOR,
                "moons_as": 30,
            },
            {
                "group": CatGroup.STARCLAN_ID,
                "rank": CatRank.WARRIOR,
                "moons_as": 10,
            }
        ],
```
2. Moonskip to add to the overall "dead_moons" value
```json
            {
                "group": CatGroup.DARK_FOREST_ID,
                "rank": CatRank.WARRIOR,
                "moons_as": 5,
            },
```
- where total in this example would be
```text
30 moons (in life)
15 moons (in death)
```
3. Save, close, reload game
4. Find that cat's "dead_moons" value to be inflated:
```text
30 moons (in life)
25 moons (in death)
```
where 25 != 15

Attempted to replicate bug again with the same steps and could not; `cat moons (in death)` remained stable on reload.

### DIFF
--- a/scripts/game_structure/load_cat.py
+++ b/scripts/game_structure/load_cat.py
@@ -279,7 +279,8 @@ def json_load():
                     if cat.get("driven_out"):
                         new_cat.status.change_group_nearness(CatGroup.PLAYER_CLAN_ID)
 
-            new_cat.dead_for = cat["dead_moons"]
+            if "dead_moons" in cat and cat["dead_moons"] is not None:
+                new_cat.dead_for = cat["dead_moons"]
             new_cat.experience = cat["experience"]
             new_cat.apprentice = cat["current_apprentice"]
             new_cat.former_apprentices = cat["former_apprentices"]
--- a/scripts/game_structure/load_cat.py
+++ b/scripts/game_structure/load_cat.py
@@ -279,7 +279,7 @@ def json_load():
                     if cat.get("driven_out"):
                         new_cat.status.change_group_nearness(CatGroup.PLAYER_CLAN_ID)
 
-            if "dead_moons" in cat and cat["dead_moons"] is not None:
+            if cat.get("dead_moons"):
                 new_cat.dead_for = cat["dead_moons"]
             new_cat.experience = cat["experience"]
             new_cat.apprentice = cat["current_apprentice"]
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -3404,7 +3404,6 @@ class Cat:
                 "scars": self.pelt.scars or [],
                 "accessory": self.pelt.accessory,
                 "experience": self.experience,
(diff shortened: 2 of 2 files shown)


scribblecrumb:
Out of curiousity, have you checked how this handles save conversions from stable saves?  I'm a little worried that the line you removed was doing something to convert the old status system into the newer one.

amilleah:
would something like handling previous saves look like:
```
if isinstance(cat["status"], str):
    new_cat.dead_for = cat["dead_moons"]
```

(Cat.age_moons being handled like this around line 73?)

scribblecrumb:
Yeah that should handle it just fine

amilleah:
changed the way `dead_moons` is checked for legacy loading but is otherwise removed; [BUGFIX] still holds / was tested again.

scripts/game_structure/load_cat.py
```diff
@@ -279,7 +279,8 @@ def json_load():
                     if cat.get("driven_out"):
                         new_cat.status.change_group_nearness(CatGroup.PLAYER_CLAN_ID)
 
-            new_cat.dead_for = cat["dead_moons"]
+            if "dead_moons" in cat and cat["dead_moons"] is not None:
```
scribblecrumb:
```suggestion
            if cat.get("dead_moons"):
```
should also work!

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #4688, opened 2026-04-01
[FEATURE] Unified input system

larkgz:
## About The Pull Request

* Adds an input system for controllers and keyboard + custom events for input actions.
* Uses the new input system for specifically the cat profile screens as an example (the other ones are TBA)
* There's still some issues (there's a lag spike if starting the game while controllers are plugged in; controller and keyboard remaps aren't persistent yet), but I hope to fix those in separate PRs.

* Separates input actions from implementation
* Allows code reuse between keyboard and controller inputs
* Adds fun controller things like changing the LED colour and rumble

## Proof of Testing

Keyboard:
* Enable keybinds
* Go to cat profile page
* Press left / right / escape to see the keybinds working, even with the new system

Controller:
* Go to cat profile page
* Press left shoulder / right shoulder / Back button to see the controls working with the unified input system


main.py
```diff
@@ -131,9 +135,9 @@ def loading_animation(scale: float = 1):
         i += 1
         if i >= total_frames:
             i = 0
-
-        for event in pygame.event.get():
-            if event.type == pygame.QUIT:
+        for ev in pygame.event.get():
+            controller_manager.process_event(ev)
+            if ev.type == pygame.QUIT:
```
scribblecrumb: can we go back to using `event` instead of `ev`? just for peak readability

scribblecrumb (changes requested):
just a little nitpick

scribblecrumb (approved):

PR #4716, opened 2026-04-05
Fix Missing Absent in scarless_conditions

blackkat312:
## About The Pull Request
Fixes a bug where injuries (lingering shock and head damage) couldn't give absent due to it being missing in scarless_conditions. Found the bug myself when (ironically) trying to stop absent from spawning due to triggers.
(Edit: also added crooked jaw to scarless_conditions!)

## Proof of Testing
I tried 8 times (a ~96% chance of it happening once) to get absent from lingering shock without the fix and it didn't happen. However, with the fix, the first try gave me absent!
image
image

## Changelog/Credits
- Fixes absent being unable to spawn from lingering shock and head damage
- (Edit: also fixes crooked jaw!)
- Discord user is blackkat312 if you wanna add me as a dev!

### DIFF
--- a/scripts/events_module/short/condition_events.py
+++ b/scripts/events_module/short/condition_events.py
@@ -483,6 +483,7 @@ class Condition_Events:
             "persistent headaches",
             "selective mutism",
             "absent",
+            "crooked jaw",
         )
 
         got_condition = False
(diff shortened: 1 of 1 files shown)


MacNCheeseNMac:
Crooked jaw also isn’t in scarless conditions, if you want to kill two birds with one stone

blackkat312:
Ooh yeah you're right!

scribblecrumb (approved):

Issue #4659, opened 2026-03-29
[BUG] weak leg condition event tags unrelated medicine cat


Vickarooni:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

one of the events where a cat gets a weak leg from an injury tags a random medicine cat when clicking the involved cats button even though they are not mentioned in the text.

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. debug ensure an injury that could cause a weak leg
2. moonskip until the permanent condition is triggered
3. check events tab
4. click involved cats button
5. see error

### Commit # or Game Version Number

a8e9c98

### Screenshots

<img width="516" height="128" alt="Image" src="https://github.com/user-attachments/assets/bc3ec85a-7d3d-48a6-9abb-3920f68b6a59" />

Maxxicloudd:
I believe this due to how we handle our event initialization for these specific events

<img width="648" height="110" alt="Image" src="https://github.com/user-attachments/assets/600e5261-a70d-42aa-a5bb-2995942c8b55" />

<img width="632" height="120" alt="Image" src="https://github.com/user-attachments/assets/b096865c-8002-4970-a801-16963bf71e00" />

<img width="811" height="193" alt="Image" src="https://github.com/user-attachments/assets/664ac496-2971-45c4-83fe-c22e6f407d90" />

when the event is instantiated is is done so w/o a cats_involved parameter, so it takes the cat_dicts, which includes a medcat, despite a med cat not being mentioned in the text.

so its an issue specifically with events written w/o a slot for a med cat.

scribblecrumb:
fixed by #4723

PR #4723, opened 2026-04-06
[BUGFIX] Permanent condition events display correct cats involved

Kippanovaa:
## About The Pull Request

- Fixes permanent condition events so the med cat doesn't get tagged if there's not one in the event text
- Changes my name in the credits

## Linked Issues
#4659 

## Proof of Testing
Game_proof

Additional regression check - the events _with_ med cats still tag them
Game_proof2

## Changelog/Credits

- Fixes permanent condition events so the med cat doesn't get tagged if there's not one in the event text


Maxxicloudd:
Actually. This'll cause problems with the way how some events are structured in the dict. Maybe that's why the test fails occasionally and causes the NoneType error.

scripts/events_module/short/condition_events.py
```diff
@@ -828,16 +828,18 @@ def handle_already_injured(cat):
...
 
+                    # Ensure Clan's med cat number aligns with event text
                     if (
                         not med_cat
                         and random_index < 2
                         and len(possible_string_list) >= 3
                     ):
                         random_index = 2
+                    elif random_index < 2:
+                        cat_dict["r_c"] = med_cat
```
scribblecrumb: Instead of us going off of index and assuming specific indices will be appropriate for medicine cats (a confusing system, i hate that i started that in the first place), what we could do instead is:
- determine if we have a med_cat to use (handled already at line `829`)
- introduce a for loop to filter out the possible events.  if we have a `med_cat`, then we essentially "skip" the for loop and assume all events are appropriate. if we *don't* have a `med_cat`, then we use the for loop to toss out any events that have `"r_c"` in them.
- then choose an event out of our remaining possible events.
- lastly, if we have a `med_cat`, we check if `"r_c"` is in the event we've chosen, and, if it is, we include the `med_cat` in the `cat_dict`

scribblecrumb (changes requested):

Maxxicloudd:
Some more test proof
proof

 regression check:
regression_check

and additional checks that the events filter properly:

no_med_check2
no_med_check1

scribblecrumb (approved):
Very nice!

PR #4729, opened 2026-04-07
[BALANCING] Conditions

scribblecrumb:
## About The Pull Request
Intended to help balance conditions, especially perm conditions

- Lowered frequencies of all injury events that give broken bones, based on feedback that cats were breaking bones at silly rates
- Application of herb-lack affects is now chance based instead of guaranteed. Now 70% of the time, if a perm condition does not receive preventative herbs, they will worsen. Based on feedback that permanent condition risks felt prohibitive.  
- The decision to provide preventative herbs to a perm condition or not was previously based on a flat 20 chance threshold. Aka, if a risk had a chance of 20 or higher, then no treatment would be given. However, this meant that some risks could still be allowed far lower than necessarily intended (a risk's default chance might be more like 70, but then be held perpetually at 20 due to the treatment threshold instead of being held at it's default 70).  The change I implemented now makes the threshold the highest number between 20 and the default chance.  So if the default chance is 70, then treatment will be attempted until the chance is 70 or above.  If the default chance is 5, then treatment will be attempted until the chance is 20 or above.

This is all based off feedback given by players.  Complaints are mainly around the frequency of broken bones and their eventual consequences of weak legged cats then leading to cats who are perpetually plagued by their perm condition risks.  Hopefully these changes lead to less broken legs (and thus less perm conditions) as well as better support for cats who have perm conditions.  Players likely won't see immediate change though, as they'll need to let their herb supplies build up and cats receive treatment to pull those risk chances back up.  It'll be a little gradual, but hopefully eventually noticeable.

## Proof of Testing
image

Risk of sore is lower than it's default (which is 20) so we're going to treat it

image

treatment being applied later down the line in code i didn't touch ;)

image

the opposite! chance of sore is 200+, obv higher than our default. so we don't treat

image

instead we apply our lack of herb

image

lack of herbs being applied after no herbs were available and getting through that little 70% chance gateway!

## Changelog/Credits
- Lowered frequencies of all injury events that give broken bones, based on feedback that cats were breaking bones at silly rates
- Application of herb-lack affects is now chance based instead of guaranteed. Now 70% of the time, if a perm condition does not receive preventative herbs, they will worsen.  
- The decision to provide preventative herbs to a perm condition or not was previously based on a flat 20 chance threshold. Aka, if a risk had a chance of 20 or higher, then no treatment would be given. However, this meant that some risks could still be allowed far lower than necessarily intended (a risk's default chance might be more like 70, but then be held perpetually at 20 due to the treatment threshold instead of being held at it's default 70).  The change I implemented now makes the threshold the highest number between 20 and the default chance.  So if the default chance is 70, then treatment will be attempted until the chance is 70 or above.  If the default chance is 5, then treatment will be attempted until the chance is 20 or above.

### DIFF
--- a/scripts/clan_resources/herb/herb_supply.py
+++ b/scripts/clan_resources/herb/herb_supply.py
@@ -1,10 +1,8 @@
-import statistics
-from random import choice, randint, choices
+from random import choice, randint, choices, random
 from typing import Optional
 
 import i18n
 
-from scripts.cat.enums import CatRank
 from scripts.cat.skills import SkillPath
 from scripts.clan_resources.herb.herb import Herb, HERBS
 from scripts.clan_resources.herb.herb_effects import HerbEffect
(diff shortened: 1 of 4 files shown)


scripts/clan_resources/herb/herb_supply.py
```diff
@@ -659,22 +657,31 @@ def _use_herbs(self, treatment_cat, source_dict):
...
 
+            # check if perm condition gets treatment
             if (
                 treatment_cat.is_disabled()
                 and name in treatment_cat.permanent_condition
             ):
-                # if chance of death is already low, med cat doesn't treat
-                if condition.get("mortality") and condition.get("mortality", 0) > 20:
+                condition_default = source_dict[name]
+                will_not_treat = False
```
j-gynn, line 666: this seems to actually be the opposite of what the variable name would suggest

j-gynn: oh no, it's just how weird this section of code is. ignore me.

j-gynn (approved):

PR #4743, opened 2026-04-10
["BUG"FIX] Death Reactions that don't mention m_c

tobes03:
## About The Pull Request

Now that we've split up the prev "major" grief indicators into their own reactions, I went through to make sure every grief event actually mentions the cat being grieved. Along the way, I found a few grammatically awkward or incorrect bits. I also deleted all the blank lists with strings for "empathetic" death reactions and moved existing strings into compassionate.

## Proof of Testing

image

i think this is one of the ones i added an m_c to


resources/lang/en/events/death/death_reactions/general/general_like.json
```diff
@@ -53,7 +53,6 @@
             "r_c is rocked by the loss of m_c. All those memories and shared moments are now burdens that {PRONOUN/r_c/subject} must carry alone.",
             "m_c was so important to r_c. How could the world be so cruel as to take {PRONOUN/m_c/object} away from {PRONOUN/r_c/object}?",
             "Grief swells up to swallow r_c when {PRONOUN/r_c/subject} {VERB/r_c/hear/hears} the news about m_c.",
-            "m_c, dead? But... r_c shakes {PRONOUN/r_c/poss} head, bewildered, insisting they spoke just a sunrise ago.",
```
scribblecrumb, line 56: why'd we remove this one?

tobes03: That one is for no_body deaths, which are common with lost cats - so I figured it was better to just remove it instead of potentially generating it as a reaction to deaths that occurred outside the Clan, where the r_c in question CAN'T have spoken with m_c recently lol.

It's still there as a reaction for body-existing deaths

scribblecrumb (approved):

Issue #4767, opened 2026-04-13
[BUG] Driven off exile still found via prev/next cat buttons


Ssuncreature:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

Exiled and drove off a kit, and noticed they still appear when scrolling through the Cats Outside The Clan Tab. Unsure if its me misunderstanding the drive off option or an actual bug.

Seems to happen randomly if clicking next/previous cat.

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. exile a kit, ensure there is another outsider besides this kit
2. use the drive off option from the leaders den
3. skip a moon
4. open the cats outside the clan tab
5. click on any cat within the tab
6. click either next or previous cat option until driven off kit appears

### Commit # or Game Version Number

5131bbf

### Screenshots

<img width="634" height="102" alt="Image" src="https://github.com/user-attachments/assets/bbfaeb09-6a1f-4cb7-a807-f41e3abcaa36" />

<img width="783" height="310" alt="Image" src="https://github.com/user-attachments/assets/f9cb6db6-112e-4686-8dde-1da2be65c5f5" />

<img width="1560" height="638" alt="Image" src="https://github.com/user-attachments/assets/9c44f625-6818-4684-a048-a892e95eb718" />

scribblecrumb:
Just a matter of excluding driven off cats from the (iirc, hellish) list comprehension used for determining next/prev cats.

yanghede88:
Hi! I noticed this issue is still open. Is anyone actively working on it? If not, I’d be interested to help out.

scribblecrumb:
> Hi! I noticed this issue is still open. Is anyone actively working on it? If not, I’d be interested to help out.

We don't accept AI submissions, no thank you.

PR #4768, opened 2026-04-14
fix: exclude driven-off cats from outsider next/prev navigation

mvanhorn:
## Summary

#4767. Navigating next/prev from a cat in the "Cats Outside The Clan" tab could surface driven-off cats that the tab itself filters out. This change mirrors the tab's filter inside `determine_next_and_previous_cats()`.

## Root cause

`determine_next_and_previous_cats()` in `scripts/cat/cats.py` (line 3415) filtered only by `dead`, `alive_in_player_clan`, and `faded`. The Cats Outside The Clan tab (`ListScreen.get_cotc_cats()`, line 766) additionally restricts to cats that are `is_near(PLAYER_CLAN_ID)`. Driven-off cats have `near=False`, so they were excluded from the tab list but included in the navigation walk.

## Fix

When the current cat is an outsider or other-clan cat AND is still near the player clan, filter `sorted_specific_list` to also require `is_near(PLAYER_CLAN_ID)`. Dead-cat branches and the default group-ID path are untouched. One new branch, 10ish lines.

Matches the maintainer's suggestion on the issue.

## Steps to verify

1. Exile a kit, ensure at least one other outsider is near the clan
2. Drive off the kit via the leader's den
3. Skip a moon
4. Open Cats Outside The Clan tab
5. Tap any cat, hit next/prev repeatedly

Before: driven-off kit eventually appears. After: navigation stays within the visible tab list.

## Testing

Syntax check passes (`python3 -c "import ast; ast.parse(...)"`). Full test suite needs pygame which is not available in this environment; the change is a single list-comprehension filter mirroring an existing path above it.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -3438,6 +3438,18 @@ class Cat:
                 for check_cat in sorted_specific_list
                 if check_cat.status.group_ID == self.status.group_ID
             ]
+        elif (
+            self.status.is_outsider or self.status.is_other_clancat
+        ) and self.status.is_near(CatGroup.PLAYER_CLAN_ID):
+            # Mirror the Cats Outside The Clan tab: only include outsiders who
+            # are still near the player clan. Driven-off cats (near=False) are
+            # filtered out of that tab, so next/prev navigation from a cat in
+            # that tab should not walk into them either (issue #4767).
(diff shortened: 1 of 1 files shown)


j-gynn:
Closing as likely AI-generated.

PR #4774, opened 2026-04-14
[DEVFIX] adjustments to debug behavior

scribblecrumb:
## About The Pull Request
This PR seeks to make the patrol debug more reliable.

- Frequency no longer blocks a debug event from triggering
- If patrol is romantic and the two cats are allowed to be romantic, we bypass the chance roll for a romance patrol. This removes the random factor that could prevent these patrols from appearing reliably.
- Romance patrol debugging would also trigger the "this event wasn't found" print even when they were found due to not being included in the check for debugged patrol in filtered patrols.
- Added some print info about if the debugged patrol is romantic or not

## Proof of Testing
image

Debugged for a romance patrol. Didn't have any trouble getting it to appear!

image

You can see the print output makes sense.

## Changelog/Credits
- Frequency no longer blocks a debug event from triggering
- If patrol is romantic and the two cats are allowed to be romantic, we bypass the chance roll for a romance patrol. This removes the random factor that could prevent these patrols from appearing reliably.
- Romance patrol debugging would also trigger the "this event wasn't found" print even when they were found due to not being included in the check for debugged patrol in filtered patrols.
- Added some print info about if the debugged patrol is romantic or not


scripts/events_module/patrol/patrol.py
```diff
@@ -643,7 +654,10 @@ def _filter_patrols(
         # makes sure that it grabs patrols in the correct biomes, season, with the correct number of cats
         while not filtered_patrols:
             for patrol in possible_patrols:
-                if patrol.frequency != chosen_frequency:
+                if patrol.frequency != chosen_frequency and not (
+                    patrol.patrol_id
+                    == constants.CONFIG["patrol_generation"]["debug_ensure_patrol_id"]
+                ):
```
larkgz:
```suggestion
                if (
                    patrol.frequency != chosen_frequency
                    and patrol.patrol_id
                    != constants.CONFIG["patrol_generation"]["debug_ensure_patrol_id"]
                ):
```

larkgz (approved):

PR #4798, opened 2026-04-16
[DOCS] Herb Dictionary

LaurBee:
## About The Pull Request
_Collaboration between myself, Morgan (@painttyy), and Ion (@MacNCheeseNMac)_
Added a doc containing information on all herbs currently found in-game

> **Note:** I referenced the preexisting docs while making this, so everything ***should*** be in order. That being said, I've never done anything with the docs before, so please let me know if anything is missing

Providing this information will not only help give players an idea of what herbs their Clans need for certain conditions, but it may also help Writing Devs (both current and future) come up with new content to add

[Discord Thread](https://discord.com/channels/1003759225522110524/1491594384238706688/1491594384238706688)

## Proof of Testing
Game opens!
Herb Doc


docs/dev/writing/reference/herbs.md
```diff
@@ -0,0 +1,438 @@
+
+# Herb Dictionary
+###### Contributors: _Laur, Morgan, Ion_
```
scribblecrumb, line 3: This line seems to be breaking the auto-created table of contents, i'd suggest not making it a header and instead just having it as italicized normal text.

docs/dev/writing/reference/herbs.md
```diff
@@ -0,0 +1,438 @@
...
+- Flowers are typically pink or purple in color and funnel-shaped. They are not overly fragrant, though they may emit a faint, minty scent in warm conditions
+- Leaves are green, oval or heart-shaped with scalloped edges, and appear somewhat veiny along the surface. When crushed, they give off a mild, earthy, and slightly bitter scent
+- Total height of plant is typically 1 - 2 feet
+- Quite attractive to bees and other pollinators
+
+### **General Uses:**
+- Increases overall health and strength
+- Helps alleviate breathing issues
+
+###### For more information on how this herb is used in-game, refer to the files located [HERE](https://github.com/ClanGenOfficial/clangen/tree/development/resources/dicts/conditions)
```
scribblecrumb, line 22: Rather than repeating this for every herb, let's put it up at the top as an `!!! tip` box

scribblecrumb (changes requested):
Open to feedback abt this, but I thought it might be good to consider other formatting methods. Atm I'm finding it a bit hard to parse, I think because of how it's formatted. It's not necessarily *bad* formatting, it just ends up feeling a little bit like a wall of text and then my eyes sorta slide off it instead of being able to latch onto the important bits.

I tried using the content grids, which produced this:

Image


This feels easier on the eyes to me. It breaks up the downward scroll a bit. It's easy to pick out the "title" of each section and then follow it to the right for it's information. You can also easily skim right through the bullet points in a smooth manner, but there's a very defined start and stop to them so that the herbs don't all blend together.

This is what it ends up looking like in the markdown:
```
## Burdock

<div class="grid" markdown>

**Description:**

- Plant is a tall-stemmed thistle with dark leaves
- Produces a prickly pink-to-purple flower from July to October
- Roots are dark and difficult for to dig up
- Can reach heights of up to 6 feet
- Gives off a sharp smell and has a bitter taste

**General Uses:**

- Used to soothe and treat wounds, particularly if infected

**Administration Methods:**

- Root is cleaned of soil, then chewed into a pulp and applied topically
</div>
```

docs/dev/writing/reference/herbs.md
LaurBee: Got it 👍 Just to clarify, do you mean the "Herb Dictionary" line, the "Contributors" line, or both?

docs/dev/writing/reference/herbs.md
```diff
@@ -1,438 +1,491 @@
...
+    For more information on how specific herbs are used in-game, refer to the files located [HERE.](https://github.com/ClanGenOfficial/clangen/tree/development/resources/dicts/conditions)
 
 # Herbs
 ## Betony
-### **Description:**
+
+<div class="grid" markdown>
+
+**Description:**
 - Flowers are typically pink or purple in color and funnel-shaped. They are not overly fragrant, though they may emit a faint, minty scent in warm conditions
```
scribblecrumb, line 19: You'll need to have a blank line between each header ("description") and their first bullet point to make it format right. it's annoyingly finicky like that

scribblecrumb (changes requested):

docs/dev/writing/reference/herbs.md
LaurBee: Got it 👍 Fixing now

docs/dev/writing/reference/herbs.md
```diff
@@ -0,0 +1,566 @@
...
+
+Herbs play an important role in ensuring the health of your Clans. Below, we have compiled a list of all the herbs currently found in the ClanGen base game, along with additional information that will hopefully prove useful for Writers, Coders, and Players alike!
+
+!!! caution
+    All medicinal practices mentioned within this article should be treated as works of fiction, and should ***never*** be used to treat cats (or humans) in real life.
+     
+!!! tip
+    For more information on how specific herbs are used in-game, refer to the files located [HERE.](https://github.com/ClanGenOfficial/clangen/tree/development/resources/dicts/conditions)
+
+# Herbs
```
scribblecrumb:
```suggestion
```
Let's just remove this. For some reason it's also breaking the table of contents. I don't think it's really necessary either, since we have no other subsections beyond Herbs that we would want to navigate to.

docs/dev/writing/reference/herbs.md
```diff
@@ -0,0 +1,566 @@
...
+- Helps alleviate breathing issues
+    
+**Administration Methods:**
+
+- Ingested by cat
+</div>
+
+
+## Blackberry
+
```
scribblecrumb, line 37:
```suggestion
</div>

***

## Blackberry

```
Very last thing, I promise! Let's add dividers `***` between each herb block

scribblecrumb (changes requested):

docs/dev/writing/reference/herbs.md
LaurBee: Goodness, those tables really are finicky, aren't they? All done!

docs/dev/writing/reference/herbs.md
LaurBee: > Very last thing, I promise!

I'm holding you to that, scribble! /j
Consider it done!

scribblecrumb (approved):

PR #4810, opened 2026-04-17
[CONTENT] New Cat Meeting Event (Queen Seeking Help)

tobes03:
## About The Pull Request

An event where an outsider queen seeks c_n's help giving birth to their kits. You meet the queen and the kits. The kits have been generating with Clan names no matter what, so I attempted to make that a feature of the event... If that can be solved in the future, then I'll change this.

Fun flavour about meeting outsiders

## Proof of Testing

image

image


resources/lang/en/events/new_cat/general.json
```diff
@@ -2094,5 +2094,46 @@
...
+    "event_text": "n_c:0 seeks m_c's help for giving birth to {PRONOUN/n_c:0/poss} kits. {PRONOUN/n_c:0/subject/CAP}{VERB/n_c:0/'re/'s} so grateful for c_n's aid that {PRONOUN/n_c:0/subject} {VERB/n_c:0/give/gives} {PRONOUN/n_c:0/poss} kittens c_n names.",
+    "m_c": {
+        "status": ["medicine cat"]
+    },
+    "new_cat": [
+        ["can_birth",
+        "meeting",
+        "exists",
+        "old_name"
+        ],
```
scribblecrumb, line 2115: if you add "loner" to this, does it fix that problem with the names?

tobes03: Man I really thought it'd do it, but nope. Still generating with Clan names 

image

scribblecrumb: man what the heck

resources/lang/en/events/new_cat/general.json
```diff
@@ -2095,5 +2095,46 @@
...
+    "relationships": [
+        {
+            "cats_from": ["n_c:0"],
+            "cats_to": ["m_c"],
+            "mutual": false,
+            "values": ["trust", "respect"],
+            "amount": 15,
+            "log": {
+                "cats_from": "cat_from sought cat_to's help when {PRONOUN/cat_from/subject} {VERB/cat_from/were/was} giving birth."
+            }
```
scribblecrumb, line 2133: might i suggest a humble addition of positivity from kits toward m_c? could be logged as "cat_from was told tales of how cat_to aided in their birth." probably a pretty small amount, i just think it's a fun bit of history to preserve there

scribblecrumb (changes requested):

resources/lang/en/events/new_cat/general.json
tobes03: oh that's so cute love. I gotta run to work but I'll add it in when I get a moment!

resources/lang/en/events/new_cat/general.json
```diff
@@ -2095,5 +2095,57 @@
         "current_rep": ["hostile"],
         "changed": 3
     }
+},
+{
+    "event_id": "gen_new_cat_queenseekinghelp",
+    "location": ["any"],
+    "season": ["any"],
+    "tags": ["clan:medicine cat"],
```
scribblecrumb: don't need this btw since we already constrain m_c to be a medicine cat. the `clan:rank` tags are only really necessary if we mention a rank in the text without constraining some cat to *be* that rank.

scribblecrumb (changes requested):
just one other thing

resources/lang/en/events/new_cat/general.json
tobes03: ahh yeah, I think I originally wrote the event not to mention r_c, then decided it would be better to actually call out a medicine cat for relationship stuff. all fixed c:

scribblecrumb (approved):

PR #4823, opened 2026-04-19
prevent driven off cats from being found via prev/next cat buttons

### LINKED ISSUE #4767: [BUG] Driven off exile still found via prev/next cat buttons

rustykitty:
## About The Pull Request
This pull request fixes the bug where driven off cats can still be found accessed using the previous cat and next cat buttons. This is achieved by making sure that alive cats are "near" to the Clan.

## Linked Issues
Fixes #4767

## Proof of Testing
https://github.com/user-attachments/assets/ce6357b8-4a7f-4c90-b8b9-2d394f1a7fee

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -3438,6 +3438,14 @@ class Cat:
                 for check_cat in sorted_specific_list
                 if check_cat.status.group_ID == self.status.group_ID
             ]
+        # driven off cats
+        else:
+            sorted_specific_list = [
+                check_cat
+                for check_cat in sorted_specific_list
+                # pretty sure a cat is considered "near" their own group
+                if check_cat.status.is_near(CatGroup.PLAYER_CLAN_ID)
+            ]
 
         if filter_func is not None:
             sorted_specific_list = [
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -3443,7 +3443,6 @@ class Cat:
             sorted_specific_list = [
                 check_cat
                 for check_cat in sorted_specific_list
-                # pretty sure a cat is considered "near" their own group
                 if check_cat.status.is_near(CatGroup.PLAYER_CLAN_ID)
             ]
 


scripts/cat/cats.py
```diff
@@ -3438,6 +3438,13 @@ def determine_next_and_previous_cats(
                 for check_cat in sorted_specific_list
                 if check_cat.status.group_ID == self.status.group_ID
             ]
+        # driven off cats
+        else:
+            sorted_specific_list = [
+                check_cat
+                for check_cat in sorted_specific_list
+                if check_cat.status.is_near(CatGroup.PLAYER_CLAN_ID)
+            ]
```
scribblecrumb, line 3447: Have you tested how this handles *dead* driven off cats?

rustykitty: Clarify what is meant by "dead driven off cats" here? As in driven off cats that have died? I would have assumed that they cannot die because they are no longer accessible, and thus the Clan probably wouldn't hear about their death; also you can't drive off a dead cat, can you...?

scribblecrumb: I do mean a driven off cat that has died.  While they aren't accessible, I'm pretty sure they *do* still die as typical for an outsider. The events just aren't shown to the player. I worry that your current implementation will result in a driven off cat dying, then still being accessible via afterlife prev/next cats since they'll be handled by the `if self.dead` before the `else` that handles driven off cats.

scribblecrumb:
Rusty has indicated that he doesn't have the time/interest in addressing the changes for this PR, so I'll be closing it!

PR #4822, opened 2026-04-19
[CONTENT] guard asleep

fulminousnight:
## About The Pull Request

- Added a simple event of a warrior falling asleep on nightly guard duty. Room to expand on the concept with several variations.

## Why This Is Good For ClanGen

- More diverse events = more interesting gameplay. 

## Proof of Testing

Screenshot 2026-04-19 at 10 11 25 AM

PR #4874, opened 2026-04-26
[CONTENT] Guard Duty Events

Vickarooni:
## About The Pull Request

Adds 3 events relating to guard duty, building off of #4822 . One is a miscellaneous event where a cat catches a Clanmate trying to sneak out of camp. The other two are variations of the same death where a cat doesn't return from guard duty and is later found dead - either due to a rival Clan or an outsider.

event variety!! #4822 expands on the concept of cats going on guard duty so let's make stuff based off of that!

## Proof of Testing

gen_misc_guardsneakout
image
image
<br>
gen_death_guarddisappearance_any1
image
image
<br>
gen_death_guarddisappearance_any2
image
image

## Changelog/Credits

- Adds three new events relating to guard duty! One miscellaneous, two deaths. Be careful of what's out there...

### DIFF
--- a/resources/lang/en/events/death/general.json
+++ b/resources/lang/en/events/death/general.json
@@ -8783,6 +8783,30 @@
       ],
       "death": "m_c was killed by a rogue while on guard duty."
     }
-  ]
+  ],
+        "relationships": [
+            {
+                "cats_from": ["clan"],
+                "cats_to": [
+                    "n_c:0"
+                ],
+                "mutual": false,
+                "values": [
+                    "like",
+	        "comfort",
+	        "trust"
+                ],
+                "amount": -15,
+                "log": {
+                    "cats_from": "cat_from felt outraged after hearing that cat_to killed m_c while on guard duty."
+                }
+            }
+        ],
+  "outsider": {
+    "current_rep": [
+      "any"
+    ],
+    "changed": -3
+  }
 }
 ]
--- a/resources/lang/en/events/death/general.json
+++ b/resources/lang/en/events/death/general.json
@@ -8711,5 +8711,78 @@
                 "death": "m_c was killed by a rooster defending his flock."
             }
         ]
+    },
+{
+  "event_id": "gen_death_guarddisappearance_any1",
+  "location": [
+    "any"
+  ],
+  "season": [
+    "any"
+  ],
+  "tags": [],
+  "frequency": 2,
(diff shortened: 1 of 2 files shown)


resources/lang/en/events/death/general.json
```diff
@@ -8711,5 +8711,78 @@
...
+    {
+      "cats": [
+        "m_c"
+      ],
+      "death": "m_c was killed by o_c_n warriors while on guard duty."
+    }
+  ],
+  "other_clan": {
+    "current_rep": [
+      "any"
```
tobes03:
```suggestion
      "hostile", "neutral"
```

I think an allied Clan coming in to randomly murder your guard would be a bit extreme and confusing. I'm not opposed to the drama of having a previously-allied Clan do something like that, but I would want to see a variant on this event where the text acknowledges how shocking it is that their ally did this.

resources/lang/en/events/death/general.json
```diff
@@ -8711,5 +8711,78 @@
...
+  "event_id": "gen_death_guarddisappearance_any2",
+  "location": [
+    "any"
+  ],
+  "season": [
+    "any"
+  ],
+  "tags": [],
+  "frequency": 2,
+  "event_text": "The Clan grows concerned after m_c doesn't return from guard duty. {PRONOUN/m_c/poss/CAP} body is later found out in the territory with a familiar scent reeking off {PRONOUN/m_c/object}.",
```
tobes03:
```suggestion
  "event_text": "The Clan grows concerned after m_c doesn't return from guard duty. {PRONOUN/m_c/poss/CAP} body is later found out in the territory with a rogue's scent wreathing around {PRONOUN/m_c/object}.",
```

This text change is because of how the "exists" tag currently works -- it's completely possible that the game will not be able to find a rogue in the player's "Cats outside the Clan" list, so it will have to generate a new one. In that case, it won't be a "familiar" scent. Scribble was talking about modifying how this code works, so once those changes are implemented, the event can go back to your intended version.

resources/lang/en/events/death/general.json
```diff
@@ -8711,5 +8711,78 @@
...
+  ],
+  "other_clan": {
+    "current_rep": [
+      "any"
+    ],
+    "changed": -3
+  }
+},
+{
+  "event_id": "gen_death_guarddisappearance_any2",
```
tobes03, line 8752: This event could also incur a negative reaction from the Clan toward the rogue!

It should also probably worsen the Clan's relationship with outsiders.

tobes03 (changes requested):

tobes03 (approved):

PR #4885, opened 2026-04-27
[DOCS] expanded on EXP documentation

fulminousnight:
## About The Pull Request

- Adds more documentation on the amount of EXP granted on various types of patrols from failures to average to extreme scenarios. 

- Very helpful for writers to have on hand for balancing patrols.

## Proof of Testing

- ClanGen does not crash / the formatting for docs is not messed up by it. 

Screenshot 2026-04-27 at 4 01 53 PM


tobes03:
So I know it feels counter-intuitive for training patrols to have the lowest reward, but the types' EXP rewards are supposed to go border -> hunting -> training, from most to least. You should flip the values in the training/border rows.

https://clangen.io/docs/dev/writing/patrols/#types-liststr

Here's our current documentation on the types of patrols. The rationale is that border patrols are high risk and high reward, whereas training patrols give a lower EXP quantity in exchange for safety.

PR #4919, opened 2026-04-30
[BUGFIX] apprentice in leader den screen

### LINKED ISSUE #4877: [UI] Apprentice in leader den screen
### Pre-requisites

### Description

Encountered while testing some new events, leader and deputy injured + no medicine cat. An apprentice's sprite appears in the leader den screen rather than any of the available warriors

Stops happening if you reopen the leaders den tab

### Grade

1. Have injured leader and deputy
2. Have no medicine cat, and at least one apprentice in the clan
3. Open the leader den, and theres a chance of the apprentice appearing

_No response_

 b43fdaa

No

_No response_

_No response_

### Additional context

"my evil apprentice whispering in my ear" - bellamy

Probably just a minor flaw in the cat selector for when leader is unavailable
(note: eaglespots is a warrior)

### Screenshots

Image

Image

Inkpawwc:
## About The Pull Request

Excludes apprentices from being selected as an "adult" to appear in the leader's den when getting a random adult cat is necessary.

## Linked Issues

#4877

## Proof of Testing

image
image
image
image
image
image
image

## Changelog/Credits

UI bugfix

### DIFF
--- a/scripts/screens/LeaderDenScreen.py
+++ b/scripts/screens/LeaderDenScreen.py
@@ -224,7 +224,7 @@ class LeaderDenScreen(Screens):
                     for i in Cat.all_cats.values()
                     if i.status.alive_in_player_clan
                     and i.status.rank
-                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.LEADER]
+                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.is_any_apprentice_rank(), CatRank.LEADER]
                 ]
                 if adults:
                     self.helper_cat = random.choice(adults)
--- a/scripts/screens/LeaderDenScreen.py
+++ b/scripts/screens/LeaderDenScreen.py
@@ -224,7 +224,7 @@ class LeaderDenScreen(Screens):
                     for i in Cat.all_cats.values()
                     if i.status.alive_in_player_clan
                     and i.status.rank
-                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.is_any_apprentice_rank(), CatRank.LEADER]
+                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.APPRENTICE, CatRank.MEDICINE_APPRENTICE, CatRank.MEDIATOR_APPRENTICE, CatRank.LEADER]
                 ]
                 if adults:
                     self.helper_cat = random.choice(adults)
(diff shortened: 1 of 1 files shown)


scripts/screens/LeaderDenScreen.py
```diff
@@ -224,7 +224,7 @@ def screen_switches(self):
                     for i in Cat.all_cats.values()
                     if i.status.alive_in_player_clan
                     and i.status.rank
-                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.LEADER]
```
scribblecrumb, line 227: Point of clarification: Apps aren't actually meant to be excluded! Kits/newborns are the only things we don't want here.

However, I definitely see the issue with apps managing to take precedence over warriors at times. A better method here might be to use the available cat with the highest exp. That should naturally prefer warriors and apps should only get in if no "older" ranks are available.

Something like this:
```
adults = sorted(adults, key=lambda x: x.experience, reverse=True)
```
should work

scribblecrumb (changes requested):

scripts/screens/LeaderDenScreen.py
Inkpawwc: should it still randomly select from a few different highest exp cats or should it just only return the same highest exp cat each time as long as they're still available?

scribblecrumb: I think the highest available cat would be best! I like the consistency of that, and it makes the most sense. If none of the normally leader-y ranks are available, then of course the most experienced cat is chosen to represent!

Inkpawwc: Great! That makes things very simple to add in then too lol

scripts/screens/LeaderDenScreen.py
```diff
@@ -224,10 +224,12 @@ def screen_switches(self):
                     for i in Cat.all_cats.values()
                     if i.status.alive_in_player_clan
                     and i.status.rank
-                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.APPRENTICE, CatRank.MEDICINE_APPRENTICE, CatRank.MEDIATOR_APPRENTICE, CatRank.LEADER]
+                    not in [CatRank.NEWBORN, CatRank.KITTEN, CatRank.LEADER]
                 ]
```
scribblecrumb, line 229: One more thing! Let's add `and not i.not_working()` to this list comprehension to ensure we don't end up with anyone who's sick/injured

scribblecrumb (changes requested):

scripts/screens/LeaderDenScreen.py
Inkpawwc: oop yeah good point, wasn't thinking about that!

PR #4920, opened 2026-04-30
[EDITS] tall tree death clarification

Vickarooni:
## About The Pull Request

very minor edit to the "fell from the top branches of a tall tree" death event to clarify that it's fatal.

for clarity's sake! i always had issues remembering this event was supposed to be a death due to the event's brevity and sort of vague wording. this makes sure the player is aware of the fact it kills the cat involved.

## Proof of Testing

image

## Changelog/Credits

- Added an extra sentence to a death event to clarify that it is, in fact, a death.

### DIFF
--- a/resources/lang/en/events/death/general.json
+++ b/resources/lang/en/events/death/general.json
@@ -5309,7 +5309,7 @@
         ],
         "tags": [],
         "frequency": 4,
-        "event_text": "m_c fell from the top branches of a tree.",
+        "event_text": "m_c fell from the top branches of a tree. {PRONOUN/m_c/subject/CAP} never get up again.",
         "m_c": {
             "age": [
                 "-kitten"
--- a/resources/lang/en/events/death/general.json
+++ b/resources/lang/en/events/death/general.json
@@ -5309,7 +5309,7 @@
         ],
         "tags": [],
         "frequency": 4,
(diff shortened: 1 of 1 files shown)


resources/lang/en/events/death/general.json
```diff
@@ -5309,7 +5309,7 @@
         ],
         "tags": [],
         "frequency": 4,
-        "event_text": "m_c fell from the top branches of a tree.",
+        "event_text": "m_c fell from the top branches of a tree. {PRONOUN/m_c/subject/CAP} never {VERB/m_c/get/gets} up again.",
```
tobes03, line 5312: I know this is a crazy nitpick, even for me, but I want to point out that this can generate for leaders, so... technically... they would get back up.

That being said, I like the flavour of the text, so my suggestion is to exclude leaders from this death and then (for bonus points) add another version of it that targets *only* leaders and acknowledges it in the text

tobes03 (changes requested):

resources/lang/en/events/death/general.json
Vickarooni: edit: disregard what i said earlier im actually gonna do this

tobes03 (approved):

PR #4965, opened 2026-05-04
[CRASHFIX] exclusionary values on patrols

### LINKED ISSUE #4962: [CRASH] -SPEAKER is not a real skill path
### Pre-requisites

### Description

The game crashes on a patrol outcome because "-SPEAKER is not a real skill path".

It makes me think that the exclusionary tags on the skills in patrols need something extra to prevent this

### Grade

Game-breaking - makes it impossible to play the game

1. Patrol with two cats (non arrogant, non confident, non speakers)
2. proceed

gen_train_speakingromance1

4ef20d05

### Error log

```clangen\scripts\cat\skills.py", line 622, in meets_skill_requirement
    raise KeyError(f"{path} is not a real skill path")
KeyError: '-SPEAKER is not a real skill path'
```

scribblecrumb:
## About The Pull Request
let me just casually make 9 lines into 71. i feel insane and this code should really see a refactor instead of what i've done. but what i've done is the most feasible without a much larger rewrite.

The problem was that the success chance can be modified based on skills and traits being present on the patrol. However we weren't accounting for exclusionary values.  I've accounted for them now and have ensured they have at least nominally the same sort of effect as their inclusionary counterparts.

## Linked Issues
fixes #4962

## Proof of Testing
image

successfully got to the outcome which had been crashing prior

image

you can also see that the cats updated the chance of success as they would have if these were inclusionary instead of exclusionary values

## Changelog/Credits
- exclusionary values no longer crash patrols

### DIFF
--- a/scripts/events_module/patrol/patrol.py
+++ b/scripts/events_module/patrol/patrol.py
@@ -1000,12 +1000,12 @@ class Patrol:
 
             if is_exclusionary and not hits:
                 # if they don't have a disallowed skill, we decrease chance
-                success_chance -= (
+                success_chance += (
                     1 * constants.CONFIG["patrol_generation"]["fail_stat_cat_modifier"]
                 )
             else:
                 # if they had the required skill, we decrease chance
-                success_chance -= (
+                success_chance += (
                     hits
                     * constants.CONFIG["patrol_generation"]["fail_stat_cat_modifier"]
                 )
@@ -1040,7 +1040,7 @@ class Patrol:
             if (is_exclusionary and kitty.personality.trait not in trait_to_check) or (
                 kitty.personality.trait in trait_to_check
             ):
-                success_chance -= constants.CONFIG["patrol_generation"][
+                success_chance += constants.CONFIG["patrol_generation"][
                     "fail_stat_cat_modifier"
                 ]
 
--- a/scripts/events_module/patrol/patrol.py
+++ b/scripts/events_module/patrol/patrol.py
@@ -24,6 +24,7 @@ from scripts.events_module.event_filters import (
     get_personality_compatibility,
     event_for_location,
     event_for_season,
+    cat_for_event,
 )
 from scripts.events_module.patrol.patrol_event import PatrolEvent
 from scripts.events_module.patrol.patrol_outcome import PatrolOutcome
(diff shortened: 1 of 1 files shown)


scripts/events_module/patrol/patrol.py
```diff
@@ -962,23 +963,84 @@ def calculate_success(
...
+            else:
+                skills_to_check = success_outcome.stat_skill
+
+            hits = kitty.skills.check_skill_requirement_list(skills_to_check)
+
+            if is_exclusionary and not hits:
+                # if they had a disallowed skill, we reduce chance
+                success_chance += (
+                    1 * constants.CONFIG["patrol_generation"]["win_stat_cat_modifier"]
+                )
```
larkgz, line 983:
```suggestion
                success_chance -= (
                    1 * constants.CONFIG["patrol_generation"]["win_stat_cat_modifier"]
                )
```
So that the chance is reduced.

scripts/events_module/patrol/patrol.py
```diff
@@ -962,23 +963,84 @@ def calculate_success(
...
+            if is_exclusionary:
+                skills_to_check = [
+                    x.replace("-", "") for x in success_outcome.stat_skill
+                ]
+            else:
+                skills_to_check = success_outcome.stat_skill
+
+            hits = kitty.skills.check_skill_requirement_list(skills_to_check)
+
+            if is_exclusionary and not hits:
```
larkgz, line 979: I'm a little unsure about these cases. If `skills_to_check` in the `is_exclusionary` case contains the skills that we want to disallow, then I think we'd want to reduce chances if we *do* find them in the skill requirements list?

scripts/events_module/patrol/patrol.py
```diff
@@ -962,23 +963,84 @@ def calculate_success(
...
+            is_exclusionary = any(
+                value.find("-") == 0 for value in fail_outcome.stat_skill
+            )
+            if is_exclusionary:
+                skills_to_check = [x.replace("-", "") for x in fail_outcome.stat_skill]
+            else:
+                skills_to_check = fail_outcome.stat_skill
+            hits = kitty.skills.check_skill_requirement_list(skills_to_check)
+
+            if is_exclusionary and not hits:
```
larkgz, line 1001: See previous comment.

scripts/events_module/patrol/patrol.py
```diff
@@ -962,23 +963,84 @@ def calculate_success(
...
+                # if they had the disallowed skill, we increase chance
+                success_chance -= (
+                    1 * constants.CONFIG["patrol_generation"]["fail_stat_cat_modifier"]
+                )
+            else:
+                # if they had the required skill, we decrease chance
+                success_chance -= (
+                    hits
+                    * constants.CONFIG["patrol_generation"]["fail_stat_cat_modifier"]
+                )
```
larkgz:
```suggestion
                success_chance += (
                    hits
                    * constants.CONFIG["patrol_generation"]["fail_stat_cat_modifier"]
                )
```
So that the chance is decreased (`fail_stat_cat_modifier` is a negative value).

larkgz (changes requested):

scripts/events_module/patrol/patrol.py
scribblecrumb: I felt it better that we follow the current convention, which is "if we have the required skills, then the success chance increases" *not* "if we don't have the required skills, the success chance decreases". In that sense, if the cat doesn't have the *disallowed* skills, then that means they *do* have a *required* skill, and thus the success chance increases.

scripts/events_module/patrol/patrol.py
scribblecrumb: Not the intention. I realize my comment is incorrect here, twas written when i was trying to fix this a different way. i'll update it.

larkgz (approved):

PR #5014, opened 2026-05-10
[FEATURE] Keybind Matrix Maps

scribblecrumb:
## About The Pull Request
This PR adds element mapping to our screens to allow us to automatically utilize arrow keybinds for navigation.

Some notes:
- This is only set up for the Start Screen at current, as a "proof of concept"
- I figure we can slowly expand to more and more screens over time. Switching a screen to this functionality is not particularly difficult, given a screen is reasonably coded (looking at you, Switch Clan screen >.>) For screens that *aren't* reasonably coded, we'll likely need to do some refactoring.
- I have already written documentation explaining the basic method of switching a screen to the new system, so that devs can jump in with a ready-made guide.
- Per dev discussion, the keybind setting has been killed and keybinds are available as default.

Now, for how this all works:
- We have a new script that holds all the matrix functions
- Screens access this script via the base `Screens` class func `update_map` in order to create and modify the matrix map.
- This matrix map is, essentially, a map of the interactable elements on the screen which can be read by the code. It's automatically compiled based on the positional coords of the elements given. 
- The Screens class also has event handling set up for arrow key presses which which move the "focus" onto a new element. This includes some smart wrapping logic. For example, if a player presses the UP key, but no element is above the currently focused element, then it'll move to the lowest possible element. If a player presses the RIGHT key, but no element is to the right, it'll move to the left-most element of the next row down.
- We keep track of the focused element via a `Screens` attribute: `self.current_focus` so that we can utilize it when the ENTER key is pressed.
- I've integrated the new unified input system into this, which means that controllers also work with this!
- Likewise, I've set up our two most-used button elements to properly process the new input events into the pygame_gui events so that we don't actually need to adjust the code for event handling in each screen.

## Proof of Testing

Uploading 2026-05-21 17-46-10.mp4…

works on the Start screen!

## Changelog/Credits
- Adds code functionality for adding robust keybind navigation to our screens
- Currently only functional for the Start Screen
- Adds documentation regarding how to integrate keybind navigation onto screens, for use by other devs as we being integrating this into more and more screens.


scripts/game_input/input_manager.py
larkgz, line 1: Since we're removing the `game_setting_get("keybinds")` call, can the `game_setting_get` import be removed as well?

scripts/game_structure/audio/sound.py
```diff
@@ -69,6 +70,13 @@ def handle_sound_events(self, event):
                 if self.pressed != event.ui_element:
                     self.play("button_hover")
             self.pressed = None
+        elif event.type == INPUT_ACTION_PRESSED and event.action in (
+            Action.LEFT,
+            Action.RIGHT,
+            Action.DOWN,
+            Action.UP,
+        ):
+            self.play("button_hover")
```
larkgz: Not a big deal, but would it be possible to do this based on an event specific to buttons rather than directly on `Action` input? Just so that it doesn't play if the button doesn't change for whatever reason.

scripts/ui/elements/image_button.py
```diff
@@ -209,3 +210,45 @@ def check_hover(self, time_delta: float, hovered_higher_element: bool) -> bool:
...
+        self.drawable_shape.set_active_state("normal")
+        if self.tool_tip is not None:
+            self.tool_tip.kill()
+            self.tool_tip = None
+        super().unfocus()
+
+    def on_unhovered(self):
+        super().on_unhovered()
+        if self.is_focused:
+            self.drawable_shape.set_active_state("selected")
```
larkgz: Can you add a comment clarifying why `set_active_state` is set to `selected`?

scripts/screens/Screens.py
```diff
@@ -212,32 +221,94 @@ def screen_switches(self):
...
+        # reset matrix map so that it can be recreated cleanly later
+        self.matrix_map.clear()
         pass
 
+    def update_map(self, element_list: list[UIElement], remove=False):
+        """
+        Updates the matrix map with the given list of elements.
+        :param element_list: The list of elements to update.
+        :param remove: Default False, set to True if the list of elements should be removed from the matrix map instead of added.
+        """
```
larkgz: Why one function for adding and removing from the matrix map instead of separate ones?

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+from typing import Optional
+
+from pygame_gui.core import UIElement
+
+from scripts.game_input import Action
+from scripts.game_structure.game import switch_get_value, Switch
+from scripts.ui.scale import ui_scale_value
+
+
+def add_to_map(current_map: list[list], new_elements: list[UIElement]) -> list[list]:
```
larkgz:
```suggestion
def add_to_map(current_map: list[list[UIElement]], new_elements: list[UIElement]) -> list[list]:
```

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    Takes the given elements and adds them to the map
+    :param current_map: The current matrix map
```
larkgz, line 18: Can these docstrings explain how the matrix map is structured?

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+    :param new_elements: The list of interactable elements to add
+    :return: The new map
```
larkgz, line 20: Even though this function is returning the new map, it's also modifying the provided `current_map` due to Python lists being mutable. This could be confusing, so I'd prefer if it was one or the other.

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    for element in new_elements:
+        # first, find the current y positions represented
+        current_rows: dict[int, list] = {}
```
larkgz: Can this be renamed to be more clear at a glance? Something like `rows_by_ypos` maybe?

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+        for row in current_rows:
+            # we allow a 30 px range so that elements which are slightly different y coordinates
+            # but still visibly feel side-by-side will be treated as part of the same row
+            if (
+                (row - ui_scale_value(15))
+                <= element.get_abs_rect().y
+                <= row + ui_scale_value(15)
+            ):
+                new_row = False
+                target = row
```
larkgz: Can probably break the loop here.

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+def find_next_focus(
+    current_map: list[list], direction: Action, last_element: UIElement
+) -> UIElement:
+    """
+    Moves focus from one element to the next logical element.
+    :param current_map: The current matrix map
+    :param direction: The direction in which to look for the next element
+    :param last_element: The element currently in focus
```
larkgz: Can `last_element` be renamed to `prev_focus` (or `prev_focus_element` if you want to make it really clear)?

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+                change_to_lower_row = True
+
+    # going UP!
+    if direction == Action.UP or change_to_higher_row:
+        while not _valid_row(current_map, last_element, new_row):
+            # find the new row, wrapping if necessary
+            if prior_row - 1 >= 0:
+                new_row = prior_row - 1
+            else:
+                new_row = len(current_map) - 1
```
larkgz, line 151: You can also wrap rows with `new_row = prior_row - 1 % len(current_map)`

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+
+        if already_present:
+            continue
+
+        # then position of the new element
+        position = element.get_abs_rect()
+
+        # first we check if a new row is needed
+        new_row = True
+        target = None
```
larkgz: Can this be renamed to `target_row`?

scripts/ui/focus_matrix.py
```diff
@@ -0,0 +1,257 @@
...
+    :param current_map: The current matrix map
+    :param elements_to_remove: The list of interactable elements to remove
+    :return: The new map
+    """
+    for element in elements_to_remove:
+        # first find where the element is positioned
+        element_row: Optional[int] = None
+        for row in current_map:
+            if element in row:
+                element_row = current_map.index(row)
```
larkgz, line 86: Can break from loop I think.

larkgz (commented):
Wasn't sure where to put this, but there seems to be an issue where mousing over the focused element will result in the button flickering without being highlighted.

scripts/game_structure/audio/sound.py
scribblecrumb: I did try to get the button elements to post their ON_HOVERED events in the same manner we're translating the INPUT_ACTION_PRESSED, but it wasn't working. It would trigger all the other hover effects as well, resulting in really weird visual behavior. 
This was the best solution I could come up with to get the sounds to still work, though it's definitely hacky.
It's possible we could set up some kind of custom event for the buttons to post? That's the only other idea I have for a different solution here.

scripts/screens/Screens.py
scribblecrumb: I preferred the idea of only a single function (other than `set_focus`) having to be used when developing screens. I could separate it though, if you feel that would be more prudent.

scripts/ui/elements/image_button.py
scribblecrumb: Ah we actually don't need to do this anymore, tis old code from when I didn't want to reuse the hover state image. I'll take this out.

scripts/ui/focus_matrix.py
scribblecrumb: Maybe I'm just misunderstanding, but that doesn't appear to work? In testing when `prior_row` is 0 and `len(current_map)` is 6 that's `0 - 1 % 6` which gives -1 as the new row.

larkgz: Ah, I got my modulo division confused. It only works the way I said if you’re going up, so you could probably use it for the other one. If you’re going down, it would be: `new_row = ((prior_row - 1) + len(current_map)) % len(current_map)`, but that's kind of harder to read...

If you keep it the way it is, I think flipping the order so it starts with `if prior_row - 1 < 0:` is easier to understand, since you’re already subtracting.

scripts/game_structure/audio/sound.py
larkgz: Yeah, I think custom event would maybe be the way to go... But we could probably save that for later. For now, maybe just disable the button sound effects and put in a `TODO` comment, since otherwise every time someone presses an arrow on any page, it will play the sound effects.

scripts/screens/Screens.py
larkgz: I would prefer for them to be separated, for readability.

larkgz (commented):
See comments.

larkgz (approved):

PR #5094, opened 2026-05-15
[TASK] Handle Card Save Data

### LINKED ISSUE #5082: [PROJECT] Handle Card Save Data
**Description**
For the cruel season update! We'll need to save the clan's selected cards somewhere and make sure they can be referenced in the code.

We could add a `cruel_cards` attribute to the `Clan` object. This would hold a list of card ID strings for the cards chosen by the player. This info should then, likewise, be saved within the clan json and retrieved on load.

**Scope of Project**
If you haven't handled our save data before, this might be a good way to learn! It should be a relatively small addition to the file and won't require you to set up file read or writes, since this is info that can reasonably be added to the existing clan json file.

**Relevant Links**

Inkpawwc:
## About The Pull Request

added cruel_cards attribute to clan.py & handles storing card IDs as clan_data

Fixes: #5082

### DIFF
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -72,7 +72,7 @@ class Clan:
         camp_bg=None,
         symbol=None,
         game_mode="classic",
-        cruel_cards=[],
+        cruel_cards: list[str]=[],
         starting_members=None,
         starting_season="Newleaf",
         self_run_init_functions=True,
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -761,7 +761,7 @@ class Clan:
             biome=clan_data["biome"],
             camp_bg=clan_data["camp_bg"],
             game_mode=clan_data["gamemode"],
-            cruel_cards=clan_data["cruel_cards"],
+            cruel_cards=clan_data.get("cruel_cards", []),
             self_run_init_functions=False,
         )
         game.clan.post_initialization_functions()
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -118,7 +118,7 @@ class Clan:
         self.camp_bg = camp_bg
         self.chosen_symbol = symbol
         self.game_mode = game_mode
-        self.cruel_cards = cruel_cards
+        self.cruel_cards: list[str] = cruel_cards
         self.pregnancy_data = {}
         self.inheritance = {}
         self.custom_pronouns = {}
--- a/scripts/clan.py
+++ b/scripts/clan.py
@@ -72,6 +72,7 @@ class Clan:
         camp_bg=None,
         symbol=None,
         game_mode="classic",
+        cruel_cards=[],
         starting_members=None,
         starting_season="Newleaf",
         self_run_init_functions=True,
@@ -117,10 +118,7 @@ class Clan:
         self.camp_bg = camp_bg
         self.chosen_symbol = symbol
         self.game_mode = game_mode
-        if game_mode == "cruel season":
-            self.cruel_cards = []
-        else:
-            self.cruel_cards = None
+        self.cruel_cards = cruel_cards
         self.pregnancy_data = {}
         self.inheritance = {}
         self.custom_pronouns = {}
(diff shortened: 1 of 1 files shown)


scripts/clan.py
```diff
@@ -117,6 +117,10 @@ def __init__(
         self.camp_bg = camp_bg
         self.chosen_symbol = symbol
         self.game_mode = game_mode
+        if game_mode == "cruel season":
+            self.cruel_cards = []
+        else:
+            self.cruel_cards = None
```
scribblecrumb: So in this section we're actually going to need a `cruel_cards` argument for the initialization.  We initialize a clan object when the player finishes making a Clan! At this point, they will have picked cruel cards and we'll need to be passing that list of chosen cards into the clan object.

This means that you'll want to add a `cruel_cards` argument. You can have it default to an empty list: `[]` and then just set `self.cruel_cards = cruel_cards` in the same manner that we handle the `self.game_mode` in the line above.

scripts/clan.py
```diff
@@ -437,6 +441,7 @@ def save_clan(self):
             "version_commit": get_version_info().version_number,
             "source_build": get_version_info().is_source_build,
             "custom_pronouns": self.custom_pronouns,
+            "cruel_card_IDs": self.cruel_cards,
```
scribblecrumb: Let's have the save param also be `"cruel_cards"` just to keep naming consistent

scribblecrumb (changes requested):
A couple suggestions!

You also need to handle loading the information from the save. Loading is done in `load_clan_json`.

scripts/clan.py
```diff
@@ -486,6 +491,11 @@ def save_clan(self):
         if game.clan.game_mode in ("expanded", "cruel season"):
             self.save_freshkill_pile(game.clan)
 
+        if game.clan.game_mode == "cruel season":
+            clan_data["cruel_card_IDs"] = self.cruel_cards
+        else:
+            clan_data["cruel_card_IDs"] = {}
+
```
scribblecrumb: Oh also, sorry I missed this bit. You don't need this section. You've already saved the cruel_cards in the above part.

scripts/clan.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         camp_bg=None,
         symbol=None,
         game_mode="classic",
+        cruel_cards=[],
```
scribblecrumb:
```suggestion
        cruel_cards: list[str]=[],
```
just some typehinting! this ensures that coders will get a warning if they try to pass something that isn't a string list. i know we don't do it for the other arguments, that's just our classic good ole technical debt. we might as well do it for this!

scripts/clan.py
```diff
@@ -117,6 +118,7 @@ def __init__(
         self.camp_bg = camp_bg
         self.chosen_symbol = symbol
         self.game_mode = game_mode
+        self.cruel_cards = cruel_cards
```
scribblecrumb:
```suggestion
        self.cruel_cards: list[str] = cruel_cards
```
same deal here!

scripts/clan.py
```diff
@@ -758,6 +761,7 @@ def load_clan_json(self):
             biome=clan_data["biome"],
             camp_bg=clan_data["camp_bg"],
             game_mode=clan_data["gamemode"],
+            cruel_cards=clan_data["cruel_cards"],
```
scribblecrumb:
```suggestion
            cruel_cards=clan_data.get("cruel_cards", []),
```
This is just a safety thing for old saves being converted. Since they won't have the "cruel_cards" key, it'll crash unless we're using `.get()`. 

`.get()` is essentially a safe way of retrieving a key! If it can't find the specified key, then it'll return `None` (or the specified value, here we've specified `[]`)

scribblecrumb (changes requested):

scribblecrumb:
Formatter needs to be run!

scribblecrumb (approved):

PR #5114, opened 2026-05-16
[BUGFIX] Fix Two Minor Allegiances Bugs

CL0WNTH0UGHTS:
## About The Pull Request
This PR fixes two minor bugs found in the allegiances screen:
1. Whenever a solo kit (w/o being raised by a queen) showed up in the allegencies, certain pelts were randomly capitalized (that being Rosette, Sokoke, and Masked). This removes that.
2. When non-cared-for kittens showed up in the screen, they had the short version of their pelt description. the short version is primarily used to keep the Queen lines short and succinct. This remedies this by allowing non-cared-for kittens to get their full pelt description.

## Proof of Testing
Before (both fixes): 
Screenshot 2026-05-16 160959
After (fix number 1):
Screenshot 2026-05-16 161211
After (fix number 2):
image

## Changelog/Credits
[BUGFIX] Fix two minor bugs on the allegiance screen


j-gynn:
This is weird, because those shouldn't be appearing in allegiances for non-queen-having kits. it should only use the short description for kits being cared for by a mother.

j-gynn (approved):

PR #5116, opened 2026-05-16
[REWORK] Inheritance

larkgz:
## About The Pull Request

A rework of inheritance. Does not include the extra information displayed on the Family Tree screen and the Choose Mates screen, which still use the old Inheritance code.

* More efficient. Building the new inheritance is done in O(n), and checks for family relationships are fast due to using sets.
* (In my opinion) easier to read and therefore easier to maintain. Everything is also properly type-hinted.
* Fixes bugs that were in the original, such as one where it apparently failed to load some faded offspring.

## Proof of Testing
* The tests pass using the new code, though the tests also don’t test very complex relationships.
* I initialized the old Inheritance code for EmberClan and HogClan (large Clans with complex relationships), and the Inheritance relative sets are subsets of the corresponding inheritance2 relative sets. That means that, at the very least, cats that *were* related shouldn’t become unrelated. Furthermore, when I inspected the discrepancies, they seemed to be caused by bugs in the original code (such as some faded cats not loading), or idiosyncrasies (cousins-siblings only being listed as siblings).
  * You can also do your own tests by messing around with `inheritance_db.compare_to_inheritance`, which is not called anywhere and just exists for testing purposes. I like to put it on the FamilyTreeScreen and then just flick through the cats to see if it freaks out.
* I edited the inheritance_db so the game would immediately crash if it tried to access uninitialized inheritance data, and it only crashed on `create_new_cat()` when trying to get the kits of a newly created cat being added to the Clan. Since the cat was just created, it won’t have kits (I think), so I think it's probably fine?


scripts/cat_relations/inheritance2.py
```diff
@@ -0,0 +1,257 @@
...
+
+class RelationType(StrEnum):
+    """An enum representing the possible relationships of a cat"""
+
+    BLOOD = "direct related"  # direct blood related - do not need a special print
+    ADOPTIVE = "adoptive"  # not blood related but close (parents, kits, siblings)
+
+    HALF_BLOOD = "half sibling"  # only one blood parent is the same (siblings only)
+    NOT_BLOOD = "not blood related"  # not blood related for parent siblings
+    RELATED = "blood related"  # related by blood (different mates only)
```
scribblecrumb: I'd prefer if the strings were snake case

scripts/cat_relations/inheritance2.py
```diff
@@ -0,0 +1,257 @@
...
+
+        if save:
+            self._saved_family_rels[cat.ID] = self._cat_to_rels[cat.ID]
+
+    def clear_inheritances(self):
+        """
+        Clears loaded inheritances. Make sure to call it when loading new Clans.
+        """
+        self._cat_to_litter = {}
+        self._saved_family_rels = {}
```
scribblecrumb, line 99: Maybe I'm misunderstanding, but doesn't `_cat_to_rels` also need to be cleared?

scripts/cat_relations/inheritance2.py
```diff
@@ -0,0 +1,257 @@
...
+            cat_a, cousin_allowed
+        ) or cat_a in self.get_relatives(cat_b, cousin_allowed)
+
+    def is_grandparent(self, maybe_grandparent: str, cat_a: str) -> bool:
+        return maybe_grandparent in self.get_grandparents(cat_a)
+
+    def is_parent(self, maybe_parent: str, cat_a: str) -> bool:
+        return maybe_parent in self.get_parents(cat_a)
+
+    def is_sibling(self, cat_b: str, cat_a: str) -> bool:
```
scribblecrumb: this is nitpicky, but in the spirit of consistency and general "this makes sense", lets put `cat_a` as the first arg

scripts/cat_relations/inheritance2.py
```diff
@@ -0,0 +1,257 @@
...
+        return cat_b in self.get_siblings(cat_a)
+
+    def is_uncle_aunt(self, maybe_uncle_aunt: str, cat_a: str) -> bool:
+        return cat_a in self.get_siblings_children(maybe_uncle_aunt)
+
+    def is_cousin(self, cat_b: str, cat_a: str) -> bool:
```
scribblecrumb: same here

scripts/cat_relations/inheritance2.py
```diff
@@ -0,0 +1,257 @@
...
+        return cat_b in self.get_cousins(cat_a)
+
+    def is_littermate(self, cat_a: str, cat_b: str) -> bool:
+        try:
+            return self._cat_to_litter[cat_a] == self._cat_to_litter[cat_b]
+        except KeyError:
+            return False
+
+    def compare_to_inheritance(self, cat_id: str, inheritance):
```
scribblecrumb, line 228: iirc your PR said that this func isn't actually used anymore - it's purely testing. let's have some docstrings for it that explain this and how to utilize it so that it isn't eventually deleted by someone trying to remove defunct code.

scribblecrumb (changes requested):

scripts/cat_relations/inheritance2.py
larkgz: Renamed function and added to docstring to clarify this.

scribblecrumb (approved):

PR #5120, opened 2026-05-17
[DEV] More type hinting for Cat

larkgz:
## About The Pull Request

Even though they're always IDs, I keep forgetting, so I added more type hinting to Cat.

* Less confusing.

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -9,7 +9,7 @@ import itertools
 import os.path
 import sys
 from random import choice, randint, sample, random, randrange
-from typing import Dict, List, Any, Union, Callable, Optional, TYPE_CHECKING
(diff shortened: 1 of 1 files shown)


scripts/cat/cats.py
```diff
@@ -924,7 +926,7 @@ def become_lost(self):
         for x in self.apprentice:
             Cat.fetch_cat(x).update_mentor()
 
-    def add_to_clan(self) -> list:
+    def add_to_clan(self) -> List[str]:
```
j-gynn, line 929: this doesn't always return a list, could you make it optional?

larkgz: When does it not return a list? It declares `ids = []` and returns `ids`, and it does not seem to have any early returns.

j-gynn (approved):

PR #5159, opened 2026-05-19
[BUGFIX] Romance filter

### LINKED ISSUE #5139: [BUG] Romantic apprentice moon event happening for siblings
### Pre-requisites

### Description

The wild fruit accessory moon event that increases romantic interest for two apprentices can happen for two apprentices that are siblings. No romantic interest is gained, but it still feels weird.

### Grade

### Gamemode

Expanded Mode

### Challenge cards

_No response_

1. Have apprentices that are siblings
2. Skip one moon
3. If applicable event doesn't occur, repeat

_No response_

v0.13.1

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

scribblecrumb:
## About The Pull Request
Whoopsie, seems we haven't been filtering for the `romance` tag in short events. This PR fixes that (as well as a small docs error)

## Linked Issues
fixes #5139

## Proof of Testing
image

Forced the event in my clan, which features apprentice siblings.

image
image

The two cats involved in the event aren't siblings! Additionally none of the existing siblings generated the event.

## Changelog/Credits
- Fixed the filtering for `romance` in short events


scripts/events_module/event_filters.py
```diff
@@ -584,6 +591,7 @@ def cat_for_event(
     comparison_cat_rel_status: list = None,
     injuries: list = None,
     return_id: bool = True,
+    tags: list = None,
```
larkgz: This will throw an error if `tags` is ever not set because `None` is not iterable. Note that `tags: list = []` won't work because it will cause every call of this function to use the same list.

larkgz (changes requested):

larkgz (approved):

PR #5163, opened 2026-05-20
[DOCS] specifying how interpersonal relationships are constrained in short events.

fulminousnight:
## About The Pull Request

- Adding a little bit more clarity to constraining for interpersonal events based on a convo in the discord.

- Helps writers ensure they know what they're doing with constraints. 

## Proof of Testing

Game runs, no obvious issues with the docs.

### DIFF
--- a/docs/dev/writing/reference/tag-lists.md
+++ b/docs/dev/writing/reference/tag-lists.md
@@ -454,7 +454,10 @@ You can utilize [#exclusionary tags](#exclusionary-tags).
 You can utilize [#exclusionary tags](#exclusionary-tags).
 
 !!! warning
-    Within ShortEvents, these should only be used in m_c's relationship constraints. They do not function as a pool. If you constrain a patrol to "child/parent", "app/mentor" the relationship between the cats <i>must</i> meet both criteria i.e. a parent who is the mentor to their apprentice child. 
+    Within ShortEvents, these should only be used in m_c's relationship constraints. 
+
+!!! warning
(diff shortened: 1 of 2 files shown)


docs/dev/writing/reference/tag-lists.md
```diff
@@ -454,7 +454,7 @@ You can utilize [#exclusionary tags](#exclusionary-tags).
 You can utilize [#exclusionary tags](#exclusionary-tags).
 
 !!! warning
-    Within ShortEvents, these should only be used in m_c's relationship constraints.
+    Within ShortEvents, these should only be used in m_c's relationship constraints. They do not function as a pool. If you constrain a patrol to "child/parent", "app/mentor" the relationship between the cats <i>must</i> meet both criteria i.e. a parent who is the mentor to their apprentice child.
```
scribblecrumb: This is it's functionality regardless of if it's in a short event or a patrol, so it'd be better for this addition to not be in the warning box specifically about short event use.

scribblecrumb (changes requested):

docs/dev/writing/reference/tag-lists.md
fulminousnight: Do you think it works better as its own warning or part of the body text above?

scribblecrumb: Perhaps it's own warning? Could reword to convey that most formats utilizing these tags will be treating it as a strict requirement list rather than a pool of possibilities. I'd like some sort of addendum tho that the reader should always refer back to the documentation for the specific event type to see if it's an exception to the rule.

scribblecrumb (approved):

PR #5168, opened 2026-05-20
[MODEL] Improve schemas for Points of Interest

larkgz:
## About The Pull Request

* Adds a schema for `points_of_interest.json`.
* Improves `poi` property schema to have better documentation and validation.

### Differences from documentation
* `category` uses `terrain` instead of `territory`, since `terrain` is what is used in the files.
* Includes tags for `prey:eggs`, `prey:fish`, and `nests`, which are used for some POIs.

## Proof of Testing

Tests are passing.

Autocomplete:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -21,6 +21,12 @@
                 "events/new_cat/*.json",
                 "events/injury/*.json"
             ]
+        },
+        {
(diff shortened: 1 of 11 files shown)


j-gynn (approved):
Thanks for catching my mistakes (and greatly improving on my bodged schema setup!)

PR #5173, opened 2026-05-21
[BUGFIX] Additional sprite conversion safety

scribblecrumb:
## About The Pull Request
Had someone on the discord reporting difficulty with their save being converted to the new version, specifically that they couldn't access the cat list due to a crash. I inspected their save file and this was due to a cat having a senior sprite index in their adult sprite parameter. I highly suspect this was done through save file editing, as I don't think this could have happened through normal gameplay. However, it's something easily resolved through some extra conversion handling, so I went ahead and added that. We just, essentially, default the sprite to a valid option if it's index doesn't match any of the valid sprites for that age.

## Proof of Testing
image

this is the save file from the reporter that would crash when trying to view the cat list, with my changes it now loads and doesn't crash! Chill is the specific cat who had been causing the problem, but we can see he also has a working sprite now.

## Changelog/Credits
- Added additional save conversion safety to account for sprites that might have indexes outside their expected range

### DIFF
--- a/scripts/cat/pelts.py
+++ b/scripts/cat/pelts.py
@@ -285,16 +285,16 @@ class Pelt:
                         self.cat_sprites[age] = "para_adult_short0"
                     continue
 
-                if age == CatAge.NEWBORN:
+                elif age == CatAge.NEWBORN:
                     self.cat_sprites[age] = (
                         "newborn2" if "newborn2" in self.newborn_poses else "newborn0"
                     )
                     continue
-                if age == CatAge.KITTEN:
+                elif age == CatAge.KITTEN:
                     # since these were at the top of the sheet, the pose nums were 0, 1, 2. thus they'll naturally match this fstring
-                    self.cat_sprites[age] = f"kitten{pose}"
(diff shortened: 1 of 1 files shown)


scripts/cat/pelts.py
```diff
@@ -305,8 +305,15 @@ def __init__(
...
                         self.cat_sprites[age] = f"adolescent_{fur}2"
-                    continue
-                if age in (CatAge.YOUNG_ADULT, CatAge.ADULT, CatAge.SENIOR_ADULT):
+                    else:
+                        self.cat_sprites[age] = choice(
+                            (
+                                f"adolescent{fur}0",
+                                f"adolescent{fur}1",
+                                f"adolescent{fur}2",
+                            )
```
larkgz, line 314:
```suggestion
                            (
                                f"adolescent_{fur}0",
                                f"adolescent_{fur}1",
                                f"adolescent_{fur}2",
                            )
```
Do they need underscores?

larkgz (approved):

PR #5179, opened 2026-05-22
[TASK] Addition of Config.py for Config Retrieval

### LINKED ISSUE #5083: [PROJECT] Function for retrieval of game_config modifiers
**Description**
For the cruel season update! Many of our cards will, ideally, be modifying existing game_config values. We'll want to try and make this as painless as possible, to avoid littering our code with a ton of `if "card" in clan.cruel_cards:` if-else nesting. We'll already have to deal with a bunch of these, so we'll want to do whatever we can to reduce it.

To handle modifying existing values, what we *could* do is set up a function that handles config retrieval. The idea would be to pass in the config we're looking for and then the func would check if any active cruel cards have a matching modifier ID. If they do, then the func returns the card's modifier instead of the `game_config` value.

Some things to consider from my vague pondering of this idea:
- Since our configs are a lot of nested dicts, we'll need to consider how we pass this info into the func as an arg as well as how we can properly indicate these modifiers in a cruel card's modifier dict.  I figure we could utilize a string system (i.e. `["graduation"]["min_graduating_age"]` becomes `"graduation.min_graduating_age"`) that's then unpacked by the function.
- All our current code references to game config will need to be adjusted to utilize this function either *now* or piecemeal by individual coders as they add cards which modify certain game configs. On one hand, if we get it all done now, then that will be less for us to worry about going forward. On the other hand, it'll greatly expand the scope of this PR.

**Scope of Project**
This is a more complex task to take on and might have a large scope, depending on how we handle existing config references. It's possible we could split this into multiple PRs, with the initial one handling the function set up and subsequent PRs converting chunks of the code to utilize that function.

This task should also wait until we've set up `cruel_card` attribute as mention in #5082

**Relevant Links**

Kippanovaa:
## About The Pull Request

addition of Config.py - the new and improved method of accessing game_config! more specifically the **get_config** function.
while the former implementation still exists, **this function should be used for any new accesses to the config file.**

A couple of calls have been converted already, but majority still have yet to be converted. 

Removes the necessity to go through game to access the config file, adhering more to software design principles. Also sets up the framework for cruel season ;)

## Linked Issues

Fixes: #5083 

cruel season discussion in discord: https://discord.com/channels/1003759225522110524/1492965987878961352

## Proof of Testing

cardtesting
war testing
card_testing2

### DIFF
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -664,7 +664,7 @@ def handle_focus():
         ]
 
         warrior_amount = len(healthy_warriors) * get_config(
-            game.clan, "focus.hunting." + CatRank.WARRIOR
+            game.clan, f"focus.hunting.{CatRank.WARRIOR}"
         )
 
         # handle apprentices
@@ -675,7 +675,7 @@ def handle_focus():
         ]
 
         app_amount = len(healthy_apprentices) * get_config(
-            game.clan, "focus.hunting." + CatRank.APPRENTICE
+            game.clan, f"focus.hunting.{CatRank.APPRENTICE}"
         )
 
         # finish
--- a/scripts/events.py
+++ b/scripts/events.py
@@ -7,7 +7,7 @@ TODO: Docs
 """
 import logging
 import random
-from scripts.config import get_config 
+from scripts.config import get_config
 
 # pylint: enable=line-too-long
 import traceback
@@ -663,9 +663,8 @@ def handle_focus():
             and cat.available_to_work()
         ]
 
-        warrior_amount = (
-            len(healthy_warriors)
-            * get_config(game.clan, "focus.hunting." + CatRank.WARRIOR)
+        warrior_amount = len(healthy_warriors) * get_config(
+            game.clan, "focus.hunting." + CatRank.WARRIOR
         )
 
         # handle apprentices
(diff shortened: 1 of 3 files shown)


scripts/events.py
```diff
@@ -662,9 +663,8 @@ def handle_focus():
             and cat.available_to_work()
         ]
 
-        warrior_amount = (
-            len(healthy_warriors)
-            * constants.CONFIG["focus"]["hunting"][CatRank.WARRIOR]
+        warrior_amount = len(healthy_warriors) * get_config(
+            game.clan, "focus.hunting." + CatRank.WARRIOR
```
scribblecrumb: can ya do these as f strings instead of concatenating the rank onto the end? just makes it more readable

scripts/events.py
```diff
@@ -674,9 +674,8 @@ def handle_focus():
             if cat.status.rank == CatRank.APPRENTICE and cat.available_to_work()
         ]
 
-        app_amount = (
-            len(healthy_apprentices)
-            * constants.CONFIG["focus"]["hunting"][CatRank.APPRENTICE]
+        app_amount = len(healthy_apprentices) * get_config(
+            game.clan, "focus.hunting." + CatRank.APPRENTICE
```
scribblecrumb: and here

scribblecrumb (changes requested):

scripts/events.py
Kippanovaa: yeah I just forgot f strings existed LMAO

scribblecrumb (approved):

PR #5258, opened 2026-06-01
[REVAMP] Thoughts reformat

scribblecrumb:
## About The Pull Request
This PR converts our thoughts into a new format: `TextPoolEvent`. Eventually I plan to convert other types of events (like relationships and ceremonies) into this format as well, to help increase parity.

Notes:
- Thought documentation was updated. I would suggest reading this first to get an idea of how the format is intended to function.
- The thoughts were reformatted via a script. I really don't expect anyone to read all of those changes. I skimmed through a bunch of the files and it all looked to have converted correctly.
- You'll notice some of the changes to event filter are marked with a TODO comment akin to "could replace such and such in the future".  I intend for a lot of the `TextPoolEvent` params to eventually become the norm across all our formats; such as the `stat` param, for example.  As such I decided to mark some things to help us keep track of what the "goal" is whenever we seem to have a redundant function.
- I killed the old `test_thoughts.py` because it seemed largely defunct.  Instead I wrote our new tests into the `test_event_filters.py` file.  All the new parameters have associated tests for them.
- `TextPoolEvent` is a slotted dataclass. While slightly slower to create, these are about twice as fast to use (and take less memory) compared to dicts. I also added `TypedDicts` for all of the dict parameters, just for some organization and IDE hinting.
- Thought loading was slightly tweaked so that, instead of reloading all of the thought files every single time we try to find a thought, we instead use a similar system to the short event loading where the loaded thoughts are held within the thought generation module and reused.

Brings us toward the end goal of format parity!  This is setting us up for future changes that will increase the writer's abilities to constrain for certain circumstances across all events.  While I was putting this new format together I did consult the writing team multiple times to ensure it felt functional to them.

## Proof of Testing
image
image
image
image
image
image
image
image

## Changelog/Credits
- Thought documentation was updated for our new format.
- All thoughts were reformatted via a script.
- New event filters added.
- Added new event filter tests.
- Added the new event dataclass `TextPoolEvent`
- Thought loading was slightly tweaked so that, instead of reloading all of the thought files every single time we try to find a thought, we instead use a similar system to the short event loading where the loaded thoughts are held within the thought generation module and reused.


scripts/models/common/group.py
```diff
@@ -0,0 +1,37 @@
+from enum import Enum
+from typing import Annotated, Union
+
+from pydantic import RootModel, StringConstraints, AfterValidator
+
+from scripts.models.common.cat import validate_cat_abbr
```
larkgz: Can you move this function to this file?

scripts/models/thought/cat_dict.py
```diff
@@ -0,0 +1,23 @@
...
+from scripts.models.common.age import Age
+from scripts.models.common.backstory import Backstory
+from scripts.models.common.group import Group
+from scripts.models.thought.health_dict import HealthDict
+from scripts.models.thought.standing_dict import StandingDict
+from scripts.models.thought.stat_dict import StatDict
+from scripts.models.thought.status_any import StatusAny
+
+
+class CatDict(BaseModel):
```
larkgz, line 15: Can you add `model_config = ConfigDict(extra="forbid")` so that the schema checker can more easily detect typos?

Same for all the other ones.

larkgz (changes requested):

larkgz (approved):

PR #5298, opened 2026-06-05
[BALANCE] more reasonable random herb balancing

scribblecrumb:
## About The Pull Request
Balancing for patrols gathering random herbs has been a little bit overtuned. This was due to the mismatch in intended balancing between med cats gathering over moonskip and the amount gathered during patrols.

To address this:
- I've added a way to specify a max total quantity of herbs to be gathered when using the `get_found_herbs()` func
- A matching game_config was set up to determine how much a single cat in a patrol can "carry home"
- When determining the random herbs gathered by a patrol, we now decide how many herbs they can have at a maximum based on how many cats are on the patrol and if any of them have the CLEVER skill (allows a cat to be "smarter" about carrying herbs and thus carry more)
- We then find the random herbs for the patrol. Once we hit our total quantity limit, we stop gathering herbs.

Fixes the weirdness of a patrol somehow gathering like... 50 doses or more of herbs. It just felt really odd and unfeasible.

## Proof of Testing
image

more reasonable for 3 cats to carry

image

reasonable for 2 cats to carry

image

Patrols that gather specific herbs still work as before

## Changelog/Credits
- I've added a way to specify a max total quantity of herbs to be gathered when using the `get_found_herbs()` func
- A matching game_config was set up to determine how much a single cat in a patrol can "carry home"
- When determining the random herbs gathered by a patrol, we now decide how many herbs they can have at a maximum based on how many cats are on the patrol and if any of them have the CLEVER skill (allows a cat to be "smarter" about carrying herbs and thus carry more)
- We then find the random herbs for the patrol. Once we hit our total quantity limit, we stop gathering herbs.


larkgz:
Is this targetting/based on the correct branch?

scribblecrumb:
> Is this targetting/based on the correct branch?

Yup! I wanted to fix it on the release branch since it's funky balancing that present in stable. I did at one point merge dev into it on accident and I did revert that, though I think thats why it's still listing a bunch of dev commits.  The actual files changed are correct though.

j-gynn (approved):

Issue #5303, opened 2026-06-06
[UI] text overlapping in leader's den


Inkpawwc:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

outsider_notice_text overlaps with outsider_rep text when leader's name is too long

### Grade

Non-vital - Can easily be ignored and does not have a large impact on play

### Steps to reproduce the behavior

1. Have a leader with a name that's 12 or more characters
2. go to the leader's den
3. switch to the outsiders tab
4. observe overlapping text

### Commit # or Game Version Number

52930a4e

### Is this bug only happening in fullscreen?

No

### Screenshots

<img width="699" height="612" alt="Image" src="https://github.com/user-attachments/assets/e6e996a1-4d0c-41ee-8668-86072a139634" />

PR #5304, opened 2026-06-06
[BUGFIX / UI] Leader Den Text Overlapping

Inkpawwc:
## About The Pull Request

updates position of text box based on the status of the leader (healthy, sick with helper, sick without helper, no leader) to keep the textbox centered and prevent text overlap. the clan selection frame was moved down slightly to prevent overlap as well.

## Linked Issues

Fixes:#5303 and #5324 
## Proof of Testing

https://discord.com/channels/1003759225522110524/1005586496142704800/1516359900668756070
https://discord.com/channels/1003759225522110524/1005586496142704800/1516360726552383638

video files were too big to add here directly so here's the link to them in discord

## Changelog/Credits

Fixed leader den text overlapping when cat, clan, and/or temper names are long
-inkpawwc


scripts/screens/LeaderDenScreen.py
```diff
@@ -1043,9 +1056,13 @@ def update_text(self, clan=True):
...
 
-            self.screen_elements["temper_text"].set_text(
-                "screens.leader_den.outsider_rep",
+            self.screen_elements["outsider_rep"] = pygame_gui.elements.UILabel(
+                relative_rect=ui_scale(pygame.Rect((68, -7), (445, -1))),
+                text="screens.leader_den.outsider_rep",
+                object_id=get_text_box_theme("#text_box_30_horizcenter"),
+                manager=MANAGER,
                 text_kwargs={"reputation": i18n.t(f"screens.leader_den.{reputation}")},
+                anchors={"top_target": self.screen_elements["outsider_notice_text"]},
```
scribblecrumb: This should be a text box, not a label!

I would also say it's better to be creating this up alongside `temper_text` and just hiding/showing it instead of making it down here.

scripts/screens/LeaderDenScreen.py
```diff
@@ -279,6 +287,7 @@ def screen_switches(self):
                 "count": 1,
                 "m_c": game.clan.leader if not self.no_leader else None,
             },
+            anchors={"top_target": self.screen_elements["lead_image"]},
         )
```
scribblecrumb, line 285: I would highly recommend against this `clan_notice_y` and `outsider_notice_y` stuff you have going on. It's just really unmaintainable.

Rather, I'd like to see you use the anchors for this.  Make the `top_target` of `clan_notice_text` and `outsider_notice_text` the `bg_image`, since we know that will always be present and always be a fixed position/size, unlike the cat images.

scribblecrumb (changes requested):
Image

selection frame now awkwardly overlaps with longer temperaments

scripts/screens/LeaderDenScreen.py
Inkpawwc: image
^ current look of everything now w/ the requested changes ^^
thank you for explaining this so kindly, looking back at this code I have no idea why I decided to do it like that lmao

scribblecrumb (approved):

Issue #5213, opened 2026-05-26
[BUG] Moon event not appearing in other clans tab


little-creature:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

The moon event shown in the first screenshot below, about the death of an otherclan member, doesn't appear in the other clans tab on the events menu

### Grade

Unimportant - Little to no impact on play

### Gamemode

Expanded Mode

### Steps to reproduce the behavior

1. Have cats in other clans
2. Timeskip until you get the event

### Commit # or Game Version Number

v0.13.2

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/93c8e7f9-eb5d-4b83-aa0f-251f9f67d67b" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/245baf93-4217-4170-a3a1-68ff48b6a944" />

ssuncreature:
Should be closed - PR was merged

PR #5318, opened 2026-06-11
[BUGFIX] Other Clan Deaths show in Other Clans tab

ssuncreature:
## About The Pull Request

- Makes sure events involving other clan cats dying actually appear in other clans section

## Linked Issues

Fixes: #5213

## Proof of Testing

Screen Shot 2026-06-11 at 13 05 09
Screen Shot 2026-06-11 at 13 05 17
Screen Shot 2026-06-11 at 13 05 33

### DIFF
--- a/scripts/events_module/outsider_events.py
+++ b/scripts/events_module/outsider_events.py
@@ -81,7 +81,9 @@ class OutsiderEvents:
                 and not cat.status.get_last_valid_group_id() == CatGroup.PLAYER_CLAN_ID
             ):
                 game.cur_events_list.append(
-                    Single_Event(text, ["birth_death", "other_clans"], cat_dict={"m_c": cat})
+                    Single_Event(
+                        text, ["birth_death", "other_clans"], cat_dict={"m_c": cat}
+                    )
                 )
             else:
                 game.cur_events_list.append(
--- a/scripts/events_module/outsider_events.py
+++ b/scripts/events_module/outsider_events.py
@@ -39,10 +39,7 @@ class OutsiderEvents:
             elif cat.status.is_lost(CatGroup.PLAYER_CLAN_ID):
                 text = random.choice(deaths["lost"])
                 death_history = i18n.t("events.death.outsider_deaths.history.lost")
-            elif cat.status.is_other_clancat or (
-                cat.status.is_former_clancat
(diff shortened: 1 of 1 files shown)


scripts/events_module/outsider_events.py
```diff
@@ -76,6 +76,16 @@ def killing_outsiders(cat: "Cat"):
 
             cat.history.add_death(death_text=death_history)
             cat.die(grief_allowed=False)
-            game.cur_events_list.append(
-                Single_Event(text, "birth_death", cat_dict={"m_c": cat})
-            )
+            if cat.status.is_other_clancat or (
+                cat.status.is_former_clancat
+                and not cat.status.get_last_valid_group_id() == CatGroup.PLAYER_CLAN_ID
```
scribblecrumb: why are we including former clancats in this? They're no longer part of any clan, so I don't feel they should be included in that tab.

scribblecrumb (changes requested):

scripts/events_module/outsider_events.py
ssuncreature: Cant remember my reason for that - something to do with the previous code. Looking back, I think i'll remove that

ssuncreature:
"formatting check failed"
Screenshot 2026-06-19 at 18 19 25
Why must you forsake me github

scripts/events_module/outsider_events.py
```diff
@@ -76,6 +76,16 @@ def killing_outsiders(cat: "Cat"):
 
             cat.history.add_death(death_text=death_history)
             cat.die(grief_allowed=False)
-            game.cur_events_list.append(
-                Single_Event(text, "birth_death", cat_dict={"m_c": cat})
-            )
+            if (
+                cat.status.is_other_clancat
+                and not cat.status.get_last_valid_group_id() == CatGroup.PLAYER_CLAN_ID
```
scribblecrumb: you shouldn't need to check the group ID. `is_other_clancat` should only return `True` if the cat is `CLANCAT` and isn't part of the player clan. So the extra group ID check here is redundant.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5321, opened 2026-06-12
[ENHANCEMENT] Remove Hardcoding for removal of tail accessories

### LINKED ISSUE #4341: [ENHANCE] Fix handling of accessory removal when losing tail
### Description

I am currently working in this area so please hold off on claiming this until that's merged (PR incoming shortly, will link when it's ready).

I have noticed that `get_permanent_condition` in `cats.py` has a nasty section where it removes invalid accessories that is currently hardcoded. Maybe we can use the slots that were added for accessories to determine this instead?

Edit to add: The same code appears in `Scar_Events` in `handle_scars`. Can we consolidate these?

Inkpawwc:
## About The Pull Request

accessories are now removed if they are specified as "tail" accessories in their sprite data json when the tail is lost. 

dehardcodes accessories that get removed when the tail is lost
## Linked Issues

Fixes: #4341
## Proof of Testing

image
image
image
image

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -27,7 +27,6 @@ from scripts.cat.enums import (
 from scripts.cat.history import History
 from scripts.cat.names import Name
 from scripts.cat.pelts import Pelt
-from scripts.cat.sprites.load_sprites import Sprites
 from scripts.cat.personality import Personality
 from scripts.cat.skills import CatSkills
 from scripts.cat.status import Status, StatusDict
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -123,18 +123,6 @@ class Cat:
     all_cats_list: List[Cat] = []
     ordered_cat_list: List[Cat] = []
 
-    must_have_tail: list = []
-
-    for sprite_list in Sprites.PLANT_DATA["sprite_list"]:
-        for acc in sprite_list:
-            if sprite_list[acc] == "tail":
-                must_have_tail.append(acc)
-
-    for sprite_list in Sprites.WILD_DATA["sprite_list"]:
-        for acc in sprite_list:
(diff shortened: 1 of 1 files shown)


scripts/cat/cats.py
```diff
@@ -2050,20 +2063,7 @@ def get_permanent_condition(self, name, born_with=False, event_triggered=False):
...
-                    "BLUE FEATHERS",
-                    "JAY FEATHERS",
-                    "GULL FEATHERS",
-                    "SPARROW FEATHERS",
-                    "CLOVER",
-                    "DAISY",
-                    "WISTERIA",
-                    "GOLDEN CREEPING JENNY",
-                )
+                acc for acc in self.pelt.accessory if acc not in Cat.must_have_tail
```
scribblecrumb: you could do
```suggestion
                acc for acc in self.pelt.accessory if acc not in Pelt.tail_accessories
```
you don't actually need to compile a new list of tail accs, we already compile one in `Pelt` called `tail_accessories`.

scribblecrumb (changes requested):

scribblecrumb (approved):

scripts/cat/cats.py
```diff
@@ -27,6 +27,7 @@
 from scripts.cat.history import History
 from scripts.cat.names import Name
 from scripts.cat.pelts import Pelt
+from scripts.cat.sprites.load_sprites import Sprites
```
scribblecrumb: is this import statement needed?

scribblecrumb (changes requested):

scripts/cat/cats.py
Inkpawwc: oh whoops forgot abt that, that was used for getting the info used by must_have_tail

scribblecrumb (approved):

Issue #5212, opened 2026-05-26
[BUG] Warrior treated as med cat on patrol


little-creature:
### Pre-requisites

- [x] I have searched the existing issues
- [x] This issue was encountered (or replicated) on a non-edited Clan

### Description

On the specified herb gathering patrol, a non-medcat is treated as the medicine cat for both the patrol outcome and the resulting relationships.

### Grade

Regular - Should be handled soon, but doesn't necessarily break the game or inhibit the player

### Gamemode

Expanded Mode

### Steps to reproduce the behavior

1. Send cats on this patrol
2. Observe

### Patrol ID + Outcome Print (if applicable)

gen_med_gatheringcatmint_leaf-fall3

### Commit # or Game Version Number

v0.13.2

### Additional context

Patchpelt is a warrior, and Yellowfang is the medicine cat

### Screenshots

<img width="494" height="238" alt="Image" src="https://github.com/user-attachments/assets/43a79732-6c62-4a7c-ad9e-20e43bd27d50" />

<img width="775" height="185" alt="Image" src="https://github.com/user-attachments/assets/f178fbb8-c9cc-4340-9c57-8a03ec802bbc" />

ssuncreature:
Can be closed - PR merged

scribblecrumb:
Reopening as I don't believe #5221 actually fixed this

scribblecrumb:
Closing this as we haven't been able to replicate and, based on the screenshot, this seems to have occurred on an edited save (considering that those look to be recreations of canon cats).  please rereport if you encounter this bug on an un-edited save!

PR #5354, opened 2026-06-24
[BUGFIX] Adjust patrol leader selection

ssuncreature:
## About The Pull Request

- Adjust patrol leader selection to only run eldest/most experience selection if there is not already a patrol leader
## Linked Issues

Fixes: #5212

## Proof of Testing

With two warriors:
Screen Shot 2026-06-24 at 12 01 12

With leader:
Screen Shot 2026-06-24 at 12 02 50

Training Patrol:
Screen Shot 2026-06-24 at 12 12 37

Hunting Patrol:
Screen Shot 2026-06-24 at 12 12 55

Border Patrol:
Screen Shot 2026-06-24 at 12 13 38
*apprenticeleap is a warrior in this - didnt change name but is youngest/least experienced

Non-reported med patrol:
Screen Shot 2026-06-24 at 12 13 11

### DIFF
--- a/scripts/events_module/patrol/patrol.py
+++ b/scripts/events_module/patrol/patrol.py
@@ -246,7 +246,7 @@ class Patrol:
             game.patrolled.append(cat.ID)
 
         # PATROL LEADER AND RANDOM CAT CAN NOT CHANGE AFTER SET-UP
-
+        # if self.patrol_cats[index] == None as failsafe
         # DETERMINE PATROL LEADER
         # sets medcat as leader if they're in the patrol
         if CatRank.MEDICINE_CAT in self.patrol_status_list:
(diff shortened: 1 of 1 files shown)


scripts/events_module/patrol/patrol.py
```diff
@@ -246,29 +246,32 @@ def add_patrol_cats(self, patrol_cats: List[Cat], clan: Clan) -> None:
...
             index = self.patrol_status_list.index(CatRank.LEADER)
             self.patrol_leader = self.patrol_cats[index]
-        elif CatRank.DEPUTY in self.patrol_status_list:
+        elif CatRank.DEPUTY in self.patrol_status_list and self.patrol_leader == None:
             index = self.patrol_status_list.index(CatRank.DEPUTY)
             self.patrol_leader = self.patrol_cats[index]
-        else:
-            # Get the oldest cat
+        elif self.patrol_leader == None:
+            # if still no patrol leader, get the oldest cat
```
scribblecrumb, line 274: Why doesn't the initial `CatRank.MEDICINE_CAT` `if` statement also include the `self.patrol_leader == None` bit that all the others do?

ssuncreature: I assumed it wouldn't be necessary as it's the first patrol_leader change in the file. I'm fully willing to add it in if it would be preffered

scribblecrumb: I spose I'm a little confused on the problem we're solving. Is it that the patrol leader could "change" between intro and outcome?  If so, then I would assume we wouldn't want to flip flop between medicine cats either, in which case line 252 would also need to check for `None`.

Though, honestly, I think it feels a little weird to *be* checking every single one of these statements for `None`.  It's a bit too repetitive. I'd rather you start with a beginning `if self.patrol_leader == None:` and then have all the rest of this nested behind it.  That way we just establish from the start that everything we're doing assumes that there isn't a `self.patrol_leader`.

ssuncreature: What I got when I was trying to recreate the bug and play around to find the problem point was that the else statement was overriding the series of if/elif before the start of patrol

I'll remove the extra checks that were just in case if thats more preferable? (sorry if i dont make much sense I'm tired)

PR #5381, opened 2026-06-30
[CONTENT] parent/child thoughts

fulminousnight:
## About The Pull Request

- A series of thoughts for parent cats to have about their children, both young and old.

- More variety and a new niche in thoughts as far as I am aware.

## Proof of Testing
Game runs smoothly and thoughts generate organically.
Screenshot 2026-06-30 at 3 55 57 PM
Screenshot 2026-06-30 at 3 55 35 PM

### DIFF
--- a/resources/lang/en/thoughts/while_alive/general.json
+++ b/resources/lang/en/thoughts/while_alive/general.json
@@ -1397,7 +1397,7 @@
     "strings": [
         "Wonders if {PRONOUN/m_c/subject} {VERB/m_c/are/is} a good parent",
         "Worries about r_c's health",
-        "Worries about {PRONOUN/m_c/poss} kit's futures",
+        "Worries about {PRONOUN/m_c/poss} kit's future",
         "Worries {PRONOUN/m_c/subject} {VERB/m_c/are/is} a bad parent",
         "Hopes StarClan watches over r_c"
     ],
--- a/resources/lang/en/thoughts/while_alive/general.json
+++ b/resources/lang/en/thoughts/while_alive/general.json
@@ -1391,5 +1391,80 @@
                 ]
             }
         }
-    }
+    },
+    {
+    "id": "gen_parent_gen",
+    "strings": [
+        "Wonders if {PRONOUN/m_c/subject} {VERB/mc/are/is} a good parent",
+        "Worries about r_c's health",
+        "Worries about {PRONOUN/m_c/poss} kits' futures",
+        "Wishes {PRONOUN/m_c/subject} felt like a better parent",
+        "Hopes StarClan watches over {PRONOUN/m_c/poss} kits"
+    ],
+    "involved_cats": {
+        "m_c": {},
+        "r_c": {}
+        },
+        "relationship_constraint": [
+            {
+                "cats_from": ["m_c"],
+                "cats_to": ["r_c"],
+                "mutual": false,
+                "constraints": ["parent/child"]
+            }
+        ]
+},
+    {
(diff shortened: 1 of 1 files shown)


resources/lang/en/thoughts/while_alive/general.json
```diff
@@ -1391,5 +1391,80 @@
...
+                "constraints": ["parent/child"]
+            }
+        ]
+},
+    {
+    "id": "gen_parent_adultkit",
+    "strings": [
+        "Misses when {PRONOUN/m_c/poss} kits were young",
+        "Is so proud of {PRONOUN/m_c/poss} kits",
+        "Worries about {PRONOUN/m_c/poss} kits' futures",
```
fulminousnight: I checked the grammar on this like six times and afaik this is correct in American English.

scribblecrumb (changes requested):
okay potential problem: lots of these reference "kits" as in more than one, but we can't guarantee that the cat actually has more than one kitten. so I'd suggest reassessing those thoughts.

fulminousnight:
> okay potential problem: lots of these reference "kits" as in more than one, but we can't guarantee that the cat actually has more than one kitten. so I'd suggest reassessing those thoughts.

To resolve this, do you think adding r_c in place of "kits" or simply rewording from plural to singular would be better?

resources/lang/en/thoughts/while_alive/general.json
```diff
@@ -1391,5 +1391,79 @@
...
             }
         }
-    }
+    },
+    {
+    "id": "gen_parent_gen",
+    "strings": [
+        "Wonders if {PRONOUN/m_c/subject} {VERB/m_c/are/is} a good parent",
+        "Worries about r_c's health",
+        "Worries about {PRONOUN/m_c/poss} kit's futures",
```
scribblecrumb:
```suggestion
        "Worries about {PRONOUN/m_c/poss} kit's future",
```

resources/lang/en/thoughts/while_alive/general.json
```diff
@@ -1391,5 +1391,79 @@
...
+                "constraints": ["parent/child"]
+            }
+        ]
+},
+    {
+    "id": "gen_parent_adultkit",
+    "strings": [
+        "Misses when r_c was young",
+        "Is so proud of r_c",
+        "Worries about {PRONOUN/m_c/poss} kit's future",
```
scribblecrumb: no need to repeat this imo

resources/lang/en/thoughts/while_alive/general.json
```diff
@@ -1391,5 +1391,79 @@
...
+    "strings": [
+        "Is exhausted chasing after r_c",
+        "Is snuggled in {PRONOUN/m_c/poss} nest with {PRONOUN/m_c/poss} litter",
+        "Grooms r_c while {PRONOUN/r_c/subject} squeak complaints",
+        "Doesn't want r_c to grow up and face the dangers of the world"
+    ],
+    "involved_cats": {
+        "m_c": {},
+        "r_c": {
+            "age": ["kitten"]
```
scribblecrumb:
```suggestion
            "age": ["newborn", "kitten"]
```
could include both! while we generally would see this overridden by the `on_birth` thoughts for newborns, it's possible someone will reload the game and cause the thoughts to all be regenerated. in which case, this thought would be possible.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5383, opened 2026-07-01
[REFACTOR] New cat creation func

scribblecrumb:
## About The Pull Request
This adds the `updated_create_new_cat()` func and full test coverage for all of its usage. Like, maybe the most thorough testing I've ever written.

The new func isn't actually plugged into the code, so it won't run in-game at all yet. It's part of the larger patrol refactor I'm doing, but I'm trying to chunk the refactor up as much as possible so that the PRs aren't as insane as they could be.  Thus I decided to just PR this function in on its own.  The tests show that it's working as intended, so when it *is* plugged into the code there shouldn't be any problems.

Notes:
- The new func is utilizing the `involved_cats` format that has been established for rel events and thoughts.  A couple additions were made to accommodate all the needs of new cat creation.  Essentially, rather than acting as *constraints*, the parameters of `involved_cats` is now acting as *options* to be chosen from. This means that there's now significantly more functionality in the sorts of new cats that can be created!
- All prior functionality has been preserved, it's just be rewritten to be so much cleaner.
- The `set_mate` func in `cat` now has an arg to disable the inheritance recalculation. This was necessary for mate assignment during cat creation since inheritance is already recalculated at the end of cat creation.  Also calculating it when setting mates would mean we do it twice, which is bad for optimization.
- `add_to_group` func now lets you specify a rank for the cat to immediately become. This was needed for streamlining the assignment of groups and ranks for new cats that we create.

We're cleanin things up!  When this is actually hooked up, it'll add a ton of functionality that writers have been requesting.

## Proof of Testing
If the tests pass then there you go!

## Changelog/Credits
- Rewrites the function used to create new cats. This is just preparatory though, as the function isn't yet hooked in to the game (I'm just trying to split up my work and be kind to the one who reviews it all)


scripts/cat/status.py
```diff
@@ -539,7 +541,9 @@ def add_to_group(
             standing_with_past_group = None
 
         # if we're moving an afterlife cat, they don't change rank
```
j-gynn: maybe move/amend this comment so it's clear that the afterlife part isn't `become_rank`

scripts/events_module/patrol/create_new_cat.py
```diff
@@ -0,0 +1,470 @@
...
+        categories = BACKSTORIES["backstory_categories"]
+
+        if social == CatSocial.LONER:
+            created_cat.backstory = (
+                choice(categories["baby_loner_backstories"])
+                if baby
+                else choice(categories["loner_backstories"])
+            )
+        elif social == CatSocial.ROGUE:
+            # there's no baby category for these for some reason...
```
j-gynn: don't wait! don't ask! report that bug! /hj

j-gynn (commented):
two things for housekeeping

scripts/events_module/patrol/create_new_cat.py
scribblecrumb: ended up just fixing it and adding a baby rogue category.  looks like none of the baby loner ones were actually specific to loners, so i just copy-pasted all of those entries into the baby rogue one.

j-gynn (approved):

PR #5393, opened 2026-07-04
[BUGFIX / ENHANCEMENT] ChangeGenderScreen non-hardcoded cases

### LINKED ISSUE #5376: [ENHANCE] Rework change gender screen
### Description

Currently ChangeGenderScreen is still hardcoded to use English pronoun sets so the game crashes if using another language that has different pronouns.
This needs to be reworked before merging any localisation work.

afellowcorn:
## About The Pull Request
Fixes game crashing when opening the change gender screen with a language that uses different cases than English. The pronoun set display now iterates through all cases provided in the pronoun set instead.

## Linked Issues
Closes #5376 

## Proof of Testing
Zrzut ekranu 2026-07-04 183743
Zrzut ekranu 2026-07-04 183724

>(Note stray slashes are due to display text shortening)

## Changelog/Credits
Fixes game crashing when opening the change gender screen with a language that uses different cases than English.


scripts/screens/ChangeGenderScreen.py
```diff
@@ -310,10 +309,16 @@ def pronoun_update(self):
...
+            pronounset_keys = list(pronounset.keys())
+            pronounset_keys.remove("conju")
+            pronounset_keys.remove("gender")
+
+            for i, case in enumerate(pronounset_keys):
+                if i >= len(pronounset_keys) - 1:
+                    displayname.append(f"{pronounset[case]}")
+                else:
+                    displayname.append(f"{pronounset[case]}/")
             short_name = shorten_text_to_fit(displayname, 170, 13)
```
scribblecrumb: could this be simplified by using `"/".join()`?

scripts/screens/ChangeGenderScreen.py
```diff
@@ -424,10 +429,16 @@ def preset_update(self):
...
+            pronounset_keys = list(pronounset.keys())
+            pronounset_keys.remove("conju")
+            pronounset_keys.remove("gender")
+
+            for i, case in enumerate(pronounset_keys):
+                if i >= len(pronounset_keys) - 1:
+                    displayname.append(f"{pronounset[case]}")
+                else:
+                    displayname.append(f"{pronounset[case]}/")
             short_name = shorten_text_to_fit(displayname, 140, 13)
```
scribblecrumb: And if we're doing the exact same thing here, then we could make this code block into a function to be called instead. That way we avoid code repetition.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5398, opened 2026-07-06
[CLEANUP & BALANCE] Romantic events

### LINKED ISSUE #5395: [ENHANCE] Rejections & Confessions
### Description

As discussed multiple times in the Discord, rejections & confessions could use a bit of a tweak. The main issues seen are:
- Cats aren't grieving their mates enough, which often causes them to immediately take a new mate on the same moon their previous mate dies because they have high relationships with everyone. A side effect of this is that when a popular, well-liked cat loses their mate, half the Clan will instantly confess to them.

- The relationship change when a cat's confession is rejected is extremely small and barely does anything. (It seems to be a tiny _one-sided_ decrease in comfort in the cat that got confessed to from what I've noticed.) This tends to lead to cats continously confessing to the same cat over and over for several moons until they either die or the other cat finally caves in and begins dating them, which is extremely uncomfortable when it happens every time as opposed to sometimes (which would give the cats more character. They can be like Ashfur every now and then, but not always!)

Suggested ways to address this would be a) decreasing the chance for confessions in general, b) make rejections negatively impact relationships more heavily, c) addressing the constant confessions on the situations where they _do_ happen with dedicated event text, and/or d) making confession chance be determined by trait/facets (so for example, lonesome cats are less likely to confess, arrogant cats are more likely to be insistent, and loving/compassionate cats can confess quickly even if the romance level isn't that high)

scribblecrumb:
## About The Pull Request
This was mainly focused on implementing some changes to balance confessions as suggested in #5395, however this file is also SO very old, so it needed some TLC to get it back up to par.  I focused on cutting down redundant checks, implementing `get_config()`, and general readability.

- Declassified `romantic_events.py` cus it was all staticmethods anyways
- Interactions with grief stricken cats is more in-depth now.  There's a general chance to pass when a cat attempts to become mates with a grief stricken cat. This chance is influenced by facets.
- Cats moving on from dead mates is a bit more in-depth.  The chance to move on is now influenced by facets.
- Confessions were given more oomf when it comes to relationship changes for both rejections and acceptances.
- All mate/breakup events now add to the relationship log
- Lots of previously hardcoded numbers have been added to game_config
- Renamed `breakup` arg in `unset_mate` to `user_initiated_breakup` since that's it's true use and it was confusing 
- Added `moons_as` property to `Status` to be a quick way of gaining access to that info
- Made some necessary changes to `get_config()` as I found the functionality a little lacking when actually putting it to the test. Biggest issue was that you could end up directly modifying config values without really realizing it. I've set it up to `deepcopy()` when necessary to prevent that problem.  Also made it use `game.constants.CONFIG` cus we really don't need to be loading the file again.
- Renamed a lot of the functions in `romantic_events.py` to be clearer
- Moved some tests out of `test_relation_events.py` because they were better suited to `test_romantic_events.py`
- Rewrote existing romance tests and added more extensive test coverage

Hopefully helps rejections feel more meaningful!  Makes our code generally easier to read and modify in the future.

## Linked Issues
closes #5395

## Proof of Testing
Test coverage added!

## Changelog/Credits

- Interactions with grief stricken cats is more in-depth now.  There's a general chance to pass when a cat attempts to become mates with a grief stricken cat. This chance is influenced by facets.
- Cats moving on from dead mates is a bit more in-depth.  The chance to move on is now influenced by facets.
- Confessions were given more oomf when it comes to relationship changes for both rejections and acceptances.
- All mate/breakup events now add to the relationship log
- Lots of previously hardcoded numbers have been added to game_config
- Added `moons_as` property to `Status` to be a quick way of gaining access to that info
- Made some necessary changes to `get_config()` as I found the functionality a little lacking when actually putting it to the test. Biggest issue was that you could end up directly modifying config values without really realizing it. I've set it up to `deepcopy()` when necessary to prevent that problem.  Also made it use `game.constants.CONFIG` cus we really don't need to be loading the file again.
- Moved some tests out of `test_relation_events.py` because they were better suited to `test_romantic_events.py`
- Rewrote existing romance tests and added more extensive test coverage


scripts/config.py
```diff
@@ -1,14 +1,12 @@
...
 from scripts.game_structure import constants, game
 
-with open("resources/game_config.toml", "r", encoding="utf-8") as read_file:
-    CONFIG = tomllib.loads(read_file.read())
-
 
 # config_path passed as a string using dot notation - ex "graduation.min_graduating_age"
 def get_config(config_path):
-    config_value = CONFIG
+    config_value = game.constants.CONFIG
```
j-gynn: should be `constants.CONFIG`, which is... this file.

`game.constants.config` is deprecated, it warns about it, and then points right back here. constants no longer exist.

this doesn't load config *again*, it loads it *at all* 😂

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+        # check valid mate
+        if mate_id not in Cat.all_cats:
+            print(f"WARNING: Cat #{cat} has a invalid mate. It will be removed.")
+            cat.mate.remove(mate_id)
+            continue
+
+        mate: Cat = Cat.fetch_cat(mate_id)
+        # check the mate's setting
+        if mate.no_mates:
             return
```
j-gynn: quits early for poly cats? intended?

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
 
-        if cat_from.ID not in cat_to.mate:
-            return False
+    for x in cat.mate:
+        mate = Cat.fetch_cat(x)
 
-        if cat_from.no_mates or cat_to.no_mates:
-            return False
+        # check the mate's setting
+        if mate.no_mates:
```
j-gynn, line 109: again, intended early quit for poly mates? i think these need renaming or clarifying or something if that's intended...

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+BREAKUP_STRINGS = {}
+POLY_MATE_DICTS = {}
 
-        RomanticEvents.current_loaded_lang = i18n.config.get("locale")
 
-    @staticmethod
-    def handle_mating_and_breakup(cat):
-        """Handle events related to making new mates, and breaking up."""
+def handle_mates_and_breakup(cat: Cat):
+    """Checks if the given cat should move on from prior mates, breakup current ones, or take on a new one."""
```
j-gynn: docstring variables pls

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
-            Single_Event(
-                text,
-                ["relation", "misc"],
-                [cat_from.ID, cat_to.ID],
-                cat_dict={"m_c": cat_from, "r_c": cat_to},
-            )
-        )
-        return True
+    if chance_number == 0:
+        return False
```
j-gynn: what about `disable_random`?

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+            mate_string = _get_mate_string(
+                "rejected",
+                poly,
+                existing_from_cat_mates,
+                existing_to_cat_mates,
+            )
+            confession_changes = get_config("mates.confession.reactions.rejected")
+            cat_from_change, cat_to_change = _get_relationship_change_dict(
+                confession_changes, variability
+            )
```
j-gynn: Too big. Should be its own function (`_try_confess` maybe?) that returns `become_mates, mate_string, rel_changes` (with the latter containing `cat_from_change` and `cat_to_change`.

A new function would stop there being so many if-elses, as you could just early return when needed.

May not have perfectly encapsulated the logic and/or included too much, so use your judgement.

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
 
-        # Gather relationships
-        if cat_to.ID in cat_from.relationships:
-            relationship_from = cat_from.relationships[cat_to.ID]
-        else:
-            relationship_from = cat_from.create_one_relationship(cat_to)
+    if poly and not _check_current_mates_allow_new_mate(
+        cat_from, cat_to, existing_from_cat_mates, existing_to_cat_mates
+    ):
+        return False
```
j-gynn: should be its own function, it's repeated later in `_attempt_mutual_interest_mates`

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
-        VALUES:
-            - 0: no condition
-            - positive number: value has to be higher than number
-            - negative number: value has to be lower than number
-
-        """
-        if not relationship:
+def _check_against_grief(cat_from: Cat, cat_to: Cat) -> bool:
+    """
+    Checks if cat_to will still approach a grief stricken cat_from.
```
j-gynn: what does "approach" mean in this context (/genq)

Does it mean "hey i want to be your mate" or is it "Hey are you OK". term may need revising for clarity.

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+    :return: True if they will approach, False if they won't
+    """
+    if "grief stricken" in cat_to.illnesses:
```
j-gynn:
```suggestion
    if "grief stricken" not in cat_to.illnesses:
        return True
```
then the rest can just keep going without allathat

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+            cat_from.personality.sociability < 8,
+            cat_from.personality.aggression > 8,
+        ]:
+            if threshold_reached:
+                chance += get_config("mates.approach_grief.facet_influence")
+        if random.random() > chance:
             return False
 
-        return relationship.relationship_qualifies(condition)
+    return True
```
j-gynn:
```suggestion
        return random.random() < chance
```
in tandem with the above early-return

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+    return cat_from_change, cat_to_change
+
+
+def _get_mate_string(
+    key: str,
+    poly: bool,
+    cat_from_mates: list[str],
+    cat_to_mates: list[str],
+) -> str:
+    """Returns the mate string with the certain key, cats and poly."""
```
j-gynn: what is the "certain key" /genq

docustring ambiguous, consider revising

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+            poly_key = "m_c_mates"
+        elif not cat_from_mates and cat_to_mates:
+            poly_key = "r_c_mates"
+        if not poly_key:
+            # none of the other involved mates are alive
+            return choice(MATE_DICTS[key])
+        return choice(POLY_MATE_DICTS[key][poly_key])
+
+
+def _prepare_relationship_string(mate_string: str, cat_from: Cat, cat_to: Cat):
```
j-gynn: should live with all the other text adjustment code over in `events_module/text_adjust.py` for consistency. Also should probably be renamed `relationship_text_adjust`, again for consistency

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
-        alive_inclan_to_mates = [
-            mate
-            for mate in cat_to.mate
-            if cat_to.fetch_cat(mate).status.alive_in_player_clan
+        mate_name_string = mate_names[0]
+        if len(mate_names) == 2:
+            mate_name_string = mate_names[0] + " and " + mate_names[1]
+        if len(mate_names) > 2:
+            mate_name_string = ", ".join(mate_names[:-1]) + ", and " + mate_names[-1]
+        mate_string = mate_string.replace("[m_c_mates]", mate_name_string)
```
j-gynn: Use `adjust_list_text()`

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
-        # Then, at least a 1/5 chance
-        chance_number = max(chance_number, 5)
-
-        return chance_number
+        mate_name_string = mate_names[0]
+        if len(mate_names) == 2:
+            mate_name_string = mate_names[0] + " and " + mate_names[1]
+        if len(mate_names) > 2:
+            mate_name_string = ", ".join(mate_names[:-1]) + ", and " + mate_names[-1]
+        mate_string = mate_string.replace("[r_c_mates]", mate_name_string)
```
j-gynn: Use `adjust_list_text()`

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+
+    if "(m_c_mate/mates)" in mate_string:
+        insert = "mate"
+        if len(cat_from.mate) > 1:
+            insert = "mates"
```
j-gynn: unlocalized text, use `i18n.t()` with the appropriate mates label and `count=len(cat_from.mate)`

scripts/events_module/relationship/romantic_events.py
```diff
@@ -19,759 +18,704 @@
...
+        mate_string = mate_string.replace("(m_c_mate/mates)", insert)
+
+    if "(r_c_mate/mates)" in mate_string:
+        insert = "mate"
+        if len(cat_to.mate) > 1:
+            insert = "mates"
```
j-gynn: as above

j-gynn (changes requested):

j-gynn:
A few comments here, and also some conflicts to resolve <3

scripts/events_module/relationship/romantic_events.py
scribblecrumb: I'm gonna be real, this is an annoying bit of code to turn into a function due to how many variables it sets up along with checking if it's poly but mating isn't allowed.  It means the func would be returning a ton of values and I'm not sure that's desirable. I always find it confusing when a func is returning a massive tuple.

scripts/events_module/relationship/romantic_events.py
scribblecrumb: I also hate this btw, for the same reason, but I've done it.

scripts/config.py
scribblecrumb: You're right on me using the wrong one, though you're wrong on `constants.CONFIG` being this file! `constants.CONFIG` is the one in game_structure.

j-gynn (approved):

j-gynn:
Fix your conflicts and merge when you're ready!

PR #5402, opened 2026-07-06
[ENHANCEMENT] Non-recruited Starter Cats become Outsiders

### LINKED ISSUE #4765: [ENHANCE] Non-recruited starter cats persist as outsiders
### Description

Per a conversation on the discord: we could have all the cats that a player *doesn't* initially recruit still persist within the game as outsiders. This provides the Clan with an initial group of outsiders rather than the area just being empty.

Inkpawwc:
## About The Pull Request
Cats that aren't selected to be part of the clan in the ChooseCatsScreen become outsiders.

Gives some starting outsiders that can be fun for storytelling

## Linked Issues

Fixes: #4765 
## Proof of Testing
image
image
image
image

## Changelog/Credits
Non-recruited starter cats now become outsiders

### DIFF
--- a/scripts/screens/make_clan_screens/ChooseModeScreen.py
+++ b/scripts/screens/make_clan_screens/ChooseModeScreen.py
@@ -227,4 +227,14 @@ class ChooseModeScreen(MakeClanScreenBase):
             )
             members.append(create_cat(rank=random_rank))
 
+        for _ in range(randrange(2,4)):
+            random_rank = choice(
+                [
+                    CatRank.KITTYPET,
+                    CatRank.LONER,
+                    CatRank.ROGUE,
+                ]
+            )
+            members.append(create_cat(rank=random_rank))
+
         self.clan_info.starting_members = members
--- a/scripts/screens/make_clan_screens/ChooseModeScreen.py
+++ b/scripts/screens/make_clan_screens/ChooseModeScreen.py
@@ -227,14 +227,9 @@ class ChooseModeScreen(MakeClanScreenBase):
             )
             members.append(create_cat(rank=random_rank))
 
-        for _ in range(randrange(2, 4)):
-            random_rank = choice(
-                [
-                    CatRank.KITTYPET,
-                    CatRank.LONER,
-                    CatRank.ROGUE,
-                ]
-            )
-            members.append(create_cat(rank=random_rank))
+        switch_set_value(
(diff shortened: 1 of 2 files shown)


scripts/screens/make_clan_screens/ChooseModeScreen.py
```diff
@@ -227,4 +227,14 @@ def random_quick_start(self):
...
+        for _ in range(randrange(2,4)):
+            random_rank = choice(
+                [
+                    CatRank.KITTYPET,
+                    CatRank.LONER,
+                    CatRank.ROGUE,
+                ]
+            )
+            members.append(create_cat(rank=random_rank))
+
```
scribblecrumb, line 234: Alright there's an issue here.

You're creating these cats and then appending them to the `starting_members`. I'm, first off, not a huge fan of this, as it muddles what the purpose of `starting_members` is.

Additionally, since the `save_clan` then runs through `Switch.possible_cats` and makes any cats in there that haven't been used in `clan_info` into outsiders, you essentially end up with these 2-4 cats created here *as well* as all 12 `possible_cats` being make into outsiders.  We can agree that's not ideal behavior.

What I would suggest to fix this without us having to greatly rework how `quick_start` interfaces with `Switch.possible_cats` (that is to say, currently not at all), would be to just adjust the `Switch.possible_cats` list during this func.

`Switch.possible_cats` is given its cats in the `screen_switches()` of this class, so by the time we get to this function in particular there's already 12 cats created.  I would have you replace this whole `for` loop with:
```python
switch_set_value(
            Switch.possible_cats,
            switch_get_value(Switch.possible_cats)[: randint(2, 4)],
        )
```
What this would do is essentially "shorten" the `Switch.possible_cats` list to just it's first 2-4 members.  Then when we eventually get to `save_clan()`, there will be an appropriate number of possible cats that will then be made into outsiders.

Sorry for the long comment here lmao, feel free to ask for clarification in the discord if anything feels unclear.

scribblecrumb (changes requested):

scripts/screens/make_clan_screens/ChooseModeScreen.py
Inkpawwc: ty! No need to apologize for the long comments, i appreciate the clarity ^^

scribblecrumb (approved):

PR #5408, opened 2026-07-07
[TASK] Active Card Window

### LINKED ISSUE #5092: [PROJECT] Active Card Display
**Description**
For the Cruel Season update! We'll need a way for players to view their currently active cards.

We could add a new button to the top menu, though we do try to keep that menu pretty pared down and it's a bit frustrating to juggle menu elements which change between game modes. However, this would let it be accessible from most screens.

We could also try adding the button to the Event page specifically. This means it's not accessible from other screens, but it also gives us a bit more space. We could give it extra flair, such as being a clickable deck of cards. 

We could also consider bundling it under the Supplies dropdown. It's been suggested that we rename this dropdown to Resources, perhaps cruel cards could be considered "resources"? However this implies more of a cat-level awareness that I think we're not looking for. The cats themselves shouldn't be "interacting" with cards in the same way they interact with freshkill and herbs.

No matter what we do, the end result could open up a pop-up which displays all current cards. We could use the small symbols for this, perhaps, and have a space for the larger card art to be displayed if a smaller symbol is clicked.  Alternatively, we borrow the same set up as the card choice screen, where all the full cards are displayed and can be sifted through.

**Scope of Project**
Requires a solid understanding of the UI code, including how to handle pop-ups.

**Relevant Links**

scribblecrumb:
## About The Pull Request
Adds a way for players to view the cards they've chosen while playing.

- Added a deck button to the event page
- Clicking the button opens a new window that displays all active cards and their effects

## Linked Issues
closes #5092

## Proof of Testing

https://github.com/user-attachments/assets/bd051a6c-f64a-4d3e-ae79-995cd1fd6294

## Changelog/Credits
- Adds a button to the events page that will open a window displaying all cards active in Cruel Season mode.


j-gynn:
There seems to be some hover shenanigans on the icon - can you confirm whether it behaves consistently for you or not?

scribblecrumb:
> There seems to be some hover shenanigans on the icon - can you confirm whether it behaves consistently for you or not?

It does!  If you're referring to the little jiggle it does when clicking, that's just due to the click switching it to a different state. I've gone ahead and adjusted that now though, so it'll no longer do that.

j-gynn (approved):

PR #5416, opened 2026-07-07
[TASK] For The Throne card

scribblecrumb:
## About The Pull Request
Adds the card Anything For The Throne. This card encourages deputies to kill their leaders through a variety of methods:
- increases chance of negative rel event (gotta get them in the mindset to kill)
- increases chance that the negative rel event will be respect based (now we make them envious of the leader's position)
- if the deputy is looking for murder targets, they'll always prefer the leader over other options
- if the deputy has chosen the leader for murder, they'll receive a big buff to the chance to kill

## Proof of Testing
image
image

had to debug for this guy, but hey we finally hit it! you can see it go from 75 to 45

image

end chance of 25

image

oop we're getting close to a murder

image

that leader made it out unscathed, but this one didn't!

image

chris was not a canela fan by the end

## Changelog/Credits
- Adds the card "Anything For The Throne", which encourages deputies to murder their leader


scripts/events_module/relationship/generate_pair_event.py
```diff
@@ -126,6 +126,9 @@ def _get_type_of_change(
     elif comp == CatCompatibility.NEGATIVE:
         bool_ballot.append(False)
 
+    if get_config("relationship.deputy_to_leader_neg_buff"):
+        bool_ballot.append(False)
```
j-gynn: doesn't this apply for everyone always then?

j-gynn (changes requested):

scripts/events_module/relationship/generate_pair_event.py
scribblecrumb: oop yes! i forgot to fix that lol

j-gynn (approved):

PR #5427, opened 2026-07-08
[TASK] Lone Steps card

scribblecrumb:
## About The Pull Request
Adds the Lone Steps card, which begins the Clan with a single member.

Had to do some finagling to allow this to work with a cat of any age, while still preserving the Cool Aesthetics. I think it's a decently clean implementation.  If the cat *is* old enough to be a leader, then they'll be made into one after they're chosen. Otherwise the cat stays as their "normal" rank.

I also added some new localization to account for the new possible variability in min/max members, as well as a new message to inform the player that the clan is full.

## Proof of Testing
image
image
image
image
image
image

single kitten works

image

single adult works

## Changelog/Credits
- Adds Lone Steps card, which restricts the Clan to a single member at creation. This cat can be any age, even a kitten!
- Added new message displayed when the player has chosen the maximum cats allowed
- Added new localization to allow min/max allowed cats to be displayed correctly
- Added cool new cat head states to Clan creation that will mark which roles *cannot* be filled due to cruel season cards


j-gynn (approved):
very elegant solution

PR #5431, opened 2026-07-09
[TASK] Stolen Vitality card

scribblecrumb:
## About The Pull Request
Adds the Stolen Vitality card, which requires a random clanmate to die for a leader to be revived.

This baby was a bit of a hassle, as it's necessarily more involved. I ended up adding a function to handle all the fiddly bits and just have that being called whenever there's a place where a leader dies. I couldn't just hook it into `.die()` unless we were willing to forget having the result text attached to the inciting event, which I personally don't think is *as* impactful an implementation.

I added a new localization file that will hold any extra text attached to cruel cards for us, this is holding the result and history text for the new card atm.

In the config I decided to just be really blatant and essentially add a "true or false, is this card active" value.  This card is just SO specific in it's behavior that I don't think there's any "general use" config that we could implement.

## Proof of Testing
image
image

grief events for killed cats work!

image

killed cats get a special history

image

manual killed of the leader results in immediate death and a new result text

image

BUNCH of deaths

image

patrols receive appropriate result text

image

if there aren't enough cats to sacrifice, then the leader dies anyways!

image

Special history for the leader!

## Changelog/Credits
- Adds the Stolen Vitality card, which requires a random Clanmate to die for a leader to be revived.


j-gynn (approved):

j-gynn:
fix ur conflicts (again) and we're good

PR #5435, opened 2026-07-10
[DEV] Switch from appdmg

larkgz:
## About The Pull Request

Switches from `appdmg` to `dmgbuild` for building the fancy dmg files on MacOS

* MacOS builds are flakey. Maybe switching tools will fix it? (said in a pleading tone)
* Removes javascript from build process

## Proof of Testing

Current .dmg file created by GitHub Actions:
image

New .dmg file created by GitHub Actions:
image


archanyhm (approved):
I don't think the race condition leading to the CI failure is necessarily related to `appdmg`, but I have never liked it and replacing it is still a good thing.

PR #5438, opened 2026-07-10
[TASK] Stranger Danger Card

Chinch-Bug:
## About The Pull Request
Adds a Stranger Danger card, which permanently locks outsider reputation as hostile

Bundled the min_reputation and max_reputation added to game config into an [outsiders] section with outsider_ex now moved into it
Also cleaned up a lot of the clamping done to the reputation value, as there's a setter that's already handling that. 

Adds an extra bit of challenge by making it difficult to get new cats to join

## Proof of Testing
image
image


scribblecrumb (changes requested):
Looks nice!  We did have some plans for a card of this nature already, so to match with those, could you change the category of this card to Behavior and the name to Malicious Aura? Thanks!

scripts/clan.py
ssuncreature: Just gonna drop some notes as I had been previously working on this card lol

The way you have coded this right now will initially set the reputation to 80, then lower it to the max. How I had handled this in my version was to add another config for the starting value, which would be modified by the cruel season card.

The default would be 80, and then lower to, in mine, 15, as a nice centre point.

scripts/clan.py
```diff
@@ -1295,11 +1298,8 @@ def reputation(self):
 
     @reputation.setter
     def reputation(self, a: int):
-        self._reputation = int(a)
-        if self._reputation > 100:
-            self._reputation = 100
-        elif self._reputation < 0:
-            self._reputation = 0
+        rep = min(int(a), get_config("outsiders.max_reputation"))
+        self._reputation = max(rep, get_config("outsiders.min_reputation"))
```
ssuncreature, line 1303: Have you tested how this would affect gains/losses beyond the minimum and maximum?

Chinch-Bug: image
threatening outsiders focus keeps it at 0 when it was already at 0 previous moons

image
entreating with outsiders sticks it at the stated max 29

scripts/clan.py
```diff
@@ -139,7 +139,7 @@ def __init__(
         # Reputation is for loners/kittypets/outsiders in general that wish to join the clan.
         # it's a range from 1-100, with 30-70 being neutral, 71-100 being "welcoming",
         # and 1-29 being "hostile". if you're hostile to outsiders, they will VERY RARELY show up.
-        self._reputation = 80
+        self._reputation = constants.CONFIG["outsiders"]["starting_reputation"]
```
scribblecrumb: better to use the `get_config()` here! You can pass `self.cruel_cards` into the `get_config()` func to ensure it uses the card list being initialized.

scripts/clan.py
```diff
@@ -301,6 +301,9 @@ def create_clan(self):
             other_clan = OtherClan()
             self.all_other_clans.append(other_clan)
 
+        # required for cruel card effects to be applied to starting value
+        self.reputation = get_config("outsiders.starting_reputation")
```
scribblecrumb: Do we need to do this again? We should be able to set the private value up at line 142 with the `get_config()`

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5441, opened 2026-07-10
[CONTENT] Two more poly mate events!

UnholyMinty:
## About The Pull Request

flowe and pebbles. Adds two more medium poly events for pre-established mates, one flower gathering and one for penguin pebbling!

Adds more silly poly things

## Proof of Testing

It doesn't really add something testable since it's just. Piggybacking off Caine's thingy so if it doesn't work I'm only half responsible (/silly), obviously willing to add proof of testing if need be though

## Changelog/Credits

Adds two new poly


j-gynn (approved):
just for u caine.

PR #5444, opened 2026-07-10
[TASK] Enmity Card

ssuncreature:
## About The Pull Request

- Moves method of handling other clan relations into game_config (relationship ranges, maximum)
- Adds new cruel season card: Enmity. All other clan relations are locked to hostile.

- Was on the spreadsheet! 

## Proof of Testing

Screenshot 2026-07-11 at 10 40 49
Screenshot 2026-07-11 at 10 41 20
Screenshot 2026-07-11 at 10 44 19
^ Start of game

v extended play w/ war events and reputation events
Screenshot 2026-07-11 at 10 42 05
Screenshot 2026-07-11 at 10 42 52
Screenshot 2026-07-11 at 10 43 46

+ Game runs normally on classic regarding other clans
Screenshot 2026-07-11 at 10 47 36
Screenshot 2026-07-11 at 10 49 58


ssuncreature:
jeez the tests

scripts/clan.py
```diff
@@ -1495,11 +1495,15 @@ def get_standing(self) -> Literal["ally", "neutral", "hostile"]:
...
+        # makes sure that other clan relations can't surpass limit
+        if self.relations > get_config("reputation.other_clans.relation_cap"):
+            self.relations = get_config("reputation.other_clans.relation_cap")
+
+        if self.relations <= get_config("reputation.other_clans.hostile"):
+            return "hostile"
+        elif self.relations <= get_config("reputation.other_clans.neutral"):
             return "neutral"
-        return "hostile"  # self.relations < 7
+        return "ally"
```
scribblecrumb: rather than adding the limit check here, I'd love to see it in a theoretical `relations` property!  The issue with having it here is that it doesn't *really* prevent the relations from going too high or too low.  It just "fixes" it after the fact whenever this function is eventually called.  While that does work in theory, it's not super clean in execution.

If you've not working with properties before, you can see a demonstration of them in this class!  `name` is actually a property.  It has a `@property.setter` as well, which is essentially a function that runs whenever the code tries to set `name` as something.  If `relations` was a property, then you could also give it a `setter` function that checks the if the new relation value is surpassing the limit or not.

scribblecrumb (changes requested):

ssuncreature:
okay so attempt one crashed the game lmao - I'll keep trying but might have to wait a bit

scribblecrumb (commented):
Just looking back at the commit with your property work in it, it looks like you wanted to use a private attribute `self._relations` but that you didn't actually initialize `self._relations` in `def __init__`!

ssuncreature:
> Just looking back at the commit with your property work in it, it looks like you wanted to use a private attribute `self._relations` but that you didn't actually initialize `self._relations` in `def __init__`!

sorry, could I get further explanation on that? I'm not very good with properties haha

scribblecrumb (approved):

PR #5446, opened 2026-07-10
ci: add check for stale schemas

archanyhm:
## About The Pull Request

Analogous to the formatter check, same for checking if the schemas are stale.


scribblecrumb:
Could you address the test failure or are we just pushing this as it /gen

archanyhm:
> Could you address the test failure or are we just pushing this as it /gen

Done

scribblecrumb (approved):

PR #5450, opened 2026-07-11
[TASK] Young and Old card

scribblecrumb:
## About The Pull Request
Adds Young and Old card, which starts the Clan with an elder and a kitten.

- Added functionality for specifying what age range is allowed to take each role, which allows us to bar elders from the normal membership.  Thus forcing an elder leader and kitten member.
- Adjusted head display stuff again to make it work right (this is my boulder to roll up the hill over and over) 

## Proof of Testing

image
image
image
image

can't pick a second elder

image
image

## Changelog/Credits
- Adds Young and Old card, which starts the Clan with an elder and a kitten.


scripts/screens/make_clan_screens/ChooseCatsScreen.py
```diff
@@ -524,6 +508,25 @@ def refresh_text_and_buttons(self):
...
+    def _age_valid_for_role(self) -> bool:
+        """
+        Checks if selected cat is valid for the currently required role
+        """
+        if self.need_leader and not self.clan_info.leader:
+            needed_rank = CatRank.LEADER
+        elif self.need_deputy and not self.clan_info.deputy:
+            needed_rank = CatRank.DEPUTY
+        elif self.need_med and not self.clan_info.medicine_cat:
+            needed_rank = CatRank.MEDICINE_CAT
```
j-gynn, line 512: doesn't the representation of this have a space in it? so `medicine cat != medicine_cat`

scripts/screens/make_clan_screens/ChooseCatsScreen.py
```diff
@@ -524,6 +508,25 @@ def refresh_text_and_buttons(self):
...
+        else:
+            needed_rank = "member"
+
+        if self.selected_cat.age in self.get_config_during_creation(
+            f"clan_creation.valid_rank_to_age_assignment.{needed_rank}"
+        ):
+            return True
+        return False
```
j-gynn:
```suggestion
        return self.selected_cat.age in self.get_config_during_creation(
            f"clan_creation.valid_rank_to_age_assignment.{needed_rank}"
        )
```

j-gynn (changes requested):

scripts/screens/make_clan_screens/ChooseCatsScreen.py
scribblecrumb: Both the config key and the `CatRank` string are `medicine cat`. I'm not sure where the issue is?

j-gynn: Ah, sorry. Disregard this. I was confused by `clan_creation.ranks_needed.medicine_cat` having an underscore. Though we really should standardise these at some point so that code-side `medicine cat` is `medicine_cat`. We have the technology.

j-gynn (approved):

scripts/screens/make_clan_screens/ChooseCatsScreen.py
scribblecrumb: Oh I would love for that to happen lmao, it's something that I've been finding more and more annoying lately.

PR #5456, opened 2026-07-11
[TYPO/TWEAKS] minor changes to a mountain thrush training patrol

fulminousnight:
## About The Pull Request

- Increased the frequency of the second generic success to 4 as well so there's a 50/50 chance either option. In my opinion, this makes the patrol more variable instead of feeling as repetitive. 
- Removed the word 'irritably' from one outcome that was assigning emotion to the mentor's temperament instead of letting the player extrapolate on their words based on their personality in game.
- Clarified a relationship log. 

- More clarity & variability in content

## Proof of Testing
Game runs locally, no changes outside of the patrol structure itself.

### DIFF
--- a/resources/lang/en/patrols/mountainous/training/any.json
+++ b/resources/lang/en/patrols/mountainous/training/any.json
@@ -2085,7 +2085,7 @@
                         }
                     }
                 ],
-                "frequency": 1
+                "frequency": 4
             },
             {
                 "text": "s_c stops, suddenly reminded of an old tale. The thrush below is breaking a snail open on a rock. There was a prophecy about that once, although p_l can't quite remember how the story goes...",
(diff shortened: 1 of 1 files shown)


resources/lang/en/patrols/mountainous/training/any.json
```diff
@@ -2166,7 +2166,7 @@
                         ],
                         "amount": -10,
                         "log": {
-                            "cats_from": "cat_from and cat_to squabbled during a training session when app1 wanted to actually hunt instead of practice stalking."
+                            "cats_from": "cat_from ignored cat_to's lesson during training session when app1 wanted to actually hunt instead of practice stalking."
```
tobes03:
```suggestion
                            "cats_from": "app1 ignored p_l's lesson during training session when app1 wanted to actually hunt instead of practice stalking."
```
I noticed this is a mutual change, which means you'll need to use the specific cats rather than "cat_to" and "cat_from"

tobes03 (changes requested):

resources/lang/en/patrols/mountainous/training/any.json
fulminousnight: tobes always picking up what i miss. ty boss

tobes03 (approved):

PR #5466, opened 2026-07-13
Clumsy Paws Cruel Season Card

amilleah:
## About The Pull Request

This adds Clumsy Paws, a Cruel Season card that slows down skill and experience gain with diminishing returns.

Cats gain skills and experience slower, promoting challenging, long-form gameplay over cat lifespans.

## Proof of Testing

30 moons in two clans; each with git commit history to track per-cat skill and XP gain each moon. 

separate plot of cumulative experience and skill gain across all cats for 30 moons, comparing a ControlClan (without Clumsy Paws card) and a TestClan (with Clumsy Paws card). Control clan shows consistently more experience and skill gain across all moons.

Test (with Clumsy Paws) shows noticeable slowdown in progression compared to Control (no Clumsy Paws), more failed patrols as a result of lower XP, and overall noticeably more challenging progression (but not too much!).

| Clan | Patrols | Success | Fail | Success Rate |
|---|---:|---:|---:|---:|
| Control (without Clumsy Paws) | 65 | 54 | 11 | 83.1% |
| Test (with Clumsy Paws) | 62 | 42 | 20 | 67.7% |

### DIFF
--- a/scripts/cat/skills.py
+++ b/scripts/cat/skills.py
@@ -4,13 +4,12 @@ from typing import Union
 
 import i18n
 
+from scripts.config import get_config
 from scripts.cat.enums import CatRank, CatAge
 
 
 def scale_progress(current: float, ceiling: int, amount: float) -> float:
     """adjusts skill/experience gain for difficulty and distance to ceiling"""
-
-    from scripts.config import get_config
 
     modifier = get_config("progress.difficulty_modifier")
     if not modifier or amount <= 0 or ceiling <= 0:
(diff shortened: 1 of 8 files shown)


scripts/cat/skills.py
```diff
@@ -7,6 +7,19 @@
 from scripts.cat.enums import CatRank, CatAge
 
 
+def scale_progress(current: float, ceiling: int, amount: float) -> float:
+    """adjusts skill/experience gain for difficulty and distance to ceiling"""
+
+    from scripts.config import get_config
```
scribblecrumb: does this need to be imported on a func level? we avoid this when possible, so unless you're trying to get around a circular import this should be a script-level import

scribblecrumb (approved):

scribblecrumb (approved):

PR #5482, opened 2026-07-15
[CRUEL CARD] Weakened Stars

arkrylicsun:
## About The Pull Request

- Moves Forsaken into the behaviour card category
- Makes adjustments to the forsaken card art to match the above edit (thank you scribble!)

- Adds Weakened Stars, which lowers the amount of lives leaders get to a random number from 1-5.

- Was in spreadsheet

## Proof of Testing

Weakened Stars:
Screenshot 2026-07-15 at 20 47 29
Screenshot 2026-07-15 at 20 47 48
Leader 1
Screenshot 2026-07-15 at 20 50 50
Screenshot 2026-07-15 at 20 50 57

Leader 2
Screenshot 2026-07-15 at 20 51 20
Screenshot 2026-07-15 at 20 51 30

Forsaken:
Screenshot 2026-07-15 at 20 52 00
Screenshot 2026-07-15 at 20 52 16

Normal clan (no lives cards):
Screenshot 2026-07-15 at 20 53 16
Screenshot 2026-07-15 at 20 53 24


resources/dicts/cruel_season/card_conflicts.json
```diff
@@ -2,5 +2,6 @@
     "clan_ranks": ["all_kittens", "all_apprentices", "all_elders"],
     "leader_life": ["forsaken", "stolen_vitality"],
     "seasons": ["always_newleaf", "always_greenleaf", "always_leaf-fall", "always_leaf-bare"],
-    "litters": ["underpopulated", "mouths_to_feed"]
+    "litters": ["underpopulated", "mouths_to_feed"],
+    "leader_lives": ["weak_stars", "forsaken"]
```
scribblecrumb: let's differentiate `leader_life` and `leader_lives` a little more with the naming.  perhaps `single_leader_life` and `leader_lives`

resources/game_config.toml
```diff
@@ -512,6 +512,12 @@ dev_tools = false
 [death_related]
   # the maximum lives a leader can have
   max_leader_lives = 9
+  
+  # random leader lives
+  leader_lives_random = false
+
+  # minimum only called on if lives are random
+  min_leader_lives = 1
```
scribblecrumb: Rather than a bool, I think it would be better to *just* utilize the min/max leader lives.  By default `min_leader_lives` can be 9, which means that a randint of 9, 9 would be 9

scripts/clan.py
```diff
@@ -199,7 +199,18 @@ def leader_lives(self):
...
+                value, get_config("death_related.max_leader_lives")
+            )
+        else:
+            self._leader_lives = min(
+                value,
+                randint(
+                    get_config("death_related.min_leader_lives"),
+                    get_config("death_related.max_leader_lives"),
+                ),
+            )
```
scribblecrumb: I'm not feeling super good about this method of handling life reset, as it feels a bit like a mixing of uses for this setter (for example, this func is also run when we remove a single life from a leader)

What I'd like to see instead is a new function, perhaps something like:
```py
def reset_leader_lives():
    self._leader_lives = min(
        value,
        randint(
            get_config("death_related.min_leader_lives"),
            get_config("death_related.max_leader_lives"),
        ),
    )  
```

Then, in the parts of the code where we try to "reset" the leader lives to 9, we can instead call `game.clan.reset_leader_lives()`

scribblecrumb (changes requested):

ssuncreature:
Gonna set this to draft while i work on it since i gotta commit it between my computers

ssuncreature:
Screenshot 2026-08-18 at 21 35 58
^ A leader with Weakened stars (second leader)

Screenshot 2026-08-18 at 21 36 53
^ Forsaken (second leader)

Screenshot 2026-08-18 at 21 37 35
^ Normal gameplay (second leader)

ssuncreature:
Not me closing all my prs on accident lol

resources/dicts/cruel_season/card_conflicts.json
```diff
@@ -1,8 +1,9 @@
 {
-    "clan_ranks": ["all_kittens", "all_apprentices", "all_elders", "young_old"],
+    "clan_ranks": ["all_kittens", "all_apprentices", "all_elders"],
+    "singe_leader_life": ["forsaken", "stolen_vitality"],
```
scribblecrumb:
```suggestion
    "single_leader_life": ["forsaken", "stolen_vitality"],
```

resources/lang/en/cruel_season/card_descriptions.en.json
```diff
@@ -24,6 +24,7 @@
     "always_leaf-bare": "The lack of resources that accompany leaf-bare never seem to relent.",
     "underpopulated": "Large litters are less common. How will the Clan continue with so few to carry it's legacy?",
     "heartless": "Survival makes no room for romance. Cats cannot take mates.",
+    "weak_stars": "StarClan's power is faltering. New leaders will only recieve a random number of lives from 1 to 5.",
```
scribblecrumb:
```suggestion
    "weak_stars": "StarClan's power is faltering. New leaders will receive a random number of lives from 1 to 5.",
```

scripts/clan.py
```diff
@@ -460,7 +466,12 @@ def new_leader(self, leader):
             self.leader = leader
             Cat.all_cats[leader.ID].rank_change(CatRank.LEADER)
             self.leader_predecessors += 1
-            self.leader_lives = 9
+            # for some reason calling self.reset_leader_lives() doedsn't work so using randint here
+            # unless another dev is willing to bonk me
+            self.leader_lives = randint(
+                get_config("death_related.min_leader_lives"),
+                get_config("death_related.max_leader_lives"),
+            )
```
scribblecrumb, line 474: In what way was this not working? I switched it to use the function just to see what was happening, but it seems to be working exactly as expected, so I'm not sure what the issue you were running into was.

ssuncreature: If I remember correctly it wasn't finding the return properly, but it's been a while.

I'll give it another go when I get a chance ^^

ssuncreature: Screen Shot 2026-09-11 at 10 25 41

Tried it myself, functions fine

My initial concern that led to me changing it was this error in my code editor

scribblecrumb: Your code editor is right, in a way. You shouldn't be doing
```
self.leader_lives = self.reset_leader_lives()
```
because, indeed, that function does not return a value. All you need to do is call the func, you don't need to assign it to `self.leader_lives` as it's already setting `self.leader_lives` within the function.
```
self.reset_leader_lives()
```
^^ this is all you need

ssuncreature:
apparently i had two branches for this?? lmao??

scribblecrumb (approved):

PR #5502, opened 2026-07-16
[TWEAK] add automatic save conversion for sand/beige eyes

selkirks:
## About The Pull Request

automatically converts beige -> sand in clan_cats

automatic conversion is good tbh

## Proof of Testing

image
image


scribblecrumb:
I'm personally not big on adding save conversion for conversions that are dev version specific (i.e. we aren't converting from stable version info to dev version info. rather we're converting dev version info into dev version info).  we already have a ton of save conversion code to keep track of just to manage stable compatibility.  The less we add to that the better and dev save files have never been promised to be stable or expected to convert cleanly with dev updates.

beige to sand specifically was a change introduce only hours after beige was added, so not many saves were affected in the first place. I would rather this just be a conversion issue that player sort out manually if they've been affected.

PR #5472, opened 2026-07-14
[BUGFIX] strip incorrect ICCP color profiles

rustykitty:
## About The Pull Request
I noticed an increase in the number of incorrect color profile warnings after syncing my fork today, so I decided to try and fix them.
Note: There have always been warnings, there have just been more recently. Also, the warnings have not completely been eliminated.

This PR strips ICCP color profiles and other metadata from image files, to reduce the number of `libpng warning: iCCP: known incorrect sRGB profile` warnings we get.
There were still some files with warning, that may require some manual inspection; that is beyond the scope of this PR and I may do that in the future in a separate PR.
This is the command I ran:
```bash
find -type f -name "*.png" -exec pngcrush -n -q {} \; 2>&1 | grep -A 2 "pngcrush: " | grep '\./' | tr -d ' :' | xargs -P0 -n1 mogrify -strip
```

## Why This Is Good For ClanGen
N/A

## Proof of Testing
`development`:
image
This branch:
image

PR #5520, opened 2026-07-17
[Tool] oxipng justfile

archanyhm:
## About The Pull Request

Adds a Justfile (see https://github.com/casey/just) with a target for running [oxipng](https://github.com/oxipng/oxipng) on all PNGs in the repo, stripping them of needless metadata (see #5472) as well as reducing their size. Running this regularly should cut down the PNG sizes by about 20ish percent and doesn't degrade the image quality.

The relevant file to review is the `justfile`, since the rest are just image updates.


larkgz:
Is there a way to do this with just a command using Python, like the schema updates?

I imagine that many of the devs “creating” warnings would be artists on Windows computers, so I think it would be best if they could run the command without needing to manually install other dependencies.

archanyhm:
> Is there a way to do this with just a command using Python, like the schema updates?
> 
> I imagine that many of the devs “creating” warnings would be artists on Windows computers, so I think it would be best if they could run the command without needing to manually install other dependencies.

In my opinion, there's no need for every contributor to be able to run this step easily. This is mostly a tool for us Senior Developers to use for asset optimization, e.g. before releasing a new version, since the savings in size are relatively small and the iCCP-warnings are purely logging-noise and don't actually impact the asset handling whatsoever to my knowledge. Using `oxipng` as well-established and tested tool feels more appropriate as opposed to a Python wrapper (`pyoxipng` and `oxipng_py` exist, but they don't expose all functionalities 1:1 and are less widely adapted, so I'm less inclined to trust them). As a compromise, we could eventually add a bootstrapping command to the Justfile that installs the necessary tools though?

larkgz:
> As a compromise, we could eventually add a bootstrapping command to the Justfile that installs the necessary tools though?

Ehh, if it's just for Senior Developers to run, then I don't think it's super necessary.

Could you maybe also add documentation somewhere about the script's dependencies and what it does? Maybe under https://clangen.io/docs/dev/senior/?

archanyhm:
> > As a compromise, we could eventually add a bootstrapping command to the Justfile that installs the necessary tools though?
> 
> Ehh, if it's just for Senior Developers to run, then I don't think it's super necessary.
> 
> Could you maybe also add documentation somewhere about the script's dependencies and what it does? Maybe under [clangen.io/docs/dev/senior](https://clangen.io/docs/dev/senior/)?

Done

larkgz (approved):

PR #5524, opened 2026-07-18
[BUGFIX] Random cat selection for Lone Steps Card

### LINKED ISSUE #5511: [BUG] Crash when selecting random cats with lone steps
### Pre-requisites

### Description

When the cruel card Lone Steps is active, using the "choose random cats" button causes the game to crash.

### Grade

### Gamemode

Cruel Season

### Challenge cards

Lone Steps

1. Create a new clan
2. Select cruel season and select the Lone Steps card
3. Use the "choose random cats" button
4. Game crashes

_No response_

unkown

### Error log

_No response_

### Additional context

Commit # or Game Version Number only shows as unknown my apologies. I downloaded it right after this pull if that helps at all:
[TWEAK] changes beige eye name to sand brown #5496

first time doing anything like this sorry!

### Screenshots

_No response_

amilleah:
## About The Pull Request

This fixes a bug where Lone Steps cat selection with random button creates an invalid range and crashes the game.

## Issues

Fixes #5511

## Proof of Testing

Replicated bug, added check to range, attempted to replicate bug again, no crash:

choosecatscreen, I used random cat selection here, no crash

### DIFF
--- a/scripts/screens/make_clan_screens/ChooseCatsScreen.py
+++ b/scripts/screens/make_clan_screens/ChooseCatsScreen.py
@@ -1,4 +1,4 @@
-from random import choice
+from random import choice, randint
 from typing import Optional
 
 import i18n
@@ -339,15 +339,7 @@ class ChooseCatsScreen(MakeClanScreenBase):
         maximum_needed = self.get_config_during_creation(
             "clan_creation.maximum_membership"
         ) - len(self.clan_info.get_all_cats())
-        for i in range(
-            1,
-            choice(
-                range(
-                    minimum_needed + 1,
-                    maximum_needed,
-                )
-            ),
-        ):
+        for _ in range(randint(minimum_needed, maximum_needed)):
             self.clan_info.starting_members.append(
                 choice(
                     [


scripts/screens/make_clan_screens/ChooseCatsScreen.py
```diff
@@ -339,14 +339,17 @@ def choose_random_cats(self):
...
-                range(
-                    minimum_needed + 1,
-                    maximum_needed,
-                )
-            ),
+        range_needed = range(
+            minimum_needed + 1,
+            maximum_needed,
+        )
+        for i in (
```
larkgz:
```suggestion
        for _ in (
```
Since this variable isn't used, it's better to use an underscore to make it clear.

scripts/screens/make_clan_screens/ChooseCatsScreen.py
```diff
@@ -339,14 +339,17 @@ def choose_random_cats(self):
...
+            range(
+                1,
```
larkgz: It's not necessary to specify a starting number for this because the iterator isn't being used. Also, using `1` for the first argument is why `minimum_needed + 1` needed to be used instead of just `minimum_needed`.

scripts/screens/make_clan_screens/ChooseCatsScreen.py
```diff
@@ -339,14 +339,17 @@ def choose_random_cats(self):
...
+                choice(range_needed),
+            )
+            if range_needed
+            else range(1, minimum_needed + 1)
```
larkgz: I find this a little bit difficult to understand. I think a better approach would be to just take `range(random.randint(minimum_needed, maximum_needed))`.

This works because `randint` is inclusive of both sides of the interval, so it still works when both values are the same.

larkgz (changes requested):

amilleah:
random selection of cats with lone steps card
random selection of cats on expanded mode

larkgz (approved):

PR #5542, opened 2026-07-19
[BUGFIX] ClanClan bugs, remove hardcoded instances

amilleah:
## About The Pull Request

This replaces a set of `game.clan.name` with `game.clan.prefix` to prevent "-ClanClan" bugs!

## Linked Issues

#1818 comments

## Proof of Testing

Allegiances screen:
Screenshot showing Allegiance screen's header correctly uses [Player]Clan now

Apprentice retirement from permanent condition (had to use debug to force permanent condition + retirement):
Screenshot showing Apprentice retirement to Elder's den from permanent condition correctly uses [Player]Clan now

Leader role description:
Screenshot showing role change screen's leader description correctly uses [Player]Clan now


scribblecrumb (changes requested):
Where possible, I would prefer if the localized strings simply weren't appending "Clan" to the end!  We don't want to be using `.prefix` if we can help it.

scribblecrumb (approved):

PR #5580, opened 2026-07-26
[UI] Update and remove outdated button images

Inkpawwc:
## About The Pull Request
updates some UIImageButtons to UISurfaceImageButtons 

Removes unused image files from image folder, making it easier to sort through

## Proof of Testing
image
image
image
image
image


archanyhm:
Is there a reason why the removal of the images is in this PR and not in a separate one? Feels like there's two things going on in here.

Inkpawwc:
I was just gonna remove the images but realized in some spaces we were using them where we didn't need to be so I changed it since I was already there lol. I can separate them into 2 PRs if I need to

scribblecrumb (approved):

PR #5609, opened 2026-08-01
Bugfix: prevent crash when randomizing a custom Clan name

### LINKED ISSUE #5596: [BUG] crash when randomizing clan name after typing custom one
### Pre-requisites

### Description

on the "clan creation" screen, after getting to the area where cats are picked, clicking "previous step", and clicking the button to randomize a clan prefix, my game crashed and i was shown the error log attached below.

i opened the game and tried to replicate this error again, using even the exact same cruel season cards that i had chosen before the crash happened, but it seems that the crash may only happen occasionally.

### Grade

### Gamemode

Cruel Season

### Challenge cards

- anything for the throne
- stolen vitality
- lean meat
- mouths to feed (or underpopulated, i don't remember exactly.)
- barren stockpiles
- lone steps
- public enemy

1. create clan
3. write a custom name for the clan
5. randomize clan name
6. game crash

91b1af6b67695a553c6522ae190a144808072201

### Error log

Image

### Additional context

_No response_

### Screenshots

_No response_

Sanjays2402:
## About The Pull Request

`MakeClanScreenBase.random_clan_name()` removed the current `display_name` from the possible-names list to avoid rolling the same name twice. A name the player typed themselves is not in that list, so `list.remove()` raised `ValueError` and crashed Clan creation as soon as the randomize button was pressed after a custom name had been entered.

The list is now filtered instead of mutated, which keeps the no-repeat behaviour and is safe for any `display_name`.

Bug fix.

## Linked Issues

Fixes: #5596

## Proof of Testing

Added `tests/test_make_clan_screen.py`:

- `test_custom_display_name_does_not_crash` — asserts the custom name is genuinely absent from `get_possible_clan_names()`, then calls `random_clan_name()` and checks a valid name comes back.
- `test_random_name_differs_from_current_name` — 20 rolls with an existing name set, none repeat it.

Both fail on the current code (`ValueError: list.remove(x): x not in list`) and pass with the fix; verified locally by stashing the source change and re-running. `black` is clean on both files.

## Changelog/Credits

Bugfix: fixed a crash when randomizing the Clan name after typing a custom one.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/scripts/screens/make_clan_screens/MakeClanScreenBase.py
+++ b/scripts/screens/make_clan_screens/MakeClanScreenBase.py
@@ -274,9 +274,13 @@ class MakeClanScreenBase(Screens):
         return chosen_biome
 
     def random_clan_name(self):
-        clan_names = get_possible_clan_names()
-        if self.clan_info.display_name:
-            clan_names.remove(self.clan_info.display_name)
+        # ensuring that the new random name will not be the same one
+        # a custom name typed by the player isn't in the list, so filter rather than remove
+        clan_names = [
+            name
+            for name in get_possible_clan_names()
+            if name != self.clan_info.display_name
+        ]
 
         return choice(clan_names)
 
--- a/tests/test_make_clan_screen.py
+++ b/tests/test_make_clan_screen.py
@@ -1,0 +1,35 @@
+import os
+import unittest
+from types import SimpleNamespace
+
+os.environ["SDL_VIDEODRIVER"] = "dummy"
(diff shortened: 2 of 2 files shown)


Vickarooni:
From our README:
"Issues and Pull Requests created with AI based tools are going to be closed without further comment. Repeat offenders will be blocked from this project until further notice."

If you read this, you'd know we are strictly anti-AI and do not accept PRs that involve it in any way.

PR #5612, opened 2026-08-01
[BUGFIX] Fix space being applied to filter constraints

scribblecrumb:
## About The Pull Request
The preliminary filter check was the one adding the space. So annoying.

## Changelog/Credits
- Fixes crashed caused by a space being adding incorrectly to tags


scripts/events_module/event_filters.py
```diff
@@ -1598,6 +1598,7 @@ def filter_relationship_type(group: list, filter_types: List[str], patrol_leader
     exclusionary_values = []
     inclusionary_values = []
     for value in filter_types:
+        value.strip()
```
larkgz: Can you add a comment explaining this fix? Otherwise, someone could remove it without realizing why it's there.

larkgz (changes requested):

larkgz (approved):

PR #5625, opened 2026-08-02
[ENHANCEMENT] Deck list input

### LINKED ISSUE #5526: [ENHANCE] Cruel card decklist
### Description

It would be helpful if users could somehow input a text "decklist" when selecting cruel cards, which would then select all the cards in the decklist. 

I think it would make it a lot easier for people doing challenges, as they could easily copy and paste a whole set of cards.

scribblecrumb:
## About The Pull Request
Adds a new window to allow the user to input a list of cruel season cards to select.  These can be added to a current selection, but have been barred from introducing conflicts or going over the card limit.  Unfortunately the conflict error warning isn't as robust as the normal one featured when choosing cards, but there really wasn't an easy way to handle them beyond just telling the player that their list is invalid and leaving it for them to figure out.  Hopefully players are sharing deck lists with each other that *don't* come with conflicts anyways.

Input can be the internal card name or the display name (for whatever language the player is currently using).

Makes it easy for players to share certain challenge configurations with each other!

## Linked Issues
closes #5526

## Proof of Testing

https://github.com/user-attachments/assets/39521256-24c5-4475-932b-6cb42d52d713

## Changelog/Credits
- Adds a new button to the card choosing screen while making a cruel season Clan. This button opens a window that allows you to input a list of cards to select them all at once. The intention is to make it easy for players to "share" decks with the community that can be copy-pasted into this input window! The input accepts both the internal card names (the ones seen in your save files) as well as the display names you see in game (though I would encourage players to share lists that use the internal names, as those will never change. the display names *might* change at some point and then make a list invalid)

### DIFF
--- a/scripts/ui/elements/surface_image_button.py
+++ b/scripts/ui/elements/surface_image_button.py
@@ -8,6 +8,7 @@ from pygame_gui.core.interfaces import IUIManagerInterface
 
 from scripts.game_input import INPUT_ACTION_PRESSED, Action, INPUT_ACTION_RELEASED
 from scripts.ui.elements.text_box_tweaked import UITextBoxTweaked
+from scripts.ui.icon import Icon
 from scripts.ui.scale import ui_scale_value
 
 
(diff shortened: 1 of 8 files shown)


scripts/ui/elements/surface_image_button.py
```diff
@@ -129,6 +130,9 @@ def __init__(
                 text_rect = pygame.Rect(
                     relative_rect[0], relative_rect[1], relative_rect[2], -1
                 )
+            if self.text in Icon:
+                text_rect.y -= ui_scale_value(1)
+
```
scribblecrumb, line 135: As a note to the reviewer: This is to fix a small icon offset I noticed happens to the icon tabs. It was just enough of an offset to annoy me.

Ryos00 (approved):

PR #5639, opened 2026-08-03
[REWORK] CatFactory

j-gynn:
## About The Pull Request

Part of a rework that is pulling some of the responsibility off of `Cat` as a class.

### DEPRECATES CSV SAVES
(yeah I forgot to mention this, this PR will deprecate CSV loading because it was already horrifically bugged and I didn't know how to fix it)

Introducing CatFactory! There are three of these (four if you count the one that does nothing yet) and their whole purpose in life is to do all the fiddly bits of instantiation so that `Cat` just has to focus on actually doing `Cat` stuff.

- LoadCatFactory is called when loading a save file. It creates a cat from the saved data. No RNG here, except in a tiny bit for save file conversion. Ideally, I want to get this stuff out of LoadCatFactory but not yet.
- NewCatFactory handles anything with RNG. This creates a cat from scratch using all the mechanics we're used to.
- TestCatFactory subclasses NewCatFactory and just. Replaces all the RNG with known, unambiguous data. No randomness here!
- FadedCatFactory is currently not in use, but it will be usable to build a cat that is faded (and therefore has less fields).

- Makes tests more reliable by removing RNG unambiguously from testing cats
- Makes the `Cat` file easier to read and reduces its scope by removing the initialisation logic
- Clears the way for other reworks in the future (like removing every single cat from inside `Cat`...)

## Proof of Testing

This is hard to document, as nothing has actually changed? But I've run the bulk timeskip a lot and done some of my own moonskips and nothing seems to have exploded.


scripts/cat/factories/base_factory.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from random import Random
```
scribblecrumb: This appears to be unused? Is it necessary?

scripts/cat/factories/create_example_cat.py
```diff
@@ -0,0 +1,63 @@
...
+
+from random import randint, sample, choices
+from typing import TYPE_CHECKING
+
+from scripts.cat.enums import CatRank, CatAge
+from scripts.cat.factories.enums import CatType
+from scripts.cat.factories.new_cat_factory import NewCatFactory
+from scripts.cat.factories.test_cat_factory import TestCatFactory
+from scripts.cat.pelts import Pelt
+from scripts.game_structure import game
```
scribblecrumb: clean up ur unused imports beloved <3

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,359 @@
...
+
+        pelt = cls._build_pelt(kwargs=kwargs)
+
+        gender = GenderDict(
+            sex=kwargs["gender"],
+            genderalign=kwargs.get("gender_align", kwargs["gender"]),
+            pronouns=kwargs.get("pronouns"),
+        )
+
+        # todo do I want to do this this way
```
scribblecrumb: do you?

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,359 @@
...
+        if eye_color in ["BLUEYELLOW", "BLUEGREEN"]:
+            # splits into BLUE and either YELLOW or GREEN
+            return eye_color[:4], eye_color[4:]
+        if eye_color2 == "BLUE2":
+            eye_color2 = "COBALT"
+
+        return eye_color, eye_color2
+
+    @classmethod
+    def _build_pelt(cls, kwargs):
```
scribblecrumb: return typehinting?

scripts/cat/factories/load_cat_factory.py
```diff
@@ -0,0 +1,359 @@
...
+        elif outside and not status.is_outsider:
+            status.become_lost()
+
+        if driven_out:
+            status.change_group_nearness(CatGroup.PLAYER_CLAN_ID)
+
+        return status
+
+    @staticmethod
+    def _convert_eye_color(eye_color, eye_color2) -> Tuple[str, Optional[str]]:
```
scribblecrumb, line 192: return typehinting?

scripts/cat/factories/new_cat_factory.py
```diff
@@ -0,0 +1,321 @@
...
+        if isinstance(mate, str):
+            mate = [mate]
+
+        cat_params = {
+            "ID": cls.get_free_id(),
+            "gender_dict": gender_dict,
+            "pelt": pelt,
+            "moons": moons,
+            "status": status,
+            "backstory": overrides.get("backstory", "clanborn"),
```
scribblecrumb: I know someone (amilleah?) had just implemented a fix meant to assign the "default" backstory dependent upon the cat's social status.  so clanborn would be for clancats, a rogue backstory for rogues, ect.  I'd prefer if we preserved that functionality.

scripts/cat/factories/test_cat_factory.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    @classmethod
+    def _get_random_experience(cls, age, moons: int) -> int:
+        return 0
+
+    @classmethod
+    def _get_random_skills_dict(cls, rank, age):
+        return CatSkills(
+            primary_path=SkillPath.OMEN,
```
scribblecrumb, line 57: do u wanna implement that test skill we spoke about or do u wanna do that in some other PR?

scripts/cat/cats.py
```diff
@@ -183,179 +189,74 @@ def __init__(
...
+        self.prevent_fading = toggles["prevent_fading"]  # Prevents a cat from fading
+        self.favourite = toggles["favourite"]
+
+        # misc
+        self.experience = experience
+        self.birth_cooldown = birth_cooldown
+        self.specsuffix_hidden = specsuffix_hidden  # kill this ASAP
+
+        # other misc
+        self.name: Optional[Name] = None
```
scribblecrumb, line 243: should you really be allowed to initialize a cat without a name?

scribblecrumb (changes requested):

scripts/cat/cats.py
j-gynn: Ah... name. `Name` is very fucking stupid. You have to create `Name` after the cat, because `Name` itself contains a copy of the `Cat` object. So alas, `Name` (for now) has to be None to start with, and assigned afterwards.

scripts/cat/factories/test_cat_factory.py
j-gynn: I think the scope of this one is too big? But I could do if you want.

scripts/cat/factories/load_cat_factory.py
j-gynn: Oops. Honestly, no idea. I wrote that comment in February. I think it's okay? I do want Lago to help me make some pydantic models for allathat, but that's in the future.

scripts/cat/factories/load_cat_factory.py
j-gynn: I'm not sure what you're asking here </3

scripts/cat/cats.py
scribblecrumb: sigh, very well

scripts/cat/factories/test_cat_factory.py
scribblecrumb: naw thats fair

scribblecrumb (approved):

PR #5645, opened 2026-08-04
[BUGFIX] "customN" text in custom pronoun

### LINKED ISSUE #5584: [BUG] "custom" string added at end of custom pronouns
### Pre-requisites

### Description

i noticed a bug while adding in it/its pronouns to my save where, if you create any new pronoun set in the "change [cat]'s gender" menu, it will have an additional "custom" string followed by a number, which increases per custom pronoun set that is created. 

this bug does not appear to create any functional issue within the game.

### Grade

### Gamemode

Cruel Season

### Challenge cards

- stolen vitality
- lean meat
- survival of the fittest
- fated paths
- malicious aura

1. open any cat's profile
2. go to "specify gender" under the personal tab
3. click "add new pronouns" and edit in any new pronoun set
4. save the pronoun set
5. look at "saved pronouns" menu (may need to use shorter pronouns or add the pronouns to "current pronouns" to see bug)
6. see bug

_No response_

6d7d9e7801b456086ba476da2fc9530e7cdf65b3

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

rustykitty:
## About The Pull Request
Fix the custom0, custom1, ... text (the custom pronoun ID) appearing in the pronoun display, by listing each item separately instead of naively using a join.

N/A

## Linked Issues
Fixes #5584

## Proof of Testing
image

### DIFF
--- a/scripts/screens/ChangeGenderScreen.py
+++ b/scripts/screens/ChangeGenderScreen.py
@@ -529,13 +529,15 @@ class ChangeGenderScreen(Screens):
 
     def pronoun_get_cases(self, pronounset) -> str:
         # Gets all pronoun cases in pronounset for display
-        displayname = "/".join((
-            pronounset["subject"],
-            pronounset["object"],
-            pronounset["poss"],
-            pronounset["inposs"],
-            pronounset["self"],
-        ))
+        displayname = "/".join(
+            (
+                pronounset["subject"],
+                pronounset["object"],
+                pronounset["poss"],
+                pronounset["inposs"],
+                pronounset["self"],
+            )
+        )
         return displayname
 
     def reset_buttons_and_boxes(self):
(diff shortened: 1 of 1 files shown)


j-gynn (changes requested):
We can't do this for localisation reasons. Not every language has the same noun declensions as English - some have more, some have less. If you can come up with a way that allows us to keep the naive option without the `customN`, I'd appreciate it!

rustykitty:
> We can't do this for localisation reasons. Not every language has the same noun declensions as English - some have more, some have less. If you can come up with a way that allows us to keep the naive option without the `customN`, I'd appreciate it!

Ideally we'd have another JSON. I chose to do it this way only because the pronoun screen currently, when creating a new pronoun, has the names of pronouns (e.g. subject, object, ...) hardcoded.

j-gynn:
What I'm saying is that this PR is directly undoing #5393 . So it needs to change.

rustykitty:
In that case, I'll open another PR implementing the JSON changes I was thinking about when I get home.

PR #5704, opened 2026-08-10
[UI] Separate checkmark from checkboxes

Inkpawwc:
## About The Pull Request
Separates the checkmark that appears over the checkbox when it's checked into its own png and overlays that checkmark on top of the blank checkbox image instead.

Allows us to have the checkbox be one button type (the small box) instead of using 6 images for each variant of the checkbox (i'm comin for that exit_window button next, so with that it'll be 9 images down to 3 yay)

## Proof of Testing
image
image


scripts/ui/elements/checkbox.py
```diff
@@ -52,25 +51,44 @@ def __init__(
...
+            ui_scale_dimensions((34, 34)),
+        )
+        # Creates the checkmark image that gets layered on top of the checkbox when checked
+        self.checkmark = UIModifiedImage(
+            ui_scale(relative_rect),
+            checkmark,
+            container=container,
+            manager=manager,
+            starting_height=3,
+            visible=False,
```
scribblecrumb: This needs to be `visible=self.checked`, otherwise it won't reflect the `check` arg

scripts/ui/elements/checkbox.py
```diff
@@ -52,25 +51,44 @@ def __init__(
...
+            anchors=anchors
+        )
+
```
scribblecrumb, line 68: you also need to add `self.checkmark.disable()` after you make the checkmark. disabling means it won't block the underlying button from responding to hover inputs!

scripts/screens/SpriteInspectScreen.py
```diff
@@ -319,43 +320,47 @@ def update_checkboxes(self):
...
-            "override_not_working",
-            self.override_not_working,
-            self.the_cat.not_working(),
-            disabled_object_id="@checked_checkbox",
+        self.checkboxes["show_as_healthy"] = UICheckbox(
+            position=(400, 625),
+            manager=MANAGER,
+            check=self.override_not_working,
         )
+        if not self.the_cat.not_working:
```
scribblecrumb: `if not self.the_cat.not_working()` you missed the ending func parentheses, that's why it wasn't disabling!

scripts/screens/SpriteInspectScreen.py
```diff
@@ -107,15 +108,15 @@ def handle_event(self, event):
...
                 self.make_cat_image()
                 self.update_checkboxes()
-            elif event.ui_element == self.checkboxes["override_dead_lineart"]:
+            elif event.ui_element == self.checkboxes["show_as_living"]:
                 if self.override_dead_lineart:
                     self.override_dead_lineart = False
                 else:
                     self.override_dead_lineart = True
 
                 self.make_cat_image()
```
scribblecrumb, line 117: So rather than calling the `update_checkboxes` func for all these checkbox events, you should really just do `self.checkboxes["show_as_living].toggle()`.  the update_checkboxes func kills and remakes all the boxes, which technically works, but is very inefficient when we could just use the toggle function that UICheckbox has built into it.

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5707, opened 2026-08-10
[CODE] move outsider_events.py

rustykitty:
## About The Pull Request
Move `scripts.events_module.outsider_events` to `scripts.events_module.relationship.outsider_events`, deleting the file `scripts.events_module.relationship.outsider_events` containing some hardcoded strings (which are unused completely, I checked) in the process.

Move of outsider_events was to place it in the same directory as the other event files. I know it says relationship, but pregnancy_events doesn't seem to be related to rel either.

## Proof of Testing
Bulk skip OK
image


scribblecrumb:
I don't think `outsider_events.py` really belongs in `relationships`.  The pregnancy script is because cats necessary have some sort of relationship in order to do the pregnancy (though it's on thin ice with me lol, i don't much like it being there either.)  Meanwhile, `outsider_events` is handling outsiders dying, which really doesn't have to do with relationships.

scribblecrumb (approved):

PR #5712, opened 2026-08-11
[CLEANUP] remove legacy `clan.txt` code

rustykitty:
## About The Pull Request
Removes the code for loading TXT clans. CatFactory PR (#5639) deprecated CSV/TXT clan loading but didn't fully remove the code.

FYI: clanlist.txt removal is still retained 

Not sure what to do about the tests for now...

## Proof of Testing
Modern clans are loading fine:
image


scripts/clan.py
```diff
@@ -621,7 +617,10 @@ def load_clan(self):
         elif os.path.exists(
             get_save_dir() + "/" + switch_get_value(Switch.clan_list)[0] + "clan.txt"
         ):
-            self.load_clan_txt()
+            switch_set_value(
+                Switch.error_message,
+                "TXT Clans are no longer supported. Please use an external tool to update your Clan to the modern format.",
```
keyraven, line 622: Can you change this to be more informative, and include the clan name? Something like: 

"Your saved clan {CLANNAME} is in an format that is no longer supported. Your save is in TXT format. Please use an external tool to update your Clan to the modern format."

keyraven (approved):
Looks good! I have one suggestion about the error message, but it's not a necessary change.

scripts/clan.py
rustykitty: I believe this is just copied from the clan cats load function, but sure!

PR #5730, opened 2026-08-13
[BUGFIX] Tortie combos not generating naturally

scribblecrumb:
## About The Pull Request
I realized I'd forgotten to add the tortie combos to the tortie patch list!  So while you could save file edit a new combo onto a cat, it couldn't actually generate naturally.

## Proof of Testing
image

debug is showing the combos are possible patch choices!

## Changelog/Credits
- Fixes tortie combo patches not generating on cats naturally

### DIFF
--- a/scripts/cat/pelts.py
+++ b/scripts/cat/pelts.py
@@ -94,7 +94,7 @@ class Pelt:
     )
 
     # TORTIE PATCHES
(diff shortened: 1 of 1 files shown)


keyraven (approved):
Looks good to me!

PR #5736, opened 2026-08-14
[CLEANUP] pregnancy events

scribblecrumb:
## About The Pull Request
Does a bunch of refactoring in pregnancy_events.py. I focused mostly on condensing code, removing unnecessary func arguements, cleaning up extraneous conditions (like checking len to see if a list was empty instead of just checking if it's a nonetype), and breaking some massive functions up into smaller ones.  

There's definitely still work that could be done, especially, imo, if we declassify and break this up into multiple module scripts instead of having it all in a single monster script. But I wanted to keep the scope here smaller and more reviewable, so I figured that should wait.

Makes this script easier to read.

## Proof of Testing
Beyond tests passing, I've also done some manual testing to see if the new affair/coparenting features are intact.

image
image
image

(negative relationship towards kittens they've rejected)

## Changelog/Credits
- Some refactoring of the pregnancy_events.py script to make it more readable

### DIFF
--- a/scripts/events_module/relationship/pregnancy_events.py
+++ b/scripts/events_module/relationship/pregnancy_events.py
@@ -128,13 +128,7 @@ class Pregnancy_Events:
 
         # check if the second_parent is not none and if they also can have kits
         can_have_kits, kits_are_adopted = Pregnancy_Events.check_second_parent(
-            cat,
-            second_parent,
-            get_clan_setting("single parentage"),
-            get_clan_setting("unmated parentage"),
-            get_clan_setting("affair"),
-            get_clan_setting("same sex birth"),
-            get_clan_setting("same sex adoption"),
+            cat, second_parent
         )
         if second_parent:
             if not can_have_kits:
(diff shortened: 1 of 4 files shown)


scripts/events_module/relationship/pregnancy_events.py
```diff
@@ -257,97 +253,81 @@ def biggest_family_is_big():
...
-            not other_cat.status.alive_in_player_clan or other_cat.birth_cooldown > 0
+            not other_cat.status.alive_in_player_clan or other_cat.birth_cooldown
         ):
             return
 
-        if cat.ID in clan.pregnancy_data:
-            return
-
-        if other_cat and other_cat.ID in clan.pregnancy_data:
+        # if the parents are already expecting, then they don't need to adop
```
Ryos00: adop

Ryos00 (approved):

scripts/events_module/relationship/pregnancy_events.py
scribblecrumb:
```suggestion
        # if the parents are already expecting, then they don't need to adopt
```

PR #5737, opened 2026-08-15
[UI] Separate role icons from their background

Inkpawwc:
## About The Pull Request
separates all the role icons from the background frame they sit on and layers them on top in the code instead. I also renamed the role icons as "icon_[role]" so that they stay grouped together in the image folder.

more unified ui and better organization for the image folder (i'd like to try turning the role icons into font icons at some point but that's a future project, this would just make that transition a bit easier too)

## Proof of Testing
image
image


scripts/screens/RoleScreen.py
```diff
@@ -329,6 +329,16 @@ def update_selected_cat(self):
...
+            starting_height=2,
+        )
+
+        self.selected_cat_elements["role_icon_frame"] = pygame_gui.elements.UIImage(
+            ui_scale(pygame.Rect((82, 231), (78, 78))),
+            pygame.transform.scale(
+                image_cache.load_image("resources/images/buttonrank.png"),
+                ui_scale_dimensions((78, 78)),
+            ),
+            starting_height=1,
```
scribblecrumb, line 341: Have you tried using the generated "dark round box"? It's not exactly the same, but it's close enough for me to let it slide

scribblecrumb (changes requested):

scribblecrumb (approved):

PR #5748, opened 2026-08-16
[DEVTOOL] White patch tool

scribblecrumb:
## About The Pull Request
Adds a tool for creating white patch combos!

- Created a base combo tool class so that shared code between the two combo tools wouldn't need to be repeated
- Added a new directory specifically for tool windows, just for organization
- Added pose selection, combo selection, and patch clear functionality to tortie tool
- Added white patch tool, which includes all the same functionality as the tortie tool. It displays all the different white patches, separated by coverage category. It also requires the user to select a coverage category for the new combo.
- Made small adjustments to scrolling button list and scrolling dropdown classes to make them more functional
- Both tools are now capable of modifying existing combos

There is *one* small bug that's a result of limitations in how we handle the pelt class.  When you create a new white combo, it'll display as an error sprite in the combo preview until the game is reloaded. It think this is an acceptable enough bug, tbh, it doesn't really reduce functionality of this tool.

Makes it easier to add new combos and view the current ones!

## Proof of Testing

https://github.com/user-attachments/assets/c79d555a-e36a-4d9b-b1bb-34a82ab12c4f

## Changelog/Credits
- Created a base combo tool class so that shared code between the two combo tools wouldn't need to be repeated
- Added a new directory specifically for tool windows, just for organization
- Added pose selection, combo selection, and patch clear functionality to tortie tool
- Added white patch tool, which includes all the same functionality as the tortie tool. It displays all the different white patches, separated by coverage category. It also requires the user to select a coverage category for the new combo.
- Made small adjustments to scrolling button list and scrolling dropdown classes to make them more functional
- Both tools are now capable of modifying existing combos


scripts/screens/StartScreen.py
```diff
@@ -258,6 +261,17 @@ def screen_switches(self):
...
+                ui_scale(pygame.Rect((15, 15), (200, 30))),
+                "buttons.white_patch",
+                image_dict=get_button_dict(ButtonStyles.MAINMENU, (200, 30)),
+                object_id="@buttonstyles_mainmenu",
+                manager=MANAGER,
+                anchors={
+                    "top_target": self.elements["tortie_patch"],
+                    "left_target": self.elements["continue"],
+                },
+            )
```
j-gynn, line 274: does this mean this button is always visible? I'm not seeing any code to rely on a config option to show/hide it

scripts/ui/windows/dev_tool_windows/white_patch_tool.py
```diff
@@ -0,0 +1,313 @@
...
+            ].child_button_dicts.items():
+                if button == event.ui_element:
+                    self.set_preview_cat(name)
+                    return
+            for name, button in self.elements[
+                "mostly_patch_choice"
+            ].child_button_dicts.items():
+                if button == event.ui_element:
+                    self.set_preview_cat(name)
+                    return
```
j-gynn: I feel like there's a better way of doing this.

maybe like a for loop that goes over each of the elements, then checks `if event.ui_element in looped.items()`. there's no reason to do the same code four times

scripts/ui/windows/dev_tool_windows/white_patch_tool.py
```diff
@@ -0,0 +1,313 @@
...
+        selection.extend(
+            [f"mid{p}" for p in self.elements["mid_patch_choice"].selected_list]
+        )
+        selection.extend(
+            [f"high{p}" for p in self.elements["high_patch_choice"].selected_list]
+        )
+        selection.extend(
+            [f"mostly{p}" for p in self.elements["mostly_patch_choice"].selected_list]
+        )
+        return selection
```
j-gynn:
```suggestion
        return [f"{amount}{p}" for amount in ["little", "mid", "high", "mostly"] for p in self.elements[f"{amount}_patch_choice"].selected_list]
```

scripts/ui/windows/dev_tool_windows/white_patch_tool.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def set_preview_cat(self, name):
+        self.preview_cat = self.create_cat(name)
+        self.elements["preview_cat"].set_image(
+            pygame.transform.scale(
+                self.preview_cat.sprite, ui_scale_dimensions((100, 100))
+            )
+        )
+
+    def check_current_combos(self, selection):
```
j-gynn: is this not the same code as in the tortie patches? (/genq)
could this live in the base version?

j-gynn (changes requested):

scripts/screens/StartScreen.py
scribblecrumb: Line 244 ;) It's within the if statement that checks for the config

j-gynn (approved):

PR #5755, opened 2026-08-16
[CLEANUP] `Name` class

rustykitty:
## About The Pull Request

Some fairly large changes to the Name class. Functional changes include:
- `give_prefix` and `give_suffix` methods will reroll until a suitable name is found, rather than relying on the calling function to do so. This fixes some bugs (see below).
    - Constructor is changed appropriately to account for this
- Removed the pelt parameter on the constructor, as it was unused (and a bug caused it to crash whenever the parameter was used anyways)
- A new internal class method `_usable_name` was introduced to support the above.

Changes that shouldn't affect behavior:
- names_dict changes:
    - load_localized_names was made into a classmethod, and avoids something that I think was happening before, where each Name instance could've had it's own copy of the names_dict, wasting memory.
    - names_dict was also made to be consistently a class attribute; previously, it was assigned to the instance in load_localized_names(). This change seemed to greatly improve performance during an (albeit unrealistic) benchmark I performed where I generated 1000 names.
- The module-level variable `names` has been removed and the prefix history has been moved to a class attribute.

Bugs fixed (hopefully):
- Double name (prefix == suffix or similar) when randomizing prefix or suffix.
- Naturally-occurring double name in some situations (suspected to be because of outsiders joining the clan calling `give_suffix`)
    - No issue number for now; I remember @velvetfawn was going to make a bug report about this but hasn't yet.

## Proof of Testing

https://github.com/user-attachments/assets/857e4b9e-e19b-40ca-a1f6-0b9b2f4ce54b

Quick script to test `give_suffix` behavior. Let me know if this should be included as a unit test!
image
image

Wrote a quick script to test:
```
import json

PATH = "saves/Quiver/clan_cats.json"

with open(PATH) as fp:
    data = json.load(fp)

from scripts.cat.names import Name

for cat in data:
#    print(cat)
    if not Name._usable_name(cat["name_prefix"], cat["name_suffix"]):
        print(cat["ID"], f"{cat["name_prefix"]}'{cat["name_suffix"]}", "has some issues")
```
image
No issues it seems.


scripts/cat/names.py
```diff
@@ -267,71 +236,76 @@ def give_prefix(self, eyes, colour, biome):
...
+
+            # prevent prefix duplications from happening
+            if self.prefix in self.prefix_history or not self._usable_name(
+                self.prefix, self.suffix
+            ):
+                continue
+            else:
+                self.prefix_history.append(self.prefix)
+                # Set the maximin length to 8 just to be sure
+                if len(Name.prefix_history) > 8:
```
keyraven: Change to self.prefix_history to stay consistent with the rest of the this code-block.

keyraven (changes requested):
Looks good! I requested one (very) small change - then it will good to go!

keyraven (approved):

PR #5798, opened 2026-08-18
[CLEANUP] Event retrieval

scribblecrumb:
## About The Pull Request
Once again cleaning up the TextPoolEvent stuff. This centralizes the event finding in general, hooking all of them up into the same loop of "test events until one works".  Patrols are a little different, just because they get fancy with some stuff and have a few extra constraints to test for, but they're still mostly hooked in.

Also seems that `_get_cats_with_stat` may not have been working? It does work now.

Other notes:
- Implemented a `weight` minimum of 1, cus we were sometimes ending up with negative weights
- Changed our various `load_event` funcs to return a copy of their dict entry instead of the dict entry itself. Turns out that the caches were inadvertently expanding themselves with dupes because of this, which was leading to lag.
- `get_valid_event` always utilizes `weights`, so that means rel events and thoughts will now be chosen based on weights. They weren't doing so before.
- Set `frequency` on `textpoolevent` to default at 4
- Flip flopped the backstory inclusionary/exclusionary weight calc because I realized it was doing the opposite of what I'd meant it to
- Implemented `used_thoughts`, aka a way for us to prevent thoughts from being repeated until all unused thoughts have been attempted!
- `_get_cats_with_stat` did not seem to be working, it does work now
- We were also missing a `_get_cats_with_health`, so I've added that
- TextPoolEvents now share a file load function

Makes it easier to convert other areas of the game to utilize textpoolevent in the future! Apologies if the scope seems a bit all over the place here! Centralizing this functionality inadvertently revealed inconsistencies that were hiding some bugs, fixing those felt necessary.

## Proof of Testing
Tests should be passing!

## Changelog/Credits
- Centralized event retrieval and file load into one func for all TextPoolEvent formats
- `get_valid_event` always utilizes `weights`, so that means rel events and thoughts will now be chosen based on weights. They weren't doing so before.
- Implemented `used_thoughts`, aka a way for us to prevent thoughts from being repeated until all unused thoughts have been attempted!


scripts/events_module/patrol/patrol.py
```diff
@@ -345,121 +344,55 @@ def _filter_patrols(
...
         # GET PATROL
         chosen_patrol: Optional[PatrolEvent] = None
 
         # first we see if we can get a romantic patrol
-        if romantic_patrols and not patrol_override:
-            chosen_patrol = self._get_valid_patrol(
-                romantic_patrols.copy(), chosen_frequency, patrol_override
-            )
+        if romantic_patrols and not self.debug_patrol_id:
+            chosen_patrol = self._get_valid_patrol(romantic_patrols.copy())
```
keyraven, line 352:
```suggestion
            chosen_patrol = self._get_valid_patrol(romantic_patrols.copy())
            if not self._decide_if_romantic(chosen_patrol):
                  chosen_patrol = None
```

scripts/events_module/patrol/patrol.py
```diff
@@ -345,121 +344,55 @@ def _filter_patrols(
...
 
         if chosen_patrol and not self._decide_if_romantic(chosen_patrol):
```
keyraven: See above -- move this block into the previous if-block for readability.

scripts/events_module/patrol/patrol.py
```diff
@@ -345,121 +344,55 @@ def _filter_patrols(
...
-                chosen_patrol = test_patrol
-                self.involved_cats = involved_cats
+            if not chosen_patrol:
+                if not self.used_patrols:
+                    # No patrols found even after resetting used patrols.
+                    # This should only be possible when filtering for romance patrols.
+                    return None
+
+                # if we couldn't find a patrol, then we need to clear the used_patrols and try again
+                self.used_patrols.clear()
```
keyraven: The used_patrols are cleared, but it doesn't look like it tries to find a fitting patrol again (either here for before). Try a call _get_valid_patrol recursively.

scripts/events_module/text_pool_event/event_retrieval.py
```diff
@@ -0,0 +1,150 @@
...
+def get_valid_event(
+    primary_cat: Cat,
+    involved_cats: dict,
+    interactable_cats: list[Cat],
+    possible_events: list[Union[PatrolEvent, TextPoolEvent]],
+    other_clan: Optional[OtherClan] = None,
+    ensured_id: Optional[str] = None,
+    test_general_constraints: bool = True,
+    test_cat_constraints: bool = True,
+    test_frequency: bool = True,
```
j-gynn: these would be implying that by being True, we are testing those things, but the opposite is the case. Maybe `_active`? `frequency_active`, `general_constraints_active`, `cat_constraints_active`?

scripts/events_module/text_pool_event/event_retrieval.py
```diff
@@ -0,0 +1,150 @@
...
+    chosen_frequency = get_frequency() if test_frequency else 4
+
+    chosen_event: Optional[Union[PatrolEvent, TextPoolEvent]] = None
+    temp_involved_cats = {}
+    tested_events = set()
+
+    # retrieve the ensured event from the list
+    ensured_event = None
+    if ensured_id:
+        ensured = [e for e in possible_events if e.event_id == ensured_id]
```
j-gynn:
```suggestion
        ensured = next(e for e in possible_events if e.event_id == ensured_id, None)
```
short-circuits the generator, just works a bit faster

scripts/events_module/thoughts/generate_thoughts.py
```diff
@@ -94,12 +89,41 @@ def _new_thought(
...
-            )
+            involved_cats = {
+                "m_c": main_cat,
+            }
+            if other_cat:
+                involved_cats.update({"r_c": other_cat})
+            if other_clan_id and game.clan.all_other_clans:
+                other_clan = [
+                    c for c in game.clan.all_other_clans if c.group_ID == other_clan_id
+                ][0]
```
j-gynn:
```suggestion
                other_clan = next(
                    c for c in game.clan.all_other_clans if c.group_ID == other_clan_id)
```
No need for a default value here imo, if it fails (then it'll crash), but the problem came from upstream

j-gynn (changes requested):

scripts/events_module/text_pool_event/event_retrieval.py
scribblecrumb: Sorry, maybe I'm misunderstanding, but if they're True then we *are* testing those things.  Or do you mean as in unittests?

j-gynn: Yeah, that's the problem lol. We should consider `test` to mean testing environment as a reserved/unambiguous term, imo. I'm reading it like unittesting

scribblecrumb: Ahhh okay, I see what you mean then.

scripts/events_module/text_pool_event/event_retrieval.py
```diff
@@ -0,0 +1,152 @@
...
+                    None,
+                    {},
+                )  # failed to find anything, so we send back and origin handles it
+            else:
+                used_frequencies.add(chosen_frequency)
+                chosen_frequency = find_new_frequency(used_frequencies)
+                tested_events.clear()
+            continue
+
+        if ensured_event and ensured_event in possible_events:
```
keyraven:
```suggestion
        if ensured_event:
```
You pick ensured_event from possible events earlier in the function - no need to check if it's within the list of possible events.

keyraven (changes requested):

keyraven (approved):

j-gynn (approved):

PR #5803, opened 2026-08-19
[ENHANCEMENT] Adjustments to POI constraints and text tag

### LINKED ISSUE #5171: [ENHANCE] add category to POI tagging
### Description

As discussed on discord. The current state of pulling POI from constraints via tag could pull from all three categories, Gathering, Moonplace, and Territory when it may not be appropriate. Adding the ability to constrain via category will remedy this.

scribblecrumb:
## About The Pull Request
- Changed POI text tag to just `POI` instead of `{POI/constraints}`. This is because the constraints are already specified in the parameters of the event, so there's really no reason to also specify them here.
- Ensured that the POI name is consistent between both initial patrol text and outcome text. I don't think this was happening before, so it would be possible for the initial text to mention one POI name while the outcome mentions a different one.
- Added a func for retrieving a POI name based off multiple constraint params
- Adjusted POI filter func to allow the usage of multiple constraint parameters at once. I don't super *love* the implementation of this because it just wraps the func that I made for retrieving the POI name, meaning we essentially run that func twice for a given POI event.  However, I don't think we can really manage to just run that func once, cus it would require a lot of rather tedious passing of the retrieved poi name back up a chain of func returns, *and* I don't love the thought of it being in `event_filters` while also passing more info back than just a bool, nor do I love the idea of a filter func living outside of `event_filters`. Open to ideas if it does really bother someone.  Personally I think it's not gonna cause performance problems, it's just very slightly less optimal.

## Linked Issues
closes #5171

## Proof of Testing
image
image
image

## Changelog/Credits
- Changed POI text tag to just `POI` instead of `{POI/constraints}`. This is because the constraints are already specified in the parameters of the event, so there's really no reason to also specify them here.
- Ensured that the POI name is consistent between both initial patrol text and outcome text. I don't think this was happening before, so it would be possible for the initial text to mention one POI name while the outcome mentions a different one.
- Added a func for retrieving a POI name based off multiple constraint params
- Adjusted POI filter func to allow the usage of multiple constraint parameters at once.


Ryos00 (approved):

Ryos00:
waiting for Sel to look it over before merging

selkirks (approved):

PR #5816, opened 2026-08-19
[BUGFIX] Affair Breakup Chance Key Fix

Arcticnightsky:
## About The Pull Request
...I genuinely have no idea how this happened, but for some reason affair_breakup_chance was removed from game.config? It was causing a crash where the game couldn't find the key for it. I also fixed the place where it was calling the key, as it said "mates_breakup" instead of just ".breakup", also added a little slight indentation to the affair reactions so that it's consistent to the other breakup reactions and doesn't cause any future issues.

## Linked Issues
Fixes this crash reported on discord: https://discord.com/channels/1003759225522110524/1005586496142704800/1539687049098043492

## Proof of Testing
Screenshot 2026-08-19 152200

It works again, though for this instance the affair remained undiscovered.

### DIFF
--- a/resources/game_config.toml
+++ b/resources/game_config.toml
@@ -278,7 +278,10 @@ dev_tools = false
       romance = -50
       trust = -30
       like = -40
-    
+
+    # the chance for a break up to happen after an affair has been discovered
+    # any number lower than this increases the likelyhood of a break up
+    affair_breakup_chance = 0.8
 
 [new_cat]
   # number of welcoming events to add when a new cat joins
--- a/resources/game_config.toml
+++ b/resources/game_config.toml
@@ -228,7 +228,7 @@ dev_tools = false
 
   [mates.breakup]
     # the chance for a break up to happen after an affair has been discovered
-    # any number lower than this increases the likelyhood of a break up
+    # any number lower than this decreases the likelyhood of a break up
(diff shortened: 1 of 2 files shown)


Chinch-Bug:
note: toml files don't work off pure indentation. placing it under [mates.breakup.reactions.affair_discovery_other_mate_reaction] means it will be considered part of that block regardless of indentation. you would want a standalone variable to be sitting directly under [mates.breakup] itself

Arcticnightsky:
> note: toml files don't work off pure indentation. placing it under [mates.breakup.reactions.affair_discovery_other_mate_reaction] means it will be considered part of that block regardless of indentation. you would want a standalone variable to be sitting directly under [mates.breakup] itself

Oh ok 😅 I was just making sure everything was consistent

resources/game_config.toml
```diff
@@ -227,7 +227,10 @@ dev_tools = false
       trust = 20
 
   [mates.breakup]
-
+    # the chance for a break up to happen after an affair has been discovered
+    # any number lower than this increases the likelyhood of a break up
```
larkgz: I am pretty sure that this is the opposite? i.e. that a lower number would decrease the chance of a break up

larkgz (approved):

PR #5827, opened 2026-08-20
[TASK] Tainted Territory

ssuncreature:
## About The Pull Request

- Adds Tainted Territory as a cruel season card, which makes conditions last longer

- Was in the spreadsheet. I know the feasibility notes said to use the no medicine cat stats, but I thought it might be good to make a multiplier so that herbs still have an effect. I'm happy to adjust modifier if its not extreme enough or too extreme

## Proof of Testing

Screen Shot 2026-08-20 at 13 22 46
Kitty has bee sting
Screen Shot 2026-08-20 at 13 23 12
Their conditions file says duration is 2

Clan has an active and healthy meddie, and correct herbs, so base duration of 1 is being modified correctly.


scribblecrumb (changes requested):
I'd like to see an implementation here that isn't essentially a "card toggle".  We want to, as much as possible, avoid implementation like "if has this card, do this".  For this, a better method would be adding a modifier config in some fashion of `duration = med_duration * config_modifier` so that no if statements are needed

ssuncreature:
Screen Shot 2026-08-21 at 10 12 11
tick fever multiplier correctly added

scribblecrumb:
You've got some test failures to resolve

PR #5847, opened 2026-08-22
[ENHANCE] Improve Save File Error Reporting

### LINKED ISSUE #4894: [ENHANCE] Improve save file error reporting
### Description

There has been feedback that some of the save file error messages are unhelpful:
* Error loading clan.json - Too vague and doesn’t say what actually went wrong
* Not all JSONs are reporting if they are malformed, causing irrelevant error messages to appear

Treefire33:
## About The Pull Request
Changes the way save errors are reported and shown to the player upon opening the game or switching clans. More specifically, for every unique exception that can happen, a unique string is used to inform the player of what specifically went wrong. Each error now also includes more detail as to where the broken file is, which clan the file belongs to, and a traceback of the python exception.

Clarifies what exactly went wrong during the save loading process, which should hopefully reduce the amount of save related bug reports and also help non-technical players understand what exactly they need to fix/change.

## Linked Issues
Fixes: #4894

## Proof of Testing
The errors were tested on a legacy clan.json save (in this case "Brokenclan.json").

Missing clan.json
LegacyClanMissingJson

Malformed clan.json
LegacyClanMalformedJson
Traceback
LegacyClanMalformedJsonTraceback

Missing key in clan.json
LegacyClanMissingKey
Traceback
LegacyClanMissingKeyTraceback

Missing clan_cats.json
LegacyClanMissingClanCatsJson

Malformed clan_cats.json
LegacyClanMalformedClanCatsJson
Traceback
LegacyClanMalformedClanCatsJsonTraceback

Missing cat key in clan_cats.json
LegacyClanMissingCatID

Invalid cat key in clan_cats.json
LegacyClanMissingCatKey

Malformed clan.json (newer clan save)
NewerClanMalformedJson

### DIFF
--- a/scripts/game_structure/load_cat.py
+++ b/scripts/game_structure/load_cat.py
@@ -71,12 +71,16 @@ def json_load():
         with open(clan_cats_json_path, "r", encoding="utf-8") as read_file:
             cat_data = ujson.loads(read_file.read())
     except PermissionError as e:
-        switch_set_value(Switch.error_message, f"Can\t open {clan_cats_json_path}!")
-        switch_set_value(Switch.traceback, e)
+        switch_set_value(
+            Switch.error_message,
+            f"{clan_cats_json_path} can't be accessed. See traceback below of more details.",
+        )
         raise
     except ujson.JSONDecodeError as e:
-        switch_set_value(Switch.error_message, f"{clan_cats_json_path} is malformed!")
-        switch_set_value(Switch.traceback, e)
+        switch_set_value(
+            Switch.error_message,
+            f"{clan_cats_json_path} is malformed. See traceback below of more details.",
+        )
         raise
 
     old_tortie_patches = convert["old_tortie_patches"]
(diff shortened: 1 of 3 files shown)


scripts/game_structure/load_cat.py
```diff
@@ -140,11 +149,14 @@ def json_load():
     inheritance_db.load_inheritances(Cat, get_faded_ids)
 
 
-def csv_load(all_cats):
+def csv_load(_all_cats):
```
keyraven: Why does this change the name of the function parameter?

keyraven (changes requested):

scripts/game_structure/load_cat.py
Treefire33: It usually silences any warnings or errors around an unused parameter in the function. I suppose I didn't want to change the function's signature so I added the underscore instead. If it's wanted I could remove the parameter or revert the name change.

scribblecrumb: If it's an unused parameter (as seems to be implied?) then I'd prefer it just be removed

PR #5849, opened 2026-08-22
[REWORK] Ceremonies

### LINKED ISSUE #2427: [CODE] Refactor ceremonies
**Description**
Moon ceremony events are currently handled separately from ShortEvents.  They've grown to a point of being complex enough that I believe they would benefit from a more structured format, rather than the tag spaghetti they currently rely on.  I think they would mesh well with the new ShortEvent format and could likely *be* handled as ShortEvents.  They would probably need to be a new type of ShortEvent (so adding a "ceremony" type alongside the current "death", "misc", "new_cat", and "injury").  ShortEvent format documentation is found in our wiki docs.

Deputy and Leader ceremonies are also currently handled separate from all the other ceremonies.  I would prefer if they were all handled together (or at least, more clearly intertwined) as the current iteration is rather confusing to navigate.  Note: Leader StarClan ceremonies are entirely separate and shouldn't be included here, I'm specifically talking about the Leader moon ceremonies.

### LINKED ISSUE #3237: [CODE] Transfer Ceremonies to ShortEvents
**Description**
At the moment, normal ceremonies (all ceremonies other than leader ceremonies) utilize their own format.  This format is older and a bit limiting/confusing.  Ideally, ceremonies should be transferred into ShortEvents, which is a format that should still cover all it's required functionality.  `ceremony` should become a new event type and be contained within it's own events folder.

scribblecrumb:
## About The Pull Request
Apologies for how massive this looks, a large chunk of the lines are doc and json changes.

- Ceremonies now utilize `TextPoolEvent`! Deputy events in particular are no longer hard coded.
- I've moved all the relevant ceremony code into `event_module/ceremony` to help reduce the bulk of `events.py`
- New structure is `check_for_ceremony()` for the general "see if a cat needs a ceremony" check. Deputies get a specific `check_and_promote_deputy()` just for them.
- Both of those funcs hook into `trigger_ceremony()`, which can also be called directly when we don't need to "check" for a ceremony and instead already know what ceremony is needed for a cat.  This is the function that actually applies rank changes and creates the ceremony for display.
- Among other functions that were moved into `perform_ceremony.py` are:
  - `_adult_become_mediator()` which used to be known as `mediator_events()`. Code largely unchanged here beyond changing function calls.
  - `_is_suitable_mediator_app()` code that has been moved to it's own function, is entirely unchanged from it's original form beyond necessary changes to make it work as a function.
  - `_is_suitable_medcat_app()` no changes to this code, it has just been moved
  - `_handle_leader_ceremony()` old code moved into it's own function
- Now to actually generate the ceremony event we use `create_ceremony()`. This function handles getting a valid event, applying it's consequences, processing the text, and adding it to the current events list.
- Some new event parameters added to facilitate ceremony functionality.
  - `current_exp` used to constrain per exp level
  - `has_mentor` expanded to allow specification of current vs. former
  - `has_apprentice` mirrors `has_mentor`
  - `name` used currently to specify if a cat has a suffix or not
- Existing parameters received additions to functionality
  - `past_mentor/app` and `past_app/mentor` added as relationship constraints for marking cats as former mentors/apps
  - `-clan:rank` exclusionary tags added to mark clans as specifically *not* having a rank present  
- Schemas and documentation appropriately updated
  - For docs I also created a centralized `involved_cats` documentation for all TextPoolEvents to share 
- jsons were automatically reformatted via a script and then condensed by-hand where possible. Their IDs were also made more descriptive. I ended up fixing some problems with them, such as kitten traits being used for constraints or old, now unused traits (like sneaky) still being listed. The one master file was also split up by rank into multiple jsons to better facilitate reading and editing.

Makes ceremonies easier to expand for writers and easier to handle for coders!

## Linked Issues
closes #2427
closes #3237

## Proof of Testing
image

Bulk test is also passing!

## Changelog/Credits
- Transfers ceremonies into the new `TextPoolFormat`
- Un-hardcodes the deputy ceremonies
- Adds documentation regarding how to write ceremonies


scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -0,0 +1,427 @@
...
+        if get_clan_setting("12_moon_graduation"):
+            _ready = main_cat.moons >= 12
+        else:
+            _ready = (
+                main_cat.experience_level not in ["untrained", "learning"]
+                and main_cat.moons
+                >= constants.CONFIG["graduation"]["min_graduating_age"]
+            ) or main_cat.moons >= constants.CONFIG["graduation"]["max_apprentice_age"][
+                main_cat.status.rank
+            ]
```
j-gynn: should use get_config

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def _adult_becomes_mediator(cat) -> bool:
+    """
+    Check if a cat wants to switch from their current role into the role of a mediator
+    """
+    if get_clan_setting("become_mediator"):
+        # Note: These chances are large since it triggers every moon.
+        # Checking every moon has the effect giving older cats more chances to become a mediator
+        change_chance_per_role = constants.CONFIG["roles"]["become_mediator_chances"]
```
j-gynn: get_config

j-gynn (commented):
Code side looks good. Two comments about getconfig but we talked about that already. Only needs schema and writing reviewing.

.vscode/settings.json
```diff
@@ -37,6 +37,13 @@
                 "events/relationship_events/joining_interactions/**/*.json"
             ]
         },
+        {
+            "url": "./schemas/ceremony_event.schema.json",
```
larkgz:
```suggestion
            "url": "./schemas/ceremony.schema.json",
```

scripts/models/text_pool_event/name_check_dict.py
```diff
@@ -0,0 +1,12 @@
+from typing import Union
+
+from pydantic import BaseModel, ConfigDict, Field
+from pydantic_core import MISSING
+
+
+class MentorApprenticeDict(BaseModel):
```
larkgz:
```suggestion
class NameCheckDict(BaseModel):
```

scripts/models/text_pool_event/cat_dict.py
```diff
@@ -3,10 +3,13 @@
 from pydantic import BaseModel, ConfigDict, Field
 from pydantic_core import MISSING
 
+from scripts.events_module.parameter_dicts import NameCheckDict
```
larkgz: Seem to be importing wrong `NameCheckDict`

larkgz (commented):
* I think the docs for `current_exp` are missing
* There are some unused Union imports you can remove

larkgz (approved):

PR #5852, opened 2026-08-23
[ENHANCEMENT] Debug Menu/Console Improvements and New Commands

Treefire33:
## About The Pull Request
This PR adds two commands, improves the usage documentation for existing commands, and all-round attempts to improve the experience with the debug console.

### Commands:
* `cruel`:
  * Commands for managing the cruel season specific settings of a clan.
  * Currently only allows for listing of all (or a specific category of) cruel cards or the managing of a clan's cruel cards.
* `devtools`:
  * Command for accessing the various developer tools.
  * Allows the user to either enable for the rest of the session or open the tools independently.
* `clan skip`:
  * Sub-command for `clan` which allows the user to bulk skip some amount of moons.
* `settings`:
  * A condensed version of the original `get`, `set`, and `toggle` commands. They are functionally the same. 

### Console Changes:
* The debug console now keeps track of *all* previously entered commands rather than the most recent.
  * The command list is navigated using the up and down arrow keys.
* If the command line is focused, key down events on other screens won't register.
* You can run multiple commands in series by putting an ampersand between them.
  * For example, `clear & help` clears the console then runs the `help` command.
  * You can do this multiple times: `clear & help & help eval` clears the console, runs the `help` command, then gets the help text for the `eval` command 

### Other
* Subcommands are now parsed recursively, meaning a subcommand can have a subcommand and so on.
  * Currently the only command this is used for is `cruel card` with the `list` subcommand. 

I think the debug menu is a nice tool for doing simple and quick testing. It may not be as comprehensive or effective as an actual test, but it does excel at allowing somebody to immediately test or see their changes in game. With the proper features, the debug menu could prove to be a powerful tool for writers and artists.

## Proof of Testing
https://github.com/user-attachments/assets/38e7b949-a30e-458e-8bbc-2f24d13da2b7


larkgz:
> This PR is currently a draft because I'm still trying to figure out what exactly people would use the debug console for. Most of these changes were me assuming what commands could be helpful rather than what is helpful.

If that's the only thing you are worried about, I think that you can just mark it as ready for review! People will add additional commands as they think of them / they become necessary.

scripts/debug_commands/cruel.py
```diff
@@ -0,0 +1,129 @@
...
+        elif args[0] == "behavior":
+            card_scope = constants.CRUEL_CARDS_BEHAVIOR
+        elif args[0] == "danger":
+            card_scope = constants.CRUEL_CARDS_DANGER
+        elif args[0] == "environment":
+            card_scope = constants.CRUEL_CARDS_ENVIRONMENT
+        elif args[0] == "origin":
+            card_scope = constants.CRUEL_CARDS_ORIGIN
+        limit_possible = len(args) >= 2
+        get_cards_list(card_scope, limit_possible)
```
larkgz, line 66: Causes a crash when no arguments are specified

scripts/debug_console.py
```diff
@@ -67,66 +66,78 @@ def __init__(self, rect, manager):
...
+            if not command in cmd.valid_names:
+                continue
+
+            if cmd.bypass_conjoined_strings:
+                args = raw_command.strip().split(" ")[1:]
+
+            while len(args) > 0 and len(cmd.sub_commands) > 0:
+                command_found = False
+                for sub_command in cmd.sub_commands:
+                    if args[0] not in sub_command.valid_names:
```
larkgz: Seems to be an issue where entering a subcommand without arguments can cause the game to crash here (e.g.  `spawn pelts` or `pregnancies add`)

scripts/debug_console.py
```diff
@@ -67,66 +66,78 @@ def __init__(self, rect, manager):
...
+                current_arg = ""
+                continue
+
+            current_arg += char
+
+        # flush arg buffer in case there are no arguments
+        args.append(current_arg)
+
+        # command name, arguments
+        return args[0], args[1:]
```
larkgz: I think it would be better to split this using `shlex`.

larkgz (changes requested):
I think it might be good to migrate to parsing using `argparse` or something similar in the future, though I think that is beyond the scope of this PR.

larkgz (approved):

PR #5865, opened 2026-08-24
[BUGFIX] new cat retrieval

scribblecrumb:
## About The Pull Request
- Adds a `PatrolOutcome` enum to clean up the code a tad
- Modified our event retrieval system to allow new cat creation to be disabled when necessary and to let us disable repeat constraint assessment on cats who have already been assigned
- Once a patrol outcome is chosen, we now check to see if any new cats still need to be assigned, and handle as needed.

I realized the way we were handling outcomes was a little flawed. We would find a valid outcome for both success and failure. This includes gathering all the involved cats for those outcomes. In doing so, we would also create any new cats that were necessary for those outcomes.  However, this becomes an issue when we create cats for an outcome (such as the chosen fail outcome), but then use a different outcome (in this case, the success).  Now the cats created for our potential "failure" are just.. there, without any sort of explanation to the player.

I've adjusted it so that we "skip" finding any involved cats who can be newly generated during the initial outcome filtering process.  Once we've decided on a specific outcome, we go back and fill those slots with the necessary cats.  This means that we no longer create a bunch of cats for no reason.

## Proof of Testing
image

current outsiders for this clan

image

image

chosen failure outcome

image

chosen success outcome

both of these can generate new cats

image

our outcome cats for both outcome choices show that none of those new cats have been generated and assigned yet

image

After an outcome was chosen (in this case, success was chosen) you can see that we have now filled those new cat slots

image

new cats created! (worm dying message is an unrelated bug)

image

and our ending outsider page looks exactly as it did before, no extraneous cats have been created

## Changelog/Credits
- Fixes patrols sometimes creating new cats for outcomes that the patrol did not receive


keyraven:
I like the new PatrolOutcome enum, and this does fix the patrol bug. However, I have concerns the fix, and `find_and_create_cats` in general. First, this function is in charge of creating cats for the event, which is the only write-operation not contained in `execute_outcome`. I don't think there is a strong reason to separate new cat creation from other write operations in this way.  Moving new cat creation to `execute_outcome` does not break any filtering nor text replacement. Secondly, this function is called for many events until the correct one is found.  The only reason that more unnecessary cats aren't being created is because the new_cat check is the last one. It doesn't feel like best practices to call a possible write-function for events that may not be fully executed.  And, if for whatever reason, we want to filter, but not necessarily execute, an event, we couldn't do that. 

I have drafted a fix that moves the creation of new cats to execute_outcome. [You can find it here. ](https://github.com/keyraven/clangen/tree/update_find_cats)  Could you take a look at it?

keyraven:
Closing - clangen#5938 fixes this problem.

PR #5868, opened 2026-08-24
[ENHANCEMENT] "Assign Random Mentors" Toggle and Event Status Filtering Features

keyraven:
## About The Pull Request

This adds a clan-setting role toggle to disable random assignment of mentors to apprentices. This prevents any apprentices from being given a random mentor at their ceremony or when their current mentor becomes ineligible (dies, rank change, leaves clan). An ineligible mentor will still be removed, but no new mentor will be assigned in their place.  Therefore, when toggled, all mentors, and mentor changes, will need to assigned manually using the change mentor screen. 

If there are apprentices without mentors, a warning event will pop on in the event list. This will tell the player how many apprentices don't have mentors, and link them. 

In order to add the proper constraints to ceremonies, I added some additional functionality to the status-based event tags. First, there is now a "status:warrior-like" option which will check for the total number of warriors, plus leader and deputy.  Also, I added an optional sub-tag for each status which can set the minimum number required. Before, the filtering always checked for 2 or more of any non-leader non-deputy status, and 1 or more of "apps", leader and deputy. This default behavior is preserved, but now you can add `(min:#)` after the status to override the minimum value. 

There was also a value hanging around in the "debug_ensure_ceremony_id", which I removed. 

Some players prefer to set their apprentice's mentors manually. Past mentors are tracked, which means that, previously, those players might of had unwanted mentors in the "past mentors" list of their cats, or in the "past apprentices" list of their cats.  Previously, this was (somewhat) alleviated by hard-coding it so apprentices whose mentors changes at 6 moons old did not have their previous mentor recorded. This was so that players who desired could change the mentor of new apprentices without having the previously-assigned random mentor recorded.  However, this is strange unintuitive (and non-documented) edge-case behavior that I would like to remove. 

The additions to status-based event filtering will also be helpful for more than just these ceremonies. The "warrior-like" category will be good for certain events, the option "min" option allows writers more flexibility when writing events that require a certain number of a certain status.  

## Proof of Testing

https://github.com/user-attachments/assets/e7a44f93-fce8-48e9-bfce-5f27f509fd93


resources/lang/en/events/ceremonies/apprentice.json
```diff
@@ -34,7 +34,8 @@
         ]
     },
     {
-        "event_id": "app_no_mentor0",
+        "event_id": "warr_app_no_mentors_available0",
+        "tags": ["-clan:warriors(min:1)"],
```
scribblecrumb:
```suggestion
        "tags": ["-clan:warrior(min:1)"],
```

resources/lang/en/settings.en.json
```diff
@@ -44,6 +44,8 @@
         "deputy_tooltip": "The Warrior Code rules will be taken into account when choosing a deputy.",
         "12_moon_graduation": "Disable experience-based apprentice graduation",
         "12_moon_graduation_tooltip": "All apprentices will graduate at 12 moons.",
+        "assign_mentors": "Assign random mentors to apprentices.",
```
scribblecrumb:
```suggestion
        "assign_mentors": "Assign mentors to apprentices automatically.",
```
Better mirrors the language used for the deputy setting

scribblecrumb (changes requested):

resources/lang/en/settings.en.json
keyraven: Applied!

resources/lang/en/events/ceremonies/apprentice.json
keyraven: Applied!

scribblecrumb (approved):

PR #5897, opened 2026-08-27
[CLEANUP] remove unused condition loads

### LINKED ISSUE #5895: [ENHANCE] remove leftover unused definitions from cats.py
### Description

In cats.py, lines 2576-2584 define resource directories for injuries and permanent conditions, both of which are handled elsewhere now. 'ILLNESSES' is no longer present, so it seems likely the other 2 and file path were just forgotten about and therefore not removed.

rustykitty:
## About The Pull Request
Remove some unused JSON loads for conditions from `cats.py`.

## Linked Issues
closes #5895

## Proof of Testing
Game runs fine. If there was an issue it probably would've happened by now
image

### DIFF
--- a/scripts/cat/cats.py
+++ b/scripts/cat/cats.py
@@ -2569,20 +2569,6 @@ class Cat:
 cat_class = Cat
 game.cat_class = Cat
 
-# ---------------------------------------------------------------------------- #
-#                                load json files                               #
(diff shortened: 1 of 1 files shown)


archanyhm (approved):
Looks good to me. We'll see if something breaks, but this change is easily revertable.

Issue #5363, opened 2026-06-27
[PROJECT] Polish Localisation


afellowcorn:
**Description**
Issue for tracking Polish localisation efforts.
Note that events and patrols require a minimum for balancing purposes, so the English text acts as filler until there's enough Polish events/patrols to remove it.

**Scope of Project**
- Large Long-term Project
- Requires basic understanding of json formats

**What has already been completed?**
- Create lang folder #5888 
- Pronouns and config #5903 
- general.pl.json #5903

**Relevant Links**

PR #5903, opened 2026-08-28
[LOC] pl config, pronouns, general

afellowcorn:
## About The Pull Request
The pronouns, config and general files.
`general.pl.json` is the first file with proper translation content, so feel free to ask me to clarify certain parts as needed. Screenshots include most notable uses of that file's strings (it's a pretty all over the place file).

Note that because other strings are still unmodified all their pronoun tags will now show as error.

image

- Polish words for sibling's children differ based on which side of the family they're from, so I use the general descriptor since ClanGen doesn't differentiate that.
image

- Polish does not have the word "condition" to generally refer to a medical condition, so it instead says "has a disability or chronic illness".
image

- Normally these are preceded by "pora" (season/time of year), but that makes these too long to fit on the events screen. With the "current season:" bit his will result in the display reading something like "time of year: of bare leaves".

Localisation

## Linked Issues
#5363 

## Proof of Testing

Zrzut ekranu 2026-08-28 104247
Zrzut ekranu 2026-08-28 102948
Zrzut ekranu 2026-08-28 103823
Zrzut ekranu 2026-08-28 103912
Zrzut ekranu 2026-08-28 103932

## Changelog/Credits
Polish localisation: pronouns and general

### DIFF
--- a/resources/lang/pl/config.json
+++ b/resources/lang/pl/config.json
@@ -1,64 +1,53 @@
 {
-    "lang": "en",
+    "lang": "pl",
     "pronouns": {
         "sets": {
             "-&-sets comment-&-": "These are used to determine what pronouns a cat is assigned when they are generated (or this language is swapped to for the first time). This allows languages with pronoun sets that don't match English 1:1 to display correctly.",
-            "nonbinary": "0",
-            "male": "1",
-            "trans male": "1",
-            "female": "2",
-            "trans female": "2",
-            "default": "0",
-            "plural default": "0",
-            "plural male": "0",
-            "plural female": "0",
-            "plural nonbinary": "0",
+            "nonbinary": "2",
+            "male": "0",
+            "trans male": "0",
+            "female": "1",
+            "trans female": "1",
+            "default": "2",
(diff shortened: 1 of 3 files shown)


resources/lang/pl/config.json
```diff
@@ -1,64 +1,53 @@
...
-                    3
-                ],
+                "values": [2, 3],
                 "format": " "
             },
             {
-                "values": [
-                    4,
-                    5
-                ],
```
archanyhm, line 61: Did you use an autoformatter for this? It's largely unrelated to your changes, I'm just wondering; I don't think we have a canonical one set up.

archanyhm:
Test failures seem unrelated, so I'd be good to merge. Just curious about the JSON formatting changes in the comment above. Not a change request, though.

resources/lang/pl/config.json
afellowcorn: I did not, but the lang files are created using a script, which uses indent=4 ujson write.

archanyhm (approved):

PR #5914, opened 2026-08-29
[CI] Split bulk timeskip into its own job

larkgz:
## About The Pull Request

* Splits bulk timeskip simulation into its own job so it's listed separately
* Titles it so that people are aware that it's ok if it fails 

* Easier to tell if a unit test failed or if it was just the bulk timeskip

## Proof of Testing

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
           architecture: 'x64'
       - name: install dependencies
         run: uv sync
-      - name: Run bulk timeskip simulation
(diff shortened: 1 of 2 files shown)


scribblecrumb:
Dunno that I'm okay with marking it as "it's okay if this fails", cus there's plenty of times where it fails for good reason.

larkgz:
Changed the label to `flaky; may fail for unrelated reasons)`, if that's better?

scribblecrumb (approved):

PR #5920, opened 2026-08-29
[TASK] Pride and Joy cruel card 

Inkpawwc:
## About The Pull Request
adds the pride and joy cruel card, which only allows for the kits of leaders to be selected as deputy. Also changed the art of the mouths to feed and kit season card and made a minor edit to the mouths to feed card description.

project task

## Proof of Testing
image
image
image
image

## Changelog/Credits
Pride and Joy cruel card added - inkpawwc


resources/game_config.toml
```diff
@@ -733,7 +726,11 @@ dev_tools = false
 
 [ranks]
   # permit player to change cat ranks
-  allow_manual = true
+  allow_manual_all = true
+  allow_manual_deputy = true
+
+  # force only the leader's kits to be selected as deputy
+  only_leader_kits_deputy = false
```
j-gynn: Could we nest this under something like `deputy_eligibility`? We have a few other cards planned where deputy eligibility is affected - it'd also mean we could put the "needs to have mentored at least one cat" part in config too (not in this PR, just thinking aloud)

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -151,36 +151,73 @@ def check_and_promote_deputy():
...
-        filter(
-            lambda x: x.status.alive_in_player_clan
-            and x.status.rank == CatRank.WARRIOR
-            and (x.apprentice or x.former_apprentices),
-            Cat.all_cats_list,
+    if get_config("ranks.only_leader_kits_deputy") and game.clan.leader is not None:
+        leaders_kits = []
+        for cat in Cat.all_cats_list:
+            if game.clan.leader.is_parent(cat):
+                leaders_kits.append(cat)
```
j-gynn: I think we could do something with the inheritance module to avoid iterating over every single cat twice.

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -151,36 +151,73 @@ def check_and_promote_deputy():
...
+            )
+        )
+    else:
+        possible_deputies = list(
+            filter(
+                lambda x: x.status.alive_in_player_clan
+                and x.status.rank == CatRank.WARRIOR
+                and (x.apprentice or x.former_apprentices),
+                Cat.all_cats_list,
+            )
```
j-gynn: This whole section should be restructured.

I would suggest starting with a list `eligible_cats` and then, if `get_config` etc. etc., we constrain it further.

Basically (pseudocode, don't blindly copy this haha):

```
    possible_deputies = list(
        filter(
            lambda x: x.status.alive_in_player_clan
            and x.status.rank == CatRank.WARRIOR
            and (x.apprentice or x.former_apprentices),
            Cat.all_cats_list,
        )

    if get_config("ranks.only_leader_kits_deputy") and game.clan.leader is not None:
        leader_kits = InheritanceDB.get_kits(game.clan.leader)  # This isn't necessarily the real function I'm doing it from memory
        possible_deputies = [c for c in possible_deputies if c.ID in leader_kits]  # assuming that InheritanceDB returns cat IDs

```

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -151,36 +151,73 @@ def check_and_promote_deputy():
...
-                and x.status.rank == CatRank.WARRIOR,
-                Cat.all_cats_list,
+        if get_config("ranks.only_leader_kits_deputy") and game.clan.leader is not None:
+            # If none of the leader's kits meet all the requirements for deputy, choose one randomly, with special text.
+            all_warriors = list(
+                filter(
+                    lambda x: x.status.alive_in_player_clan
+                    and x.status.rank == CatRank.WARRIOR,
+                    leaders_kits,
+                )
```
j-gynn: We do tend to use list comprehension for this instead of `list(filter()`, was this an optimisation bottleneck?

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -151,36 +151,73 @@ def check_and_promote_deputy():
...
         # from here we must have appropriate deputy choices
         main_cat = random.choice(possible_deputies)
     else:
         # If there are no possible deputies, choose someone else, with special text.
-        all_warriors = list(
-            filter(
-                lambda x: x.status.alive_in_player_clan
-                and x.status.rank == CatRank.WARRIOR,
-                Cat.all_cats_list,
+        if get_config("ranks.only_leader_kits_deputy") and game.clan.leader is not None:
```
j-gynn: Same restructuring idea here. Create the pool based on the constraints, then use the same block for the outcome. Makes it easier to make changes in future if we need to.

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -151,36 +151,73 @@ def check_and_promote_deputy():
...
         else:
-            # If there are no warriors at all, no one is named deputy.
-            game.cur_events_list.append(
-                EventInformation(i18n.t("hardcoded.ceremony_deputy_none"), "ceremony")
+            all_warriors = list(
+                filter(
+                    lambda x: x.status.alive_in_player_clan
+                    and x.status.rank == CatRank.WARRIOR,
+                    Cat.all_cats_list,
+                )
```
j-gynn: Optimisation?

scripts/screens/RoleScreen.py
```diff
@@ -63,17 +66,26 @@ def handle_event(self, event):
             #                               Ye have been warned
             #
             #
+
```
j-gynn: I think you might've disabled the rank switching mechanic that's used in another Cruel Card with these changes... please confirm? If not, you'll need to move this warning so that it still makes sense.

j-gynn (changes requested):
A few structural changes needed please!

scripts/events_module/ceremony/perform_ceremony.py
Inkpawwc: the original code was already written like this ^^; I just changed the list it's filtering

scripts/events_module/ceremony/perform_ceremony.py
Inkpawwc: The original code was written like this, this was unchanged

j-gynn (approved):

j-gynn:
Is the unit test failure related to this branch's changes?

Inkpawwc:
> Is the unit test failure related to this branch's changes?

image
I don't believe so. I didn't touch anything with patrols for this card

resources/lang/en/hardcoded.en.json
```diff
@@ -19,6 +19,7 @@
         "condition_retire_normal": "Seeing m_c struggling the last few moons, lead_name approaches {PRONOUN\/m_c\/object} and promises that no one would think less of {PRONOUN\/m_c\/object} for retiring early, and that {PRONOUN\/m_c\/subject} would still be a valuable member of the Clan as an elder. m_c agrees; later that day, {PRONOUN\/m_c\/poss} elder ceremony is held.",
         "ceremony_closer": "Visit lead_name's profile to see {PRONOUN\/lead_name\/poss} full leader ceremony.",
         "ceremony_deputy_none": "There are no cats fit to become deputy.",
+        "ceremony_deputy_none_kin": "lead_name refuses to appoint a new deputy. Only {PRONOUN/lead_name/poss} lineage is fit to lead.",
```
scribblecrumb, line 22: While I understand the love for flavor, I do hate adding further "hardcoded" string options for this. The default message is still appropriate for the card, so I don't feel it's necessary to add special text.

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -160,6 +161,13 @@ def check_and_promote_deputy():
         )
     )
 
+    if (
+        get_config("ranks.deputy_eligibility.only_leader_kits_deputy")
+        and game.clan.leader is not None
+    ):
+        leaders_kits = [c for c in inheritance_db.get_children(game.clan.leader.ID)]
```
scribblecrumb, line 168: why do we do it like this instead of `game.clan.leader.get_children()`?

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -172,13 +180,28 @@ def check_and_promote_deputy():
                 Cat.all_cats_list,
             )
         )
+
+        if (
+            get_config("ranks.deputy_eligibility.only_leader_kits_deputy")
+            and game.clan.leader is not None
+        ):
+            # If none of the leader's kits meet all the requirements for deputy, choose one randomly, with special text.
+            leaders_kits = [c for c in inheritance_db.get_children(game.clan.leader.ID)]
```
scribblecrumb, line 189: same q here

scripts/events_module/ceremony/perform_ceremony.py
```diff
@@ -172,13 +180,28 @@ def check_and_promote_deputy():
...
+            all_warriors = [c for c in all_warriors if c.ID in leaders_kits]
```
scribblecrumb, line 190: Could also likely be made into a function that's called both here and up above. That way we aren't repeating this code

resources/game_config.toml
```diff
@@ -733,7 +726,12 @@ dev_tools = false
 
 [ranks]
   # permit player to change cat ranks
-  allow_manual = true
+  allow_manual_all = true
+  allow_manual_deputy = true
```
scribblecrumb, line 730: Instead of this, could instead do something like `can_manually_change` which is a dict of ranks with bool values. Then check if the cat's rank bool is True in rolescreen. would be cleaner and more dynamic for any future cards.

scribblecrumb (changes requested):

scribblecrumb (approved):
Needs conflicts resolved

Inkpawwc:
conflicts have been resolved ^^

scribblecrumb (approved):

PR #5938, opened 2026-08-31
[BUGFIX] Move New Cat Creation to execute_outcome

### LINKED ISSUE #5845: [BUG] Elder joined my clan with no join date
### Pre-requisites

### Description

sorry if this one isn't a bug, but i noticed this cat (who i presume joined on moonskip on a moon i skimmed), and couldn't find a single clue as to how long she'd actually been in my clan. other outsiders who joined via moonskip do have a "they joined the clan on moon X" in their history, and icethrift does not.

### Grade

### Gamemode

Cruel Season

### Challenge cards

Fated Paths, Resolute, Enmity, Kit Season, Underpopulated, Forsaken, Anything For The Throne, Neverending Pollen

1. Have a cat join on moonskip (presumably)
2. See they have no join date in their history

_No response_

1b117f8

### Error log

_No response_

### Additional context

i also have no clue what moonskip event she joined from, if at all?

### Screenshots

Image

Image

Image

other outsiders who joined my clan:

Image

Image

Image

### LINKED ISSUE #5889: [BUG] Border patrol failure generating + adding kits to clan
### Pre-requisites

### Description

When failing to recruit the outsider, sometimes the game will generate the kits that outsider would have and still put those kits in the clan. These kits aren't mentioned in the patrol outcome text (just as the outsider themself is generated "invisibly").

### Grade

### Gamemode

Cruel Season

### Challenge cards

Kit Season, Mouths to Feed

1. Send a patrol
2. Roll gen_bord_healer2
3. Fail a "proceed"
4. Check if any new kits exist with their parent being outside

PATROL START ---------------------------------------------------
Patrol Leader: Alderswoop
Total Number of Possible Patrols | normal: 46, romantic: 0
starting chance: 40 | EX_updated chance: 90
PATROL ID: gen_bord_healer2 | SUCCESS: False
Patrol Frequency: 2 | Patrol Weight: 1
Outcome Frequency: 4 | Outcome Weight: 43

cb31cfe7ce0df57845c180b1219130272d8e8c9d

### Error log

_No response_

### Additional context

I believe I've had this happen on other patrols as well, but I didn't record it (sorry)

### Screenshots

_No response_

### LINKED ISSUE #5946: [BUG] Double outsider mates & the outsider mates joining when their clan-cat mate leaves
### Pre-requisites

### Description

I have no clue what even could've caused this really but what happened was one of my cats had the patrol where they go to meet up with their outsider mate and then they left and their mate randomly appeared in clan instead of outside the clan and they had a second mate not mentioned in the patrol also in the clan

### Grade

### Gamemode

Cruel Season

### Challenge cards

_No response_

Send cats out on solos until they have the patrol?

gen_bord_welcomingbinglegift1
gen_bord_binglegift1
gen_bord_hostilebinglegift1

I'm honestly not sure which one it was, the links provided for "what's a patrol ID and do I need it?" and "how do I find the patrol ID" are broken so I have no clue how to check it and when I asked in the discord these were the ones brought up when I described the patrol

I've included a screenshot of the patrol just in case, alongside the only screenshot I had of the result as I wasn't exactly expecting the bug, and the two mates

b11a89ec

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image

Image

Image

Image

keyraven:
## About The Pull Request

Moves the creation of new cats, for events and patrols, out of `find_cats` and into `execute_outcome`. This moved all write-operations for events and patrols into `execute_outcome`.  This fixes the same bug that #5865 fixes, but in a different way. 

This also adds an additional patrol debug print, displaying the name of the chosen success and failure outcome.

Fixes the bug of unnecessary cats being created on patrol, and moved all write-operations out of the event-filtering stage and into execute_outcome`.  

## Proof of Testing

https://github.com/user-attachments/assets/84c878cc-bb58-47fc-bab3-41fa06989c92


scripts/events_module/patrol/patrol.py
```diff
@@ -426,7 +428,7 @@ def _get_valid_patrol(
             not in Patrol.used_patrols["romance" if find_romance else "normal"]
         ]
         while not chosen_patrol:
-            chosen_patrol, involved_cats = get_valid_event(
+            chosen_patrol, involved_cats, cats_to_create = get_valid_event(
```
scribblecrumb: I don't feel we need to be preserving `cats_to_create`.

Within the `handle_consequences` we can just compare `event.involved_cats` and the `involved_cats` dict that has already been gathered to see what n_c cats are missing and then create them.

scripts/events_module/text_pool_event/find_involved_cats.py
```diff
@@ -70,17 +73,19 @@ def find_cats(
...
             ]
 
+            # CATS THAT CAN BE MADE
+            if "can_create_new_cat" in constraints:
+                # It's OK if we can't find a cat - we can create it later.
+                can_create_new_cat = True
+
+                # Chance to *not* look for
+                if random.random() * 100 > get_config("new_cat.search_for_existing"):
+                    possible_cats = []
```
scribblecrumb: We shouldn't do this. 

The way it's meant to work is that an `n_c` can be added and can either contain the `can_create_new_cat` parameter or not. If it doesn't, then it *must* use an existing outsider/other_clan cat. If `can_create_new_cat` is present, then it will *first* attempt to find an existing cat to use, *then* will create a new cat.  I don't think we should muddle this behavior with randomization.

scribblecrumb (changes requested):

scripts/events_module/text_pool_event/find_involved_cats.py
keyraven: I will remove this. I was playing around with the config as a possible solution to the issue of outsiders joining tending to be old elders, since outsiders tend to grow old.  But I'm not a fan of this solution either, and it shouldn't be part of this PR anyhow.

scripts/events_module/patrol/patrol.py
keyraven: Makes sense! Will do!

scripts/events_module/text_pool_event/find_involved_cats.py
scribblecrumb: Yeah, I'd prefer to see us actually modify some event parameters to require outsiders to be more specific ages. We have a lot more flexibility over that now than we did before.

scripts/events_module/patrol/patrol.py
keyraven: Made the change!

scripts/events_module/text_pool_event/find_involved_cats.py
keyraven: I agree. Change has been made!

scripts/events_module/text_pool_event/find_involved_cats.py
```diff
@@ -30,10 +29,13 @@ def find_cats(
...
     :return: Updated involved_cats dict with valid cats. If dict is empty, then valid cats were not found.
     """
+
+    empty = ({}, {})
+
     temp_involved_cats = involved_cats.copy()
     interactable_cats = interactable_cats.copy()
 
-    cats_to_create = []
+    cats_to_create = {}
```
scribblecrumb: Why are we still creating and returning this variable?

scripts/events_module/text_pool_event/handle_consequences.py
```diff
@@ -111,6 +116,32 @@ def execute_outcome(
...
+    """
+    Creates needed cats for the event, who are not already in event_involved_cats.
+    Modifies event_involved_cats with the new cats.
+    """
+    for abbr, constraints in event.involved_cats.items():
+        # Only create cats who haven't already been assigned in event_involved_cats.
+        if abbr in event_involved_cats:
+            continue
+
+        print(f"Creating new cat in outcome! {abbr}")
```
scribblecrumb: Misleading, since this func is also called for the top level patrol, not just outcomes.  This is also gonna be called for events that aren't patrol outcomes.

scribblecrumb (changes requested):

scripts/events_module/text_pool_event/find_involved_cats.py
keyraven: find_cats() creates and returns this variable to make it robust against events that don't have any existing involved cats, but do fit the requirements by the creation of new cats.  find_cats() is used as a filtering function, and if no cats for the event are found, it determines that the event doesn't meet requirements. Before this change, new cats could be created to fit roles in find_cats, so events that *only* had new cats would be accepted as valid. This new variable exists to preserve this behavior. 

I have removed preserving cats_to_create through to execute_outcome. It's only use is to check the validity of the find_cats filter check.

scripts/events_module/text_pool_event/handle_consequences.py
keyraven: Looks like one of my personal debug prints stuck around! Thanks for the catch. Removed!

scripts/events_module/patrol/patrol.py
```diff
@@ -161,6 +159,14 @@ def proceed_patrol(
 
         return self.determine_outcome(antagonize=(path == PatrolChoice.ANTAGONIZE))
 
+    def _create_needed_cats(self):
+        """
+        Creates needed cats for the patrol start.  In it's own function for units testing purposes.
```
scribblecrumb: `Creates needed cats for the patrol start.  In its own function for unit testing purposes.` typo fix

scripts/events_module/text_pool_event/find_involved_cats.py
```diff
@@ -30,10 +29,13 @@ def find_cats(
     :param other_clan: The OtherClan object involved in the event
     :return: Updated involved_cats dict with valid cats. If dict is empty, then valid cats were not found.
```
scribblecrumb: Needs to be updated

scripts/events_module/text_pool_event/find_involved_cats.py
scribblecrumb: Alright, I don't love it but I see the need.

scripts/events_module/text_pool_event/find_involved_cats.py
```diff
@@ -239,20 +223,25 @@ def _find_involved_cat(
...
-        temp_involved_cats[abbr] = possible_cats[0]
-        return temp_involved_cats
+        return {abbr: possible_cats[0]}
 
     # otherwise, let's make sure we fulfill the rel constraints with this cat
     elif possible_cats:
-        while not temp_involved_cats.get(abbr):
+        i = 0
+        while True:
+            # Protection against infinate loop.
```
scribblecrumb: `Protection against infinite loop.` typo fix

scribblecrumb (changes requested):
I'm still not loving the return of the `cats_to_create` dict, and while I was reviewing I came up with a potential solution.

I think what bothers me about it is that we're essentially returning a dict that we do nothing with. As you said, it's only use is to help determine if valid cats are found, but we don't actually use any of the information in that dict.  In this case, rather than having it be a dict, could we instead convert it to a bool value in some fashion?  And rename it? Perhaps something like `will_create_more_cats`?

keyraven:
> only use is to help determine if valid cats are found, but we don't actually use any of the information in that dict. In this case, rather than having it be a dict, could we instead convert it to a bool value in some fashion? And rename it? Perhaps something like `will_create_more_cats`?

I think that would work. Instead of a boolean, I would use a simple integer - the number of new cats to be created.  This give a bit more information, and makes more sense to me as output. And it's much less info than a dictionary. Would that work?

scripts/events_module/text_pool_event/find_involved_cats.py
keyraven: Updated!

scripts/events_module/patrol/patrol.py
keyraven: Corrected - thanks!

scripts/events_module/text_pool_event/find_involved_cats.py
keyraven: Corrected - thanks!

scribblecrumb (approved):

PR #5941, opened 2026-08-31
[CONTENT] Tint Category Rework

Inkpawwc:
## About The Pull Request
Breaks tint categories down further into more distinct groups. This will allow for stronger tints that add more pelt variation while still ensuring we don't end up with wonky colors. Adds a new dilute tint ``pinkdilute`` to showcase the new level of variation that can be added with this change.

more variation for pelts with less work

## Proof of Testing
all possible color combos for ``pinkdilute`` tint
image

``pinkdilute`` tint would appear on these pelts only
image

### DIFF
--- a/sprites/dicts/tint.json
+++ b/sprites/dicts/tint.json
@@ -23,19 +23,19 @@
   },
   "possible_tints": {
     "basic": ["pink", "gray", "red", "orange", "warmdilute"],
-    "white": ["yellow", "pinkdilute"],
+    "white": ["yellow", "pinkdilute", "earthydilute"],
     "lightgrey": ["purple", "black", "dilute", "pinkdilute"],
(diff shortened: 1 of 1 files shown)


sprites/dicts/tint.json
```diff
@@ -46,6 +54,8 @@
     "dilute": [20, 20, 20],
     "warmdilute": [25, 15, 7],
     "cooldilute": [7, 15, 25],
+    "pinkdilute": [80, 50, 45],
+    "earthydilute": [44, 33, 11],
```
scribblecrumb, line 58: Can I get examples for this tint as well?

Inkpawwc: image
image
earthy dilute tint

scribblecrumb (approved):

PR #5966, opened 2026-09-02
[LOCALISATION] PL defaults, inheritance, relationships

afellowcorn:
## About The Pull Request
Relationship tiers seen on the relationship screen, inheritance terms seen on family trees and default strings.
<hr>
image
(changed in second commit)
image

- Both "urazy" and "obrażenia" literally mean injuries. The latter is technically a more general word akin to "damages", and refers to damage that may be the result of physical injury. "Uraz" refers to the physical injury that caused damage, it can be prevented, but not treated, eg. "trauma". 
- "Obrażenia" has no singular, but I thought fit better here contextually even if the cat is healing from what is technically one injury. This string should nonetheless never be seen by the player in normal circumstances, with more specifically worded strings for each condition.
- For code references to conditions (event editor) I've chosen to use "uraz" as a code term, allowing clear distinction for number of conditions to be given.
<hr>
image

- This is a silly reference, to translate it would lose the meaning.
<hr>
image

- "zmarł" vs "zginął" — both mean "[he] died", but the former is a more neutral tone, used in obituaries or articles and other factual ways to say someone has died. The latter is a word with a more negative connotation, used for sudden or 'tragic' causes of death like disaster (or a disastrous patrol). As an aside, it can also mean "to be lost".
<hr>
image

- The literal Polish "niezainteresowany" is too long of a word (and I don't like it here), while "brak zainteresowania" is too long in general. "Brak romansu" ie. "No romance" looks nice and I think works best here to communicate to the player.

Localisation

## Linked Issues
#5363 

## Proof of Testing
Zrzut ekranu 2026-09-02 153014
Zrzut ekranu 2026-09-02 154022
Zrzut ekranu 2026-09-02 153732
Zrzut ekranu 2026-09-02 153741


afellowcorn:
Unittest failure unrelated;
Pronoun test does not take into account different pronouns.

j-gynn:
> Unittest failure unrelated; Pronoun test does not take into account different pronouns.

Can you please report this as a bug so that someone can get it fixed?

afellowcorn:
> > Unittest failure unrelated; Pronoun test does not take into account different pronouns.
> 
> Can you please report this as a bug so that someone can get it fixed?

The latter, I'm assuming?

PR #5985, opened 2026-09-06
[BUGFIX] cats doubling on patrol

### LINKED ISSUE #5975: [BUG] Med patrol chooses cat twice
### Pre-requisites

### Description

The medicine cat is helping themself interpret their own dreams

### Grade

### Gamemode

Expanded Mode

### Challenge cards

_No response_

1. Send a med cat out with 1-5 other cats
2. Have the chance of one of the cats asking for help with a StarClan message
3. Watch as the med cat talks to themselves

Patrol Leader: Lord Posyhunt
Total Number of Possible Patrols | normal: 11, romantic: 5
starting chance: 40 | EX_updated chance: 53
PATROL ID: gen_med_vision1 | SUCCESS: True
Success Outcome: gen_med_vision1_success1 | Failure Outcome: gen_med_vision1_fail1
Patrol Frequency: 1 | Patrol Weight: 5
Outcome Frequency: 2 | Outcome Weight: 1

f026ad35

### Error log

_No response_

### Additional context

_No response_

### Screenshots

Image 

Image 

Image

scribblecrumb:
## About The Pull Request
Fixes the "primary cat" also being available for use by a "plain" (not s_c) designation.

## Linked Issues
fixes #5975

## Proof of Testing
image

Same patrol as reported in the issue. It was consistently using the same cat twice while I was testing it prior to the fix, it now uses different cats as intended.

## Changelog/Credits
- Fixes cats being used multiple times in patrols when they shouldn't be
- Added an extra test to check if multiple r_c designations would take different cats

### DIFF
--- a/scripts/events_module/text_pool_event/find_involved_cats.py
+++ b/scripts/events_module/text_pool_event/find_involved_cats.py
@@ -109,8 +109,8 @@ def find_cats(
             possible_cats = interactable_cats
 
             for cat in involved_cats.values():
-                # if a cat is a list then we don't want to try and remove it
-                if not isinstance(cat, list) and cat in interactable_cats:
+                # this will only remove non-list values
+                if cat in interactable_cats:
                     possible_cats.remove(cat)
 
         if not possible_cats:
--- a/scripts/events_module/text_pool_event/find_involved_cats.py
+++ b/scripts/events_module/text_pool_event/find_involved_cats.py
@@ -14,6 +14,7 @@ from scripts.events_module.text_pool_event.text_pool_event import TextPoolEvent
 
 
 def find_cats(
+    primary_cat: Cat,
     interactable_cats: list,
     involved_cats: dict,
     outside_cats: list,
@@ -107,6 +108,10 @@ def find_cats(
         # CHECK ALL UN-USED CATS
         else:
             possible_cats = interactable_cats
+
+            # primary cat is already "in use" so we should remove them if they're present in this list
+            if primary_cat in possible_cats:
+                possible_cats.remove(primary_cat)
 
         if not possible_cats:
             # uh oh, we're out of options!
--- a/scripts/events_module/text_pool_event/find_involved_cats.py
+++ b/scripts/events_module/text_pool_event/find_involved_cats.py
@@ -14,7 +14,6 @@ from scripts.events_module.text_pool_event.text_pool_event import TextPoolEvent
 
 
 def find_cats(
-    primary_cat: Cat,
     interactable_cats: list,
     involved_cats: dict,
     outside_cats: list,
@@ -109,9 +108,10 @@ def find_cats(
         else:
             possible_cats = interactable_cats
 
-            # primary cat is already "in use" so we should remove them if they're present in this list
-            if primary_cat in possible_cats:
-                possible_cats.remove(primary_cat)
+            for cat in involved_cats.values():
+                # if a cat is a list then we don't want to try and remove it
+                if not isinstance(cat, list) and cat in interactable_cats:
+                    possible_cats.remove(cat)
 
         if not possible_cats:
             # uh oh, we're out of options!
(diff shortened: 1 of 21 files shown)


keyraven:
I think the main issue here is that find_cats is not removing the already-assigned cats in the involved cat dictionary before it goes through it's finding-cats loop. It does remove chosen cats from the list after it finds them, but does not for pre-chosen cats passed in through `involved_cats`. I think it would make more sense to remove all cats in `involved_cats` from `interactable_cats` before the finding-cats loop, since we don't ever want to repeat those.

scribblecrumb:
> I think the main issue here is that find_cats is not removing the already-assigned cats in the involved cat dictionary before it goes through it's finding-cats loop. It does remove chosen cats from the list after it finds them, but does not for pre-chosen cats passed in through `involved_cats`. I think it would make more sense to remove all cats in `involved_cats` from `interactable_cats` before the finding-cats loop, since we don't ever want to repeat those.

I considered that, but it actually wouldn't be the ideal solution.  Patrols add all of the "cat types" to the involved cats (lists of each rank on the patrol, list of normal adults, list of healers, patrol_cats, some_patrol).  So if we were to remove those cats from the interactable cats then we'd have no cats at all.

I feel the primary cat solution is a decent solution. The primary cat is always determined prior and "assumed" present, so it seems simple enough to just be passing that cat and removing them where necessary.

scribblecrumb:
Actually decided you're right on this one.  Switched it to removing the involved_cats already in use (all non-list values) and added a test for checking that multiple r_c cats will always differ.

keyraven:
Sounds good! 

It might also be helpful, although not required for this fix, to require patrols to have to have p_l listed in the involved_cats dictionary if they are mentioned in the patrol text at all. This would be consistent with all other cat's mentioned in the patrol text, like r_c's and such. Plus, would be consistent with ceremonies, where you do need to declare m_c.  

I will do a proper review of this later today!

scribblecrumb:
> Sounds good!
> 
> It might also be helpful, although not required for this fix, to require patrols to have to have p_l listed in the involved_cats dictionary if they are mentioned in the patrol text at all. This would be consistent with all other cat's mentioned in the patrol text, like r_c's and such. Plus, would be consistent with ceremonies, where you do need to declare m_c.
> 
> I will do a proper review of this later today!

You don't actually have to declare m_c on ceremonies, thoughts, transition, or relationship events. I've gone back and forth on if "primary cats" (p_l and m_c) should have to be declared or not tbh. On one hand it's less lines in the json to not declare them, on the other hand it's less explicit for the writer.

keyraven:
I would lean towards requiring it, to be explicit to the writers. But this discussion can be continued later! I will review this PR now :)

scripts/events_module/text_pool_event/find_involved_cats.py
```diff
@@ -108,6 +108,11 @@ def find_cats(
         else:
             possible_cats = interactable_cats
 
+            for cat in involved_cats.values():
+                # if a cat is a list then we don't want to try and remove it
+                if not isinstance(cat, list) and cat in interactable_cats:
```
keyraven: Do we need to check if it's a list? The intractable_cats already only contains cats, to the check "cat in interactable_cats should cover it. For clarity, you can add a comment saying that this will only remove non-list values of involved_cats from interactable_cats.

tests/test_patrols.py
```diff
@@ -195,6 +195,51 @@ def test_overall_pl_rc_difference(self):
...
+            rank=CatRank.APPRENTICE, moons=10, experience=1
+        )
+
+        patrol = PatrolEvent(
+            event_id="test",
+            types=["hunting"],
+            intro_strings=["test"],
+            decline_strings=["test"],
+            involved_cats={
+                "p_l": InvolvedCatDict(),
```
keyraven: You should add a version of this test where p_l is not stated in involved_cats, since that what the situation that caused the original error.

keyraven (changes requested):

keyraven (approved):

PR #6005, opened 2026-09-07
[CONTENT] New Tints

fulminousnight:
## About The Pull Request

- Adds six new tints: Gloaming, Tawny, Night, Patina, Quartz, and Lilac Dilute for certain pelts
- Makes rust its own tint category due to the unique nature of the pelt. 

- More variety to cat appearance without entirely new pelt color recipes. 

## Proof of Testing
No Tint:
No Tint

New Tints and the pelts applicable:
Gloaming_Tint
Lilac Dilute Tint
Night_Tint New
Patina Tint
tawny_tint
quartz

### DIFF
--- a/sprites/dicts/tint.json
+++ b/sprites/dicts/tint.json
@@ -20,23 +20,24 @@
     "GOLDEN-BROWN": "brown",
     "DARKBROWN": "darkbrown",
     "CHOCOLATE": "darkbrown",
-    "RUST": "darkbrown"
+    "RUST": "rust"
   },
   "possible_tints": {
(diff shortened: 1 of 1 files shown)


scribblecrumb:
Could we remove GINGER from night? Imo the cat looks more like they're just standing in shadow, which I'm not a fan of.

fulminousnight:
> Could we remove GINGER from night? Imo the cat looks more like they're just standing in shadow, which I'm not a fan of.

It would also remove it from Dark Ginger as the two are grouped together but I am happy to do so : )

scribblecrumb (approved):

PR #6041, opened 2026-09-13
[COLOR] Chinchilla

Yzzuwyzzuf:
## About The Pull Request

adds the new coat, chinchilla, a silvery white coat with more distinct markings inspired by chinchilla persians and silver tabbies

allows people to have distinctive markings on their white cats without needing to use the bengals and marbleds

## Proof of Testing

image

generated cats below. may be tinted
image
image
image
image
image


Yzzuwyzzuf:
What it should look like once the gradients are added
d2759c7a-e59f-4644-9de1-e8e970b7d507-1

fulminousnight:
Fuzzy is freckled supposed to be that purple or was that an accident?

